PR #33, opened 2019-01-21
CI & coverage setup

palango:
This sets up basic CI and coverage.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,4 +14,6 @@ isort:
 	isort $(ISORT_PARAMS)
 
 test:
-	py.test -v
+	py.test -v libs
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -0,0 +1,19 @@
+CODE_DIRS = monitoring/monitoring_service/ monitoring/request_collector/ pathfinding/pathfinding_service
```
karlb: What about `libs`?

palango: Good point

palango: Oh, there's quite some errors with that. I'll create a follow-up issue.

PR #41, opened 2019-01-28
Remove unused monitor_addres from MR

### LINKED ISSUE #20: Remove `monitor_address` from MonitorRequest
The current plan is to broadcast MonitorRequests to all registered MS via Matrix. With this in mind, having a specific `monitor_address` makes no sense.

- [x] check if spec needs update
- [x] wait for monorepo
- [x] remove `monitor_address` and all of its usages in a single PR

karlb:
Closes #20.


palango:
I rebased and force pushed

PR #42, opened 2019-01-28
MS on event based architecture

palango:
WIP for the event based MS.

Also changes the project setup a bit, code is now in `src` and the tests in `tests` .


monitoring/src/monitoring_service/blockchain.py
```diff
@@ -0,0 +1,215 @@
...
+            to_block=to_block,
+        )
+
+        for event in registry_events:
+            yield ReceiveTokenNetworkCreatedEvent(
+                token_network_address=event['args']['token_network_address'],
+            )
+
+        # then check all token networks
+        # the new token networks are registered by now
```
karlb: This assumes that the `ReceiveTokenNetworkCreatedEvent`s are processed immediately on yield and added to context.ms_state, right? It's hard to get a proper overview in such a huge chage, but this might violate the abstraction between BlockchainListener and Monitoring service.

monitoring/src/monitoring_service/handlers.py
```diff
@@ -0,0 +1,300 @@
...
+    An event handler needs to be idempotent as events can be triggered
+    multiple times.
+    """
+    def handle_event(self, event: Event):
+        raise NotImplementedError
+
+
+@dataclass
+class TokenNetworkCreatedEventHandler(EventHandler):
+    context: Context
```
karlb: We could also use plain function as event handlers if we add `context` as a parameter to the function call.

monitoring/src/monitoring_service/handlers.py
```diff
@@ -0,0 +1,300 @@
...
+
+    def handle_event(self, event: Event):
+        if isinstance(event, ReceiveTokenNetworkCreatedEvent):
```
karlb: Does mypy understand
```python
    assert isinstance(event, ReceiveTokenNetworkCreatedEvent)
```
? That would give us an error in the bad case and reduce the nesting level.

monitoring/src/monitoring_service/states.py
```diff
@@ -0,0 +1,117 @@
...
+    channel_identifier: int
+    token_network_address: str
+    chain_id: int
+
+    balance_hash: str
+    nonce: int
+    additional_hash: str
+    signature: str
+
+    # FIXME: necessary?
```
karlb: If that's properly baked into the signing function (and doesn't change), it's not needed.

monitoring/src/monitoring_service/states.py
```diff
@@ -0,0 +1,117 @@
...
+    # signature_prefix: str
+    # message_length: str
+    # message_type_id: int
+
+    # reward infos
+    non_closing_signature: str
+    reward_amount: int
+    reward_proof_signature: str
+
+    def packed_balance_proof_data(
```
karlb: We should check for duplication of such helper functions. I'm currently lacking the overview to say anything specific.

palango:
@karlb I'm sorry, this is horrible to review. In the middle I decided to follow some advice of the test review and split code in `src` and `tests` folders. The result is this giant PR with unrelated changes.

Give it a look whenever you find time, I'll add more tests in the mean time.

monitoring/src/monitoring_service/blockchain.py
palango: Yes, this one is hard to decide. Right now it's kind of it's own event queue, that handles the whole state lifecycle and triggers events for all monitoring related tasks.

We can combine these, but then in the BC listener it becomes a bit more difficult to properly handle all events.

monitoring/src/monitoring_service/handlers.py
palango: True, that might make it even clearer. I'll fix this.

monitoring/src/monitoring_service/handlers.py
palango: Yes, it does. Changed as well.

monitoring/src/monitoring_service/states.py
palango: Yes, this is copied from the libs messages, as these will not be needed when we use the raiden messages.

monitoring/src/monitoring_service/blockchain.py
palango: I pushed some changes we discussed, can you have another look?

monitoring/src/monitoring_service/states.py
palango: @Dominik1999 These seem to be unnecessary to send with the message. I removed them here.

PR #45, opened 2019-01-29
Cmd params

palango:
Mostly copying over stuff from the PFS.


monitoring/src/monitoring_service/service.py
```diff
@@ -0,0 +1,104 @@
...
+            db=self.database,
+            scheduled_events=self.scheduled_events,
+            w3=self.web3,
+            contract_manager=contract_manager,
+            last_known_block=0,
+        )
+
+    def start(self, wait_function: Callable = time.sleep):
+        while True:
+            last_block = self.web3.eth.blockNumber - self.required_confirmations
```
karlb: We'll miss previously opened channels if we only start here, won't we? We'll have to touch this anyway when loading/saving state between restart, again.

palango: Atm yes, but once the `BlockchainState` gets persisted (can one write that?) it should work just fine.

PR #49, opened 2019-01-30
Move all services in a single package

palango:
This will make maintenance a lot easier.


requirements.txt
```diff
@@ -1,18 +1,25 @@
...
 # until a stable version with handle_exception is released on pip
 git+https://github.com/matrix-org/matrix-python-sdk.git
 
 git+https://github.com/raiden-network/raiden-contracts.git
-git+https://github.com/raiden-network/raiden-libs.git
+# raiden-contracts==0.8.0
 
-gevent
-jsonschema
+dataclasses==0.6
```
karlb: Not really related to this PR: Raiden just moved to python3.7. If we follow suit, we won't need this as a dependency, anymore.

src/raiden_libs/utils/signing.py
```diff
@@ -138,21 +137,11 @@ def eth_sign(
         v: int = 27,
         hasher: Hasher = eth_sign_sha3,
 ) -> bytes:
-    warnings.warn(
```
karlb: Should we remove these warnings? It sounds like we should remove these functions and call functions from `raiden`, instead.

requirements.txt
palango: Good point. The porblem is that there is no python 3.7 environment on travis yet. So either we switch to CircleCI as well ort find another solution, but I agree we should use 3.7 as well if Raiden does it.

src/raiden_libs/utils/signing.py
palango: I was mostly annoyed by the warning, but I can remove the commit before merging

requirements.txt
palango: Opened #53

Issue #18, opened 2019-01-07
MS updates channel state on chain


palango:
## Description

**As Josh I want the MS to be able to**

update the channel state on-chain

**So that it can**

send my BP while I'm offline

## Acceptance Criteria

- [x] MS loads latest known BP from database
- [x] MS sends BP to smart contract and makes sure the transaction is mined before the settle timeout
- [ ] Update is sent only after a period to give the client time to react itself.

## Related story / issue
- https://github.com/raiden-network/raiden-monitoring-service/issues/38

## Reviewer
-

5 SP

PR #55, opened 2019-01-31
On-chain calls to MSC

palango:
Part of #18


src/monitoring_service/blockchain.py
```diff
@@ -34,6 +38,12 @@ def create_channel_event_topics() -> List:
     ]
 
 
+def create_monitoring_service_topics() -> List:
```
karlb: If we don't reuse this anywhere, we can also put `[None]` inline and remove the separate function.

src/monitoring_service/blockchain.py
```diff
@@ -15,13 +15,17 @@
     ReceiveChannelSettledEvent,
     ReceiveNonClosingBalanceProofUpdatedEvent,
     UpdatedHeadBlockEvent,
+    ReceiveMonitoringNewBalanceProofEvent,
+    ReceiveMonitorinRewardClaimedEvent,
```
karlb: There a "g" missing in "ReceiveMonitorin"

src/monitoring_service/handlers.py
```diff
@@ -178,24 +191,25 @@ def action_monitoring_triggered_event_handler(event: Event, context: Context):
...
+                monitor_request.additional_hash,
+                monitor_request.closing_signature,
+                monitor_request.non_closing_signature,
+                monitor_request.reward_amount,
+                monitor_request.token_network_address,
+                monitor_request.reward_proof_signature,
+            ).transact({'gas': 100_000})  # TODO: estimate gas better here
+            log.info(f'Submit MR to SC, got tx_hash {encode_hex(tx_hash)}')
+            assert tx_hash is not None
+            data = context.w3.eth.waitForTransactionReceipt(tx_hash)
```
karlb: Is this quick enough to be done with blocking?

tests/monitoring/monitoring_service/test_end_to_end.py
```diff
@@ -1,6 +1,51 @@
...
+from monitoring_service.states import MonitorRequest
+from raiden_contracts.constants import CONTRACT_MONITORING_SERVICE, MonitoringServiceEvent, CONTRACT_TOKEN_NETWORK
+from raiden_contracts.contract_manager import ContractManager
+from raiden_libs.blockchain import BlockchainListener
+
+TEST_POLL_INTERVAL = 0.1
+log = structlog.get_logger(__name__)
+
+
+class Validator(BlockchainListener):
```
karlb: The `event_handler` fixture from raiden contracts looks like a better solution than this. But we can do that in another PR.

tests/monitoring/monitoring_service/test_end_to_end.py
```diff
@@ -66,29 +113,124 @@ def test_e2e(
...
+        c2.address,
+        balance_proof_c2,
+        reward_amount,
+        monitoring_service.address,
+    )
+    # wait for channel open event to be processed by the MS
+    wait_for_blocks(1)
+    gevent.sleep(0)
+
+    mr = MonitorRequest(
```
karlb: This would be much shorter if we could use the same MR object.

src/monitoring_service/blockchain.py
palango: I already opened #56 for that, let's handle that there.

src/monitoring_service/blockchain.py
palango: Nice catch.

src/monitoring_service/handlers.py
palango: Ui, I missed that. That should go away.

src/monitoring_service/handlers.py
```diff
@@ -184,34 +219,41 @@ def action_monitoring_triggered_event_handler(event: Event, context: Context):
...
-                monitor_request.reward_proof_signature,
-            ).transact({'gas': 100_000})  # TODO: estimate gas better here
-            log.info(f'Submit MR to SC, got tx_hash {encode_hex(tx_hash)}')
-            assert tx_hash is not None
-            data = context.w3.eth.waitForTransactionReceipt(tx_hash)
-            log.info('MINED', tx=data)
-        except Exception as e:
-            log.error('Sending tx failed', exc_info=True, err=e)
+
+        if channel is not None and channel.closing_tx_hash is None:
```
karlb: If somebody else provided a more recent BP, we're not allowed to claim. I don't see a check for that case.

tests/monitoring/monitoring_service/test_end_to_end.py
palango: True, now I can fix this directly in the `RaidenMockClient` class.

tests/monitoring/monitoring_service/test_end_to_end.py
palango: See #58

src/monitoring_service/handlers.py
palango: Should be fixed, pls have another look.

PR #59, opened 2019-02-01
Add client update period of 30% settle period before MS intervention

### LINKED ISSUE #18: MS updates channel state on chain

palango:
Fixes #18


src/monitoring_service/constants.py
```diff
@@ -1,2 +1,2 @@
-WAIT_BLOCKS_BEFORE_MONITOR: int = 0
+PERCENT_SETTLE_TIMEOUT_BEFORE_MONITOR: float = 0.3
```
karlb: "percent" is misleading in this case. This should have a value of `30` for 30% or be called differently (e.g. "ratio" instead of "percent").

palango: renamed to `RATIO_OF_SETTLE_TIMEOUT_BEFORE_MONITOR`

PR #68, opened 2019-02-01
Don't generate universal wheels

palango:


setup.py
```diff
@@ -39,7 +39,7 @@ def read(*names, **kwargs):
 
 setup(
     name='raiden-services',
-    version='0.1.0',
```
LefterisJP: Did the version go down here???

palango: Yes, we never released before and I just wanted to reserve the PyPI package name

ulope:
Why don’t you want universal wheels?

ulope:
Answering my own question: Because universal implies Python 2 compatibility. TIL.

PR #69, opened 2019-02-04
Adds tests to makefile and resolves most make lint errors

Dominik1999:
3 errors remain, but will be solved this week when files are changed

### DIFF
--- a/src/raiden_libs/transport/transport.py
+++ b/src/raiden_libs/transport/transport.py
@@ -19,7 +19,7 @@ class Transport(gevent.Greenlet):
     """
     def __init__(self):
         super().__init__()
-        self.message_callbacks = list()
+        self.message_callbacks: list = list()
 
     def add_message_callback(self, callback):
         self.message_callbacks.append(callback)
--- a/src/raiden_libs/transport/transport.py
+++ b/src/raiden_libs/transport/transport.py
@@ -1,6 +1,6 @@
 import json
 import logging
-from typing import Union
+from typing import Union, List
 
 import gevent
 import jsonschema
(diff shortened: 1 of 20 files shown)


src/raiden_libs/transport/transport.py
```diff
@@ -19,7 +19,7 @@ class Transport(gevent.Greenlet):
     """
     def __init__(self):
         super().__init__()
-        self.message_callbacks = list()
+        self.message_callbacks: list = list()
```
palango: While `list` works, we always try to use classes from the `typing` module as annotations. So `typing.List` here.

tests/pathfinding/test_graphs.py
```diff
@@ -8,16 +8,17 @@
...
-from raiden_libs.types import Address
+from raiden_libs.types import Address, ChannelIdentifier
 
 
 # This test is boring right now, but should get more interesting as the routing
 # gets more options.
 def test_edge_weight(addresses):
-    a = addresses[0]
-    b = addresses[1]
+    a = ChannelIdentifier(1)
```
palango: can you sue better names here

src/raiden_libs/transport/transport.py
Dominik1999: ` self.message_callbacks: typing.List = list()` ?

palango: yes, or import list from typing first and then just use it there

tests/pathfinding/test_graphs.py
Dominik1999: done

palango: Do't see it yet

tests/monitoring/conftest.py
```diff
@@ -4,7 +4,6 @@
 """
 
 from raiden_libs.test.fixtures import *  # noqa
-from raiden_libs.test.fixtures.web3 import *  # noqa
```
palango: Don't remove this one, but the line before.

PR #71, opened 2019-02-04
Handle updates

### LINKED ISSUE #66: Take on-chain BP updates from Raiden clients into account
## Problem Definition

Currently we will always call `monitor` when we received a `MonitorRequest`, even if the client called `update` itself. In that case there is not need for the MS to intervene.

## ToDo
- [ ] Update `channel_non_closing_balance_proof_updated_event_handler` to store the latest BPs nonce
- [ ] Update `action_monitoring_triggered_event_handler` to take this information into account and do not call `monitor` when the update has already been done by the client.

8SP

palango:
Fixes #66

### DIFF
--- a/src/monitoring_service/handlers.py
+++ b/src/monitoring_service/handlers.py
@@ -303,43 +303,53 @@ def action_monitoring_triggered_event_handler(event: Event, context: Context):
         token_network_address=event.token_network_address,
         channel_id=event.channel_identifier,
     )
+    if monitor_request is None:
+        return
 
-    if monitor_request is not None:
-        channel = context.db.get_channel(
-            token_network_address=monitor_request.token_network_address,
-            channel_id=monitor_request.channel_identifier,
-        )
+    channel = context.db.get_channel(
+        token_network_address=monitor_request.token_network_address,
+        channel_id=monitor_request.channel_identifier,
+    )
 
-        if channel is not None and channel.closing_tx_hash is None:
-            # FIXME: don't monitor when closer is MR signer
-            try:
-                tx_hash = context.monitoring_service_contract.functions.monitor(
-                    monitor_request.signer,
-                    monitor_request.non_closing_signer,
-                    monitor_request.balance_hash,
-                    monitor_request.nonce,
-                    monitor_request.additional_hash,
-                    monitor_request.closing_signature,
-                    monitor_request.non_closing_signature,
-                    monitor_request.reward_amount,
-                    monitor_request.token_network_address,
-                    monitor_request.reward_proof_signature,
-                ).transact({'from': context.ms_state.address})
+    if channel is None:
+        return
 
-                log.info(
-                    'Calling `monitor` on channel',
-                    token_network_address=channel.token_network_address,
-                    channel_identifier=channel.identifier,
-                    transaction_hash=encode_hex(tx_hash),
-                )
-                assert tx_hash is not None
+    last_onchain_nonce = 0
+    if channel.update_status:
+        last_onchain_nonce = channel.update_status.nonce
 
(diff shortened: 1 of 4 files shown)


src/monitoring_service/handlers.py
```diff
@@ -135,6 +136,47 @@ def channel_non_closing_balance_proof_updated_event_handler(event: Event, contex
             token_network_address=event.token_network_address,
             identifier=event.channel_identifier,
         )
+
+        non_closing_participant = channel.participant1
+        if event.closing_participant == channel.participant1:
+            non_closing_participant = channel.participant2
```
karlb: To me, an if/else would be clearer then overwriting the variable, again.

tests/monitoring/monitoring_service/test_handlers.py
```diff
@@ -134,3 +144,246 @@ def test_channel_closed_event_handler_leaves_existing_channel(
...
+    channel_non_closing_balance_proof_updated_event_handler(event4, context)
+
+    assert len(context.db.channels) == 1
+    channel = context.db.get_channel(event3.token_network_address, event3.channel_identifier)
+    assert channel.update_status is not None
+    assert channel.update_status.nonce == 5
+    assert channel.update_status.update_sender_address == 'A'
+
+
+def test_monitor_new_balance_proof_event_handler_ignores_other_event(
```
karlb: Maybe, we can have one test to test this for all event handler? This get's a bit repetitive.

tests/monitoring/monitoring_service/test_handlers.py
```diff
@@ -134,3 +144,246 @@ def test_channel_closed_event_handler_leaves_existing_channel(
...
+    event = Event()
+
+    with pytest.raises(AssertionError):
+        monitor_new_balance_proof_event_handler(
+            event=event,
+            context=context,
+        )
+
+
+def test_monitor_new_balance_proof_event_handler_sets_update_status(
```
karlb: Does this test include everything from `test_channel_bp_updated_event_handler_sets_update_status_if_not_set`? If so, do we still need the latter?

tests/monitoring/monitoring_service/test_handlers.py
```diff
@@ -134,3 +144,246 @@ def test_channel_closed_event_handler_leaves_existing_channel(
...
+    event2 = ReceiveChannelClosedEvent(
+        token_network_address='abc',
+        channel_identifier=3,
+        closing_participant='B',
+        block_number=52,
+    )
+    channel_closed_event_handler(event2, context)
+
+    assert len(context.db.channels) == 1
+    assert context.db.channels[0].state == ChannelState.CLOSED
```
karlb: Is everything up to this point the same as in the other tests? If yes, we can put it into a shared function or fixture. It's a bit much code to see the actually relevant part of each test, otherwise.

palango: True, I needed some external motivation to do this, copy and paste is so nice ;)

tests/monitoring/monitoring_service/test_handlers.py
palango: This one tests that the events from the MSC get correctly applied, the other the update event from the TokenNetwork contract.

tests/monitoring/monitoring_service/test_handlers.py
```diff
@@ -55,6 +59,54 @@ def test_channel_opened_event_handler_ignores_other_event(
             context=context,
         )
 
+    with pytest.raises(AssertionError):
```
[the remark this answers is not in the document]
palango: Even nicer

PR #72, opened 2019-02-04
WIP: Enable request collector cli tests

karlb:
The filename didn't match `test_*` and has been skipped during test
exection.


src/request_collector/cli.py
```diff
@@ -109,7 +109,7 @@ def main(
         monitoring_channel,
     )
 
-    database = StateDBSqlite(state_db)
+    database = StateDBSqlite(state_db + log_level)
```
palango: what is the ` log_level` for?

karlb: Just left over from debugging. I thought maybe the db was still locked when the test hang. That must be removed, of course.

PR #78, opened 2019-02-05
Add basic Sphinx docs and requirements

palango:
Part of #2 

With the separate `requirements-docs.txt` it should be possible to get readthedocs running.


docs/conf.py
```diff
@@ -0,0 +1,173 @@
...
+# -- Project information -----------------------------------------------------
+
+project = 'Raiden Services'
+copyright = '2019, Brainbot Labs Est.'
+author = 'Brainbot Labs Est.'
+
+# The short X.Y version
+version = ''
+# The full version, including alpha/beta/rc tags
+release = '0.0.1'
```
karlb: Can we easily use the same value here and in `setup.py`?

palango: Raiden uses `bumpversion` to automatically update all version numbers.

palango: I'll add this to the packaging ToDos

PR #82, opened 2019-02-06
Add test for `monitor` being called after ActionMonitoringTriggeredEvent

### LINKED ISSUE #29: MS received MR after ChannelClose event
The client is most likely still online at this point and can provide the BP himself, so the MS could just discard this MR. On the other hand the client might not have a reliable connection or be unable to listen to the ChannelClosed event itself due to some implementation limitation.

Split off from #17.

5 SP

palango:
Fixes #29


palango:
Oh, forgot to check that the event is properly triggered as well.

Issue #22, opened 2019-01-11
Continuous deployment


palango:
A goal of issue https://github.com/raiden-network/raiden-pathfinding-service/issues/82 was to create a continuous deployment pipeline for the PFS. This makes sense for the MS as well.

This is blocked by the deployment of the MS contract on the testnets.

The current setup is at https://github.com/palango/services-dev
### Tasks

- [ ] Setup deployment pipeline

karlb:
> Set up accounts for the test chains (1 set for PFS, on set for MS)

Why not just use a single account for everything?

palango:
> Why not just use a single account for everything?

Should be fine as well

karlb:
Up to date PFSs are running:
https://pfs-rinkeby.services-dev.raiden.network/api/v1/info
https://pfs-ropsten.services-dev.raiden.network/api/v1/info
https://pfs-kovan.services-dev.raiden.network/api/v1/info

PR #83, opened 2019-02-06
Add .dockerignore and update Dockerfile for PFS

palango:
Part fo #22

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,10 @@
+Dockerfile*
+docker-compose*
+.dockerignore
+.git
+.gitignore
+README.md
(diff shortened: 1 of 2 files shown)


.dockerignore
```diff
@@ -0,0 +1,10 @@
+Dockerfile*
+docker-compose*
+.dockerignore
+.git
+.gitignore
+README.md
+LICENSE
+.vscode
+.idea
+venv
```
karlb: I have `.pytest_cache` and `mypy_cache` lying around and I'm probably not the only one. Those make a good addition to this list.

PR #85, opened 2019-02-07
Add POA middleware to make PFS run on POA test chains

palango:
Part of #22


palango:
@karlb https://web3py.readthedocs.io/en/stable/middleware.html#why-is-geth-poa-middleware-necessary

PR #86, opened 2019-02-07
Test event filtering boundaries and simplify limit calculation in BC listener

palango:
Add a test for the assumption and simplify the code a bit according to that.

### DIFF
--- a/tests/monitoring/monitoring_service/test_blockchain.py
+++ b/tests/monitoring/monitoring_service/test_blockchain.py
@@ -1,0 +1,97 @@
+from web3 import Web3
+
+from monitoring_service.blockchain import query_blockchain_events
+from raiden_contracts.constants import CONTRACT_TOKEN_NETWORK_REGISTRY, EVENT_TOKEN_NETWORK_CREATED
+from raiden_contracts.contract_manager import ContractManager
+
+
+def create_tnr_contract_events_query(
+    web3: Web3,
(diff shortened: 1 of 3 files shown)


tests/monitoring/monitoring_service/test_blockchain.py
```diff
@@ -0,0 +1,97 @@
...
+from raiden_contracts.constants import CONTRACT_TOKEN_NETWORK_REGISTRY, EVENT_TOKEN_NETWORK_CREATED
+from raiden_contracts.contract_manager import ContractManager
+
+
+def create_tnr_contract_events_query(
+    web3: Web3,
+    contract_manager: ContractManager,
+    contract_address: str,
+):
+    def f():
```
Dominik1999: can you give the function a proper name?

tests/monitoring/monitoring_service/test_blockchain.py
```diff
@@ -0,0 +1,97 @@
...
+    contracts_manager,
+    token_network_registry_contract,
+):
+    query = create_tnr_contract_events_query(
+        web3,
+        contracts_manager,
+        token_network_registry_contract.address,
+    )
+
+    # this generates a new token network and registers it with the registry
```
Dominik1999: what is "this" in the comments? does "generate_raiden_clients(2)" generate a new token network?

PR #87, opened 2019-02-07
Sets ´topics=[None]´ for event filtering of the MS contract

Dominik1999:


palango:
Ah, the commit message is a bit misleading, this doesn't remove the topics and it is not the ` eth_getLogs`  function... 🤓

palango:
Can you change the commit message as well?

Issue #63, opened 2019-02-01
Add persistent database


palango:
## Problem Definition

Currently the MS used a simple in-memory database. This should be replaces with a proper sqlite version that persists the state on disk.

- [x] save and load state to/from db instead of keeping data inside `Database`
- [x] initialize db properly and check parameters when db is already present
- [x] use proper pkey for saving/loading MRs (non_closing_signer is missing)
- [x] Use the same MR object in MS and request collector
- [x] Reduce state related code duplication between MS and request collector 
- [x] test that loading data after saving results in the same data
- test that MS works properly after crashing during processing a block (will be done as part of #32)

13 SP

karlb:
- [x] save and load state to/from db instead of keeping data inside `Database`
- [x] initialize db properly and check parameters when db is already present
- [x] use proper pkey for saving/loading MRs (non_closing_signer is missing)
- [ ] Use the same MR object in MS and request collector
- [ ] Reduce state related code duplication between MS and request collector 
- [ ] test that loading data after saving results in the same data
- [ ] test that MS works properly after crashing during processing a block

PR #88, opened 2019-02-08
Persist MS state to db

karlb:
Most of the work for #63, but tests are still missing.


.gitignore
```diff
@@ -106,3 +106,6 @@ venv.bak/
 # IDEs
 .vscode/
 .idea/
+
+# profiles files generated by pytest-profiling
+/prof/
```
palango: Did the output help you? Looks quite nice on their website.

src/monitoring_service/service.py
```diff
@@ -55,18 +55,15 @@ def __init__(
             address=monitor_contract_address,
         )
 
-        chain_state = BlockchainState(
-            token_network_registry_address=registry_address,
-            monitor_contract_address=monitor_contract_address,
-            latest_known_block=sync_start_block,
+        chain_id = int(web3.net.version)
+        self.database = Database(
+            filename=db_filename,
```
palango: Maybe it would be good to log the DB location somewhere

src/request_collector/store_monitor_request.py
```diff
@@ -40,27 +42,26 @@ def _run(self):
...
 
-        self.state_db.store_monitor_request(self.msg)
+        self.state_db.upsert_monitor_request(self.msg)
         return True
 
-    def check_channel(self, monitor_request: MonitorRequest, channel: sqlite3.Row):
+    def check_channel(self, monitor_request: MonitorRequest, channel: Channel):
         """We must know about the channel and it must be open"""
-        return channel is not None and channel['state'] == ChannelState.OPENED
+        return channel is not None and channel.state == ChannelState.OPENED
```
palango: I think we no longer have this constraint, as we said that we allow to provide MRs before the channel is opened (maybe due to confirmations)  and after it was closed

src/monitoring_service/database.py
```diff
@@ -1,51 +1,229 @@
...
+            hex(channel.identifier),
+            channel.participant1,
+            channel.participant2,
+            hex(channel.settle_timeout),
+            channel.state,
+            hex(channel.closing_block) if channel.closing_block else None,
+            channel.closing_participant,
+            channel.closing_tx_hash,
+            channel.claim_tx_hash,
+            dataclasses.astuple(channel.update_status) if channel.update_status else None,
```
palango: We could also flatten these into the channel table, not sure what's easiest here.

src/monitoring_service/database.py
```diff
@@ -1,51 +1,229 @@
...
+
+        if initialized:
+            old_settings = self.conn.execute("""
+                SELECT chain_id,
+                       monitor_contract_address,
+                       token_network_registry_address,
+                       receiver
+                FROM blockchain
+            """).fetchone()
+            assert tuple(old_settings) == settings, f"{tuple(old_settings)} != {settings}"
```
palango: Can you add some error handling here.

src/monitoring_service/database.py
```diff
@@ -1,51 +1,229 @@
...
+                "INSERT INTO token_network VALUES (?)",
+                [[address] for address in state.blockchain_state.token_network_addresses],
+            )
+
+    def load_state(self, sync_start_block: int) -> MonitoringServiceState:
+        """ Load MS state from db or return a new empty state if not saved one is present
 
-        return None
+        An empty state is initialized with `latest_known_block =
+        sync_start_block - 1`. If a saved state is present, `sync_start_block`
```
palango: Why the - 1 here? Won't that create problems during the first sync, where `sync_start_block` is 0?

src/monitoring_service/schema.sql
```diff
@@ -0,0 +1,54 @@
...
+
+
+CREATE TABLE channel (
+    token_network_address   CHAR(42) NOT NULL,
+    identifier              HEX_INT  NOT NULL,
+    participant1            CHAR(42) NOT NULL,
+    participant2            CHAR(42) NOT NULL,
+    settle_timeout          HEX_INT  NOT NULL,
+    -- see raiden_contracts.constants.ChannelState for value meaning
+    state                   INT NOT NULL CHECK (state >= 0 AND state <= 4),
```
palango: Not for this PR, just a thought: atm we don't use the state, maybe we can remove it all together

src/monitoring_service/schema.sql
```diff
@@ -0,0 +1,54 @@
...
+    NULL
+);
+
+CREATE TABLE token_network (
+    address                 CHAR(42) PRIMARY KEY
+);
+
+
+CREATE TABLE channel (
+    token_network_address   CHAR(42) NOT NULL,
```
palango: Would it make sense to reference the `token_network` table here? I don't see any advantage at the moment because they are never removed.

src/monitoring_service/schema.sql
```diff
@@ -0,0 +1,54 @@
...
+    additional_hash         CHAR(32)    NOT NULL,
+    closing_signature       CHAR(34)    NOT NULL,
+
+    non_closing_signature   CHAR(160)   NOT NULL,
+    reward_amount           HEX_INT     NOT NULL,
+    reward_proof_signature  CHAR(42)    NOT NULL,
+
+    non_closing_signer      CHAR(42)    NOT NULL,
+    PRIMARY KEY (channel_identifier, token_network_address, non_closing_signer)
+    --FOREIGN KEY (channel_identifier, token_network_address)
```
palango: Is this a TODO?

src/raiden_libs/messages/balance_proof.py
```diff
@@ -1,3 +1,5 @@
+from typing import Any, Type
```
palango: With the `libs.messages.monitor_request` module removed, is this still needed or can we remove it?

src/request_collector/cli.py
```diff
@@ -109,7 +109,7 @@ def main(
         monitoring_channel,
     )
 
-    database = StateDBSqlite(state_db)
+    database = SharedDatabase(state_db)
```
palango: What happens if the DB doesn't exist yet at this point?

.gitignore
karlb: As soon as gevent is involved, the output is messed up. For now, `pytest --durations` is more helpful.

src/monitoring_service/database.py
karlb: In `get_events`, we skip the skip `chain_state.latest_known_block` and continue with the next one, so we need `-1` to actually read all blocks after *and including* `sync_start_block`. Unfortunately, we have no tests yet that check that these boundaries are handled correctly.

palango: Makes sense, I guess then we need a `max(0, sync_start_block - 1)`

src/monitoring_service/schema.sql
karlb: Yes, since we have the strict ordering here (as opposed to the monitoring request -> channel relation), it makes sense to add a foreign key. I'll do so.

src/monitoring_service/schema.sql
karlb: It would be nice to have a foreign key, but since we allow inserts in the "wrong" order (MR before we get the confirmed ChannelOpened event), we can't enforce it, right now.

src/request_collector/cli.py
karlb: We get an error about missing tables. I'll add a more informative error message.

src/monitoring_service/database.py
karlb: I thought about flattening, too. But I had too many half-finished changes at that point, so I didn't want to introduce changes in even more places. I can do so, now.

palango: Doesn't need to be for this PR, and just if you feel it's worth the work. Otherwise we can see how this evolves (maybe more stuff gets into the `update_status`)

src/raiden_libs/messages/balance_proof.py
palango: Clirification: I'm talking about the whole `Message` and `BalanceProof` classes.

karlb: It's still used at least in `test_e2e` via `get_balance_proof` as well as in many raiden_libs tests. We can probably remove it with a few changes and lots of deleted raiden_libs code. But let's do that in a different PR.

src/request_collector/store_monitor_request.py
karlb: Right. In the long run we need tests that exercise these cases.

palango: Yes, we need to state these assumptions somewhere, test them (in the MS part this is done already) and also integrate the request collector in the end to end test.

src/monitoring_service/schema.sql
karlb: I would be surprised if we won't need this for some validation or idempotence guarantees.

src/monitoring_service/database.py
karlb: If we put the max here, it won't work. Then latest_known_block will be 0 and we'll start processing with block 1. We can add it inside `get_events` or just skip it because the values passed in via command line can't be smaller than. If we put it inside `get_events`, I would prefer an error to silently using 0, instead.

src/request_collector/cli.py
karlb: We now get
```
sqlite3.OperationalError: unable to open database file
```
right after the log message `Opening database at ...`. That's not super explanatory, but ok. We can add more detail once we have a clear idea how MR and RC are usually started.

PR #90, opened 2019-02-11
Fix state update when new token network is found

palango:


src/monitoring_service/service.py
```diff
@@ -92,25 +92,28 @@ def start(self, wait_function: Callable = time.sleep) -> None:
                 to_block=last_block,
             )
 
-            self.context.ms_state.blockchain_state = new_chain_state
+            if self.context.ms_state.blockchain_state != new_chain_state:
+                self.context.ms_state.blockchain_state = new_chain_state
+                self.context.db.update_state(self.context.ms_state)
```
karlb: I think I would prefer just upserting the newly found token_network instead of doing this. It's not really obvious to me which changes happen at this point.

palango: Fixed

PR #95, opened 2019-02-12
Matrix broadcasting integration into MS

### LINKED ISSUE #14: Monitoring Service listens to MR posts
## Description

**As Josh I want the MS to be able to**

listen to the BP broadcast matrix room

**So that it can**

receive my latest BPs

## Acceptance Criteria

- [x] Matrix Room is available
- [x] MS needs to join the matrix room
- [x] Format is agreed with client, smart contract and transport team
- [ ] MS collects all MRs from broadcasting room matching the chain id
- [x] MS saves MR in db

## Related story / issue
- #12 

## Reviewer
-

8 SP

palango:
This is a first try at the broadcasting integration into the MS request collector.

Until the new raiden-contracts version is released and updated this requires a custom raiden branch.

It's also unfortunate that we need to downgrade some dependencies, see https://github.com/raiden-network/raiden/issues/3460 for that.

The matrix stuff might benefit from being separated into a own class, on the other hand we might wait until the Raiden `MatrixTransport` class is ready to use. @ulope Opinions on that?

@andrevmatos @ulope Can you have a quick look if I call the stuff in the right way/order.

Closes #14


src/request_collector/server.py
```diff
@@ -24,26 +43,54 @@ def error_handler(context, exc_info):
...
+        servers=available_servers,
+        http_pool_maxsize=4,
+        http_retry_timeout=40,
+        http_retry_delay=_http_retry_delay,
+    )
+    login_or_register(client, signer=LocalSigner(private_key=decode_hex(private_key)))
+
+    monitoring_room = join_global_room(
+        client=client,
+        name=MONITORING_BROADCASTING_ROOM,
```
andrevmatos: Notice this constant is just the suffix of the room name (the one that, on raiden, goes on `global_rooms` config and is accepted by `send_global`), e.g. `monitoring`, while the name accepted by `join_global_room` is the actual, qualified name, e.g. `raiden_ropsten_monitoring`. It's just a `'_'.join(['raiden', chain_name or chain_id, MONITORING_BROADCASTING_ROOM])`, but as the utils is more generic, it expects you to do this transform

andrevmatos: I'll split that into an utils as well, just to keep that synced

src/request_collector/cli.py
```diff
@@ -43,26 +42,9 @@ def setup_logging(log_level: str, log_config: TextIO):
...
-)
-@click.option(
-    '--matrix-homeserver',
-    default='https://transport01.raiden.network',
-    help='Matrix username',
-)
-@click.option(
-    '--matrix-username',
-    default=None,
+    '--private-key',
```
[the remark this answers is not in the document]
palango: Yeah, that's something I wanted to tackle after this PR. 
Regarding the example. Is it best practice to provide the password as a cli parameter?

src/request_collector/server.py
```diff
@@ -24,26 +43,54 @@ def error_handler(context, exc_info):
...
         self.task_list: List[gevent.Greenlet] = []
+        self.client, self.monitoring_room = setup_matrix(self.private_key)
+        self.monitoring_room.add_listener(self._handle_message, 'm.room.message')
 
     def _run(self):
         register_error_handler(error_handler)
-        self.transport.start()
 
         # this loop will wait until spawned greenlets complete
         while self.stop_event.is_set() is False:
```
andrevmatos: Instead of `gevent.sleep(1)` inside the while block with the `is_set` verification on the condition, I'd suggest `while self.stop_event.wait(1) is not True`, which will return immediatelly with `True` iff the event was set, instead of blocking up to 1sec if the event was set in the middle of a loop

src/request_collector/cli.py
```diff
@@ -1,5 +1,5 @@
-from gevent import monkey  # isort:skip # noqa
-monkey.patch_all()  # isort:skip # noqa
+# from gevent import monkey  # isort:skip # noqa
+# monkey.patch_all()  # isort:skip # noqa
```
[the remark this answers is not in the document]
palango: Do we actually need this for the `GMatrixClient` to work properly or not?

src/request_collector/cli.py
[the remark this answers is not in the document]
ulope: Well you can provide it on the command line, but by default the `click.password_option` is supposed to be used without argument and then shows an interactive no-echo prompt. Also in automated use I usually specify the password via an env variable (e.g. via a docker-compose `.env` file)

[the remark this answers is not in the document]
palango: Thanks. https://github.com/raiden-network/raiden-services/issues/97 tracks this.

src/request_collector/server.py
```diff
@@ -63,9 +63,10 @@ def _http_retry_delay() -> Iterable[float]:
     )
     login_or_register(client, signer=LocalSigner(private_key=decode_hex(private_key)))
 
+    room_name = '_'.join(['raiden', str(chain_id), MONITORING_BROADCASTING_ROOM])
```
andrevmatos: I'm implementing https://github.com/raiden-network/raiden/issues/3469

palango: Yeah, this included some other stuff to make the chain_id available.

andrevmatos: `chain_id` is not enough, you need also chain name, as per `raiden_contracts.constants`. https://github.com/raiden-network/raiden/pull/3471 takes care of everything

src/request_collector/server.py
```diff
@@ -56,25 +103,112 @@ def _run(self):
...
+                'Received message body not a string',
+                peer_user=user.user_id,
+                peer_address=to_checksum_address(peer_address),
+                room=room,
+            )
+            return False
+
+        messages: List[SignedMessage] = list()
+
+        if data.startswith('0x'):
```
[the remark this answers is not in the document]
palango: Thanks, added the correct branch.

src/request_collector/server.py
```diff
@@ -24,26 +44,77 @@ def error_handler(context, exc_info):
...
+
+    try:
+        login_or_register(client, signer=LocalSigner(private_key=decode_hex(private_key)))
+    except MatrixRequestError:
+        raise ConnectionError('Could not login/register to matrix.')
+    except ValueError:
+        raise ConnectionError('Could not login/register to matrix.')
+
+    try:
+        room_name = '_'.join(['raiden', str(chain_id), MONITORING_BROADCASTING_ROOM])
```
andrevmatos: https://github.com/raiden-network/raiden/pull/3471 is merged. Please, use that instead of this, as it also takes care of network name logic and so
```suggestion
        room_name = make_room_alias(chain_id, MONITORING_BROADCASTING_ROOM)
```

src/request_collector/cli.py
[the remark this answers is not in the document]
ulope: Yes, otherwise IO will block (e.g. in the requests calls to the matrix servers)

src/request_collector/server.py
palango: Wouldn't that change the idea here? Currently this only sleeps when no tasks are waiting, otherwise it waits for the next task to complete.
With your change (and my understanding, which might be wrong) this will always wait in the loop, even if there are tasks to be completed.

andrevmatos: Oh, right, that would, didn't notice you wanted to loop also through existing tasks. But anyway you could use `stop_event.wait` as a _gevent.sleep_ which returns on also on set and break when True

src/request_collector/server.py
```diff
@@ -56,25 +128,122 @@ def _run(self):
...
+                event['type'] != 'm.room.message' or
+                event['content']['msgtype'] != 'm.text' or
+                self.stop_event.ready()
+        ):
+            # Ignore non-messages and non-text messages
+            return False
+
+        sender_id = event['sender']
+
+        user = self.client.get_user(sender_id)
```
andrevmatos: I'd suggest implementing [_some_ caching](https://github.com/raiden-network/raiden/blob/1f54f8cc85d6b27029184cb602b5cd2c2395a88c/raiden/network/transport/matrix/transport.py#L1160-L1176) on this, as `client.get_user` is just a factory function, always creating new instances of `User` with empty displayname (and therefore, needing re-fetching it inside `validate_userid_signature`). `validate_userid_signature` caches the `recover` call, but needs both `user_id` and `displayname` to be correctly set, or it'll fetch it on each call. The complete user list (including displaynames) for a room comes on initial sync, so you may just get keep them in sync on `monitoring_room._members`, as done in raiden, or other caching methods you may find useful. Or not, and let it be ;)

src/request_collector/server.py
palango: So you say further down I'd rather write:
```py
            if len(tasks) == 0:
                self.stop_event.wait(1)
                continue
```

src/request_collector/server.py
palango: Thanks for the hint, I'll do this in a follow-up PR.

PR #99, opened 2019-02-12
Add test for MS crash handling

### LINKED ISSUE #32: Handle MS crashes while processing a block
If no special precautions are taken, some events might be handled zero or multiple times after restarting the MS.

Split off from #17.

Related to #63 

8 SP

karlb:
closes #32

### DIFF
--- a/src/monitoring_service/schema.sql
+++ b/src/monitoring_service/schema.sql
@@ -3,15 +3,10 @@ CREATE TABLE blockchain (
     receiver                        CHAR(42),
     token_network_registry_address  CHAR(42),
     monitor_contract_address        CHAR(42),
-    latest_known_block              INT
+    latest_known_block              INT NOT NULL DEFAULT -1
 );
-INSERT INTO blockchain VALUES (
-    NULL,
-    NULL,
-    NULL,
-    NULL,
-    NULL
-);
+INSERT INTO blockchain DEFAULT VALUES;
+
 
 CREATE TABLE token_network (
     address                 CHAR(42) PRIMARY KEY
--- a/tests/monitoring/monitoring_service/test_crash.py
+++ b/tests/monitoring/monitoring_service/test_crash.py
@@ -57,9 +57,10 @@ def test_crash(
     events = [
         [
             ReceiveChannelOpenedEvent(
-                token_network.address,
-                channel_identifier,
-                c1.address, c2.address,
+                token_network_address=token_network.address,
+                channel_identifier=channel_identifier,
+                participant1=c1.address,
+                participant2=c2.address,
                 settle_timeout=20,
                 block_number=0),
         ],
@@ -68,8 +69,8 @@ def test_crash(
         ],
         [
             ActionMonitoringTriggeredEvent(
-                token_network.address,
-                channel_identifier,
+                token_network_address=token_network.address,
+                channel_identifier=channel_identifier,
                 non_closing_participant=c2.address,
             ),
         ],
--- a/tests/monitoring/monitoring_service/test_crash.py
+++ b/tests/monitoring/monitoring_service/test_crash.py
@@ -1,0 +1,125 @@
+import os
+from typing import List
+from unittest.mock import Mock
+
+from monitoring_service.events import (
+    ActionMonitoringTriggeredEvent,
+    Event,
+    ReceiveChannelOpenedEvent,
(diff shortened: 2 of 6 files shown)


src/monitoring_service/schema.sql
```diff
@@ -3,15 +3,10 @@ CREATE TABLE blockchain (
...
-);
-INSERT INTO blockchain VALUES (
-    NULL,
-    NULL,
-    NULL,
-    NULL,
-    NULL
+    latest_known_block              INT NOT NULL DEFAULT -1
 );
+INSERT INTO blockchain DEFAULT VALUES;
```
palango: The default for `CHAR(42)`  is `NULL`?

karlb: The default for everything is `NULL` unless specified differently.

tests/monitoring/monitoring_service/test_crash.py
```diff
@@ -0,0 +1,125 @@
...
+            locked_amount=0,
+            locksroot='0x00',
+        ),
+        reward_amount=0,
+    )
+
+    events = [
+        [
+            ReceiveChannelOpenedEvent(
+                token_network.address,
```
palango: Would be nice to add keywords to all parameters

tests/monitoring/monitoring_service/test_crash.py
```diff
@@ -0,0 +1,125 @@
...
+            nonce=1,
+            additional_hash='0x11',
+            transferred_amount=2,
+            locked_amount=0,
+            locksroot='0x00',
+        ),
+        reward_amount=0,
+    )
+
+    events = [
```
palango: Why is this a list of a list? Do you plan to extend it to have multiple items in each sublist? On the other hand it gets flattened in `MockBlockchainListener` anyways.

tests/monitoring/monitoring_service/test_crash.py
```diff
@@ -0,0 +1,125 @@
...
+        )
+        ms.bcl = MockBlockchainListener(events)  # type: ignore
+        msc = Mock()
+        ms.context.monitoring_service_contract = msc
+        ms.monitor_mock = msc.functions.monitor.return_value.transact  # type:ignore
+        ms.monitor_mock.return_value = bytes(0)  # type:ignore
+        return ms
+
+    # initialize both monitoring services
+    stable_ms = new_ms('stable.db')
```
palango: Should we add some random part here? Not sure how equal filenames on tmp are handled.

tests/monitoring/monitoring_service/test_crash.py
```diff
@@ -0,0 +1,125 @@
...
+    crashy_ms = new_ms('crashy.db')
+    for ms in [stable_ms, crashy_ms]:
+        ms.database.conn.execute(
+            "INSERT INTO token_network(address) VALUES (?)",
+            [token_network.address],
+        )
+        ms.context.ms_state.blockchain_state.token_network_addresses = [token_network.address]
+        ms.database.upsert_monitor_request(monitor_request)
+        ms.database.conn.commit()
```
palango: General question: Do we have to call commit after every db change, or at least in the `updated_head_block_event_handler`?

tests/monitoring/monitoring_service/test_crash.py
karlb: I constructed it to be a list of blocks, so that each block can have multiple events. This might come in handy when a problem only occurs when two events happen in the same block. It's a bit over-engineered for the current use case, but it's also somehow nice that it's similar to the structure of a real blockchain.

tests/monitoring/monitoring_service/test_crash.py
karlb: The `tmpdir` is a fresh clean dir for every test invocation. No necessity for random file names. See https://docs.pytest.org/en/latest/tmpdir.html

tests/monitoring/monitoring_service/test_crash.py
karlb: With our current settings, all changes which are not explicitly committed are lost on crash/shutdown. The best way to handle a transaction is to use the connection as a context manager, which automatically commits on successful exit and rolls back when an exception is thrown.

As long as we're working synchronously, we could also choose to wrap larger sections of code into a transaction to build larger sets of atomic changes.

PR #101, opened 2019-02-14
Adds price inforamation and chain-id to info endpoint

Dominik1999:


src/monitoring_service/database.py
```diff
@@ -1,9 +1,9 @@
+import dataclasses
```
palango: Don't know why this happens, seems to be a problem on travis. But for now you have to revert  this in order to make the linting check pass.

tests/pathfinding/test_rest.py
```diff
@@ -161,15 +160,20 @@ def test_get_paths(
...
     assert response.json() == {
-        'ip': socket.gethostbyname(socket.gethostname()),
+        'price_info': 0,
+        'network_info': {
+            'chain_id': pathfinding_service_full_mock.chain_id,
+            'registry_address': pathfinding_service_full_mock.registry_address,
+        },
         'settings': 'PLACEHOLDER FOR PATHFINDER SETTINGS',
-        'version': pkg_resources.require('raiden-services')[0].version,
+        'version': '0.0.1',
```
palango: Why did you change this? I think it was correct before.

PR #102, opened 2019-02-14
Use keystore file instead of raw private key

### LINKED ISSUE #97: Use keyfile in CLI arg instead of raw private key
## Problem Definition

Currently a raw private key has to be provided to the MS as a command line argument. This should be switched to a keyfile and an optional password.

See https://github.com/raiden-network/raiden-services/pull/95#discussion_r255956816

palango:
Fixes #97

### DIFF
--- a/tests/monitoring/fixtures/server.py
+++ b/tests/monitoring/fixtures/server.py
@@ -1,5 +1,6 @@
 import json
 import logging
+from typing import List
 
 import pytest
 from eth_account import Account
@@ -37,6 +38,15 @@ def keystore_file(tmp_path) -> str:
         json.dump(keystore_json, fp)
 
     return keystore_file
+
+
+@pytest.fixture
+def default_cli_args(keystore_file) -> List[str]:
+    return [
+        '--keystore-file', keystore_file,
+        '--password', KEYSTORE_PASSWORD,
+        '--state-db', ':memory:',
+    ]
 
 
 @pytest.fixture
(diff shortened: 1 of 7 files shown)


tests/monitoring/fixtures/server.py
```diff
@@ -20,6 +24,21 @@ def server_private_key(get_random_privkey, ethereum_tester):
     return key
 
 
+@pytest.fixture
+def keystore_file(tmp_path) -> str:
+    keystore_file = tmp_path / KEYSTORE_FILE_NAME
```
karlb: Nice syntax. I'll start using that, too.

tests/monitoring/monitoring_service/test_cli.py
```diff
@@ -11,35 +12,59 @@
...
-def test_success():
-    """ Calling the request_collector with default args should succeed after heavy mocking """
+def test_success(keystore_file):
+    """ Calling the monitoring_service with default args should succeed after heavy mocking """
     runner = CliRunner()
     with patch.multiple(**patch_args):
-        result = runner.invoke(main, DEFAULT_ARGS, catch_exceptions=False)
+        result = runner.invoke(
+            main,
+            DEFAULT_ARGS + ['--keystore-file', keystore_file],
```
karlb: `'--keystore-file', keystore_file` could be added to the default args if we made the default args a fixture. Not a big improvement right now, but at least if we get more complicated default args we should do it instead of repeating more args across the tests.

palango: True that! Fixed.

PR #103, opened 2019-02-14
Verify MR signatures on creation

karlb:
With this change, all `MonitorRequest` objects contain valid signatures. I like how the API turned out and might split `HashedBalanceProof`s into signed and unsigned classes too, if there is general agreement that this is a nice way to handle signing.

### DIFF
--- a/tests/monitoring/fixtures/monitor_request.py
+++ b/tests/monitoring/fixtures/monitor_request.py
@@ -1,7 +1,7 @@
 import pytest
 from eth_utils import encode_hex
 
-from monitoring_service.states import Channel, MonitorRequest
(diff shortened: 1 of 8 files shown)


tests/monitoring/fixtures/monitor_request.py
```diff
@@ -43,32 +43,27 @@ def f(
             privkey = keys[1]
             privkey_non_closing = keys[0]
 
-        monitor_request = MonitorRequest(
+        bp = HashedBalanceProof(  # type: ignore
```
palango: Why do we need the `type: ignore` stuff here?

karlb: Because I could get mypy to understand the optional `priv_key` parameter. I don't really want to ignore it.

Issue #84, opened 2019-02-07
Manage connection pool better


palango:
## Problem Definition

When running the PFS on a registry with many registered token it logs many errors like the following:
```
02-07 15:57:49 - urllib3.connectionpool - WARNING - Connection pool is full, discarding connection: parity.kovan.ethnodes.brainbot.com
```

## Actual Outcome

A lot of connections are discarded.

## Expected Outcome

The connection pool should be managed better or increased in size.

schmir:
We've ran into the same problem in the trustlines project [1]. We're interested in any insights. It would be nice  if you could write them down here. 

[1]  https://github.com/trustlines-network/relay/issues/246 . Though we didn't really investigate how to solve it.

hackaugusto:
This is not a problem at all, the pool is an optimization to reuse sockets. New connections are still made but not added to the pool if the capacity has been reached (however, this is configurable).

The point is, either you don't benefit from the pool because you need many connections, or there is a piece of code that opens too many short lived connections in a short life span. This is what happened with https://github.com/raiden-network/raiden/issues/1847

palango:
The warnings have been suppressed in #104 . We still need to investigate and come to a conclusion on this issue.

palango:
This doesn't happen any more with #230

PR #104, opened 2019-02-14
Suppress urllib3 connectionpool warnings

palango:
When starting to test we don't want to have all this noise hiding the message.

Related to #84


karlb:
Can't we get rid of these messages by using the connection pool in a better way? I'd like to understand why we get the warnings in the first place before suppressing them. Can you explain why?

palango:
I saw this more as a temporary measure towards #84 as I just wanted to get rid of them for know. But Augusto has some comments in #84 as well.

PR #105, opened 2019-02-14
Emit the UpdatedHeadBlockEvent on every block update

palango:
This commit fixes a bug that led to the UpdatedHeadBlockEvent only being
emitted every second block. The reason was a wrong comparison in the
blockchain filter logic.


palango:
Nah, it felt better, but your argument makes sense. Will change it back.

PR #106, opened 2019-02-14
Split validation of MR between saving and usage

### LINKED ISSUE #12: Add checks of MonitorRequest message data fields
Before using any MonitorRequest, MS must rigorously validate fields of the message, in particular

- [x] address of the sender
- [x] signature(s) must match the sender
- [x] existence of contract submitted in `token_network_address`  field

Since we allow MRs to be stored before a ChannelOpened event has been confirmed, we must defer the validation to the time when the MR is used. 

Postponed (will be done as part of the payment milestone):
- MSC balance of the sender (there must be enough tokens to pay out the reward)

karlb:
This has become necessary as we want to store MRs before we have the information required to fully validate them.

Closes #12.


src/monitoring_service/handlers.py
```diff
@@ -305,6 +310,21 @@ def updated_head_block_event_handler(event: Event, context: Context) -> None:
     context.db.update_state(context.ms_state)
 
 
+def _is_mr_valid(monitor_request: MonitorRequest, channel: Channel) -> bool:
+    if (
+        monitor_request.signer not in channel.participants or
+        monitor_request.non_closing_signer not in channel.participants
+    ):
+        log.info('MR signed by unknown party')
```
palango: Can you add some more context to these logs.

PR #110, opened 2019-02-15
Remove default contract addresses and make them required

palango:
These need to be set properly.


palango:
Yes, the problem is that raiden currently uses the 0.3 or 0.4 contracts, but we need the 0.5 ones for the service stuff.

PR #113, opened 2019-02-15
Start the GMatrixClients event loop

palango:
We need to start the matrix listeners event loop.

However, this solutions still feels rough and probably can be improved.

### DIFF
--- a/src/request_collector/server.py
+++ b/src/request_collector/server.py
@@ -1,34 +1,14 @@
-import json
 import sys
 import traceback
-from typing import Iterable, List, Tuple
 
 import gevent
 import structlog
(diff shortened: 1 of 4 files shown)


src/request_collector/server.py
```diff
@@ -45,47 +24,6 @@ def error_handler(_context, exc_info):
...
-        )
-    except MatrixRequestError:
-        raise ConnectionError('Could not join monitoring broadcasting room.')
-    except TransportError:
-        raise ConnectionError('Could not join monitoring broadcasting room.')
-
-    return client, monitoring_room
-
-
 class RequestCollector(gevent.Greenlet):
```
karlb: Is this still a real Greenlet? It doesn't do anything itself in it's `run` method.

PR #119, opened 2019-02-18
Checks MS's account balance, throws warning if below estimate

### LINKED ISSUE #44: MS should check account balance at startup
## Problem Definition

The MS should check the account balance during startup and show an error when it's to low to fulfill its monitoring duties.

Dominik1999:
closes #44


src/monitoring_service/cli.py
```diff
@@ -53,6 +56,30 @@ def setup_logging(log_level: str) -> None:
...
+                "WARNING\n"
+                "Your account's balance is below the estimated gas reserve of "
+                f"{estimated_required_balance_eth} Eth. You will be be unable"
+                "to perform on-chain transactions and cannot monitor any channels."
+                "Please add funds to your account as soon as possible."
+            ),
+            fg='red',
+        )
+        return False
+    else:
```
BOR4: nitpicky. I think in rest of codebase you guys avoid using else where not needed and I believe here it is not.

Dominik1999: Good point Thanks!

LefterisJP: @BOR4 always having your eyes open :)

BOR4: cheers guys!

src/monitoring_service/cli.py
```diff
@@ -53,6 +56,30 @@ def setup_logging(log_level: str) -> None:
...
 
+def check_gas_reserve(web3: Web3, private_key: str) -> bool:
+    """ Check periodically for gas reserve in the account """
+    gas_price = web3.eth.gasPrice
+    gas_limit = GAS_REQUIRED_FOR_MS_MONITOR
+    estimated_required_balance = gas_limit * gas_price * 10
+    estimated_required_balance_eth = Web3.fromWei(estimated_required_balance, 'ether')
+    current_balance = web3.getBalance(private_key_to_address(private_key))
+    if current_balance < estimated_required_balance:
+        log.info('Missing gas reserve', required_wei=estimated_required_balance)
```
palango: it says wei here, but you converted it to eth before.

src/monitoring_service/cli.py
```diff
@@ -53,6 +56,30 @@ def setup_logging(log_level: str) -> None:
     )
 
 
+def check_gas_reserve(web3: Web3, private_key: str) -> bool:
+    """ Check periodically for gas reserve in the account """
+    gas_price = web3.eth.gasPrice
+    gas_limit = GAS_REQUIRED_FOR_MS_MONITOR
```
palango: I don't see the need to this variable, you can just inline this.

src/monitoring_service/cli.py
```diff
@@ -53,6 +56,30 @@ def setup_logging(log_level: str) -> None:
...
+    estimated_required_balance = gas_limit * gas_price * 10
```
palango: Can wyou make the 10 a variable?

src/monitoring_service/cli.py
Dominik1999: I tried, but the lime becomes too long when factor 10 is introduced as a default variable. Looks like this now:
   ` estimated_required_balance = gas_limit * gas_price * DEFAULT_GAS_BUFFER_FACTOR`

src/monitoring_service/cli.py
Dominik1999: done, added it to the constants as `DEFAULT_GAS_BUFFER_FACTOR`

src/monitoring_service/cli.py
Dominik1999: Right, thanks. In Raiden, they log the WEI value but the show the ETH value to the user. Should we change or follow their approach?

palango: Ah, you're right. I missed that. Never mind then.

src/monitoring_service/cli.py
```diff
@@ -129,7 +156,8 @@ def main(
         required_confirmations=confirmations,
         db_filename=state_db,
     )
-
+    while gevent.spawn(check_gas_reserve(web3, private_key)) is True:
```
palango: This looks really strange. Are you sure it does what you intent?

tests/monitoring/monitoring_service/test_cli.py
```diff
@@ -36,6 +36,11 @@ def test_wrong_password(keystore_file, default_cli_args_ms):
     assert result.exit_code == 1
 
 
+def test_check_gas(web3):
+    private_key = '3a1076bf45ab87712ad64ccb3b10217737f7faacbf2872e88fdd9a537d8fe266'
+    check_gas_reserve(web3, private_key)
```
palango: this is not checking anything.

PR #120, opened 2019-02-18
Compare nonce when accepting new MR

### LINKED ISSUE #115: RequestMonitor should only save MR when nonce increases
## Problem Definition

Currently it seems like the request monitor saves a received MR, even when the nonce is smaller or equal to a currently know one.

## Actual Outcome

Every MR is written to the DB.

## Expected Outcome

Only MRs with higher nonce should be written to the DB.

karlb:
Also fix some bytes/hex encoding issues.

Closes #115.

### DIFF
--- a/src/monitoring_service/states.py
+++ b/src/monitoring_service/states.py
@@ -253,6 +253,7 @@ class MonitorRequest(UnsignedMonitorRequest):
 
     def __post_init__(self) -> None:
         super(MonitorRequest, self).__post_init__()
+        self.token_network_address = to_checksum_address(self.token_network_address)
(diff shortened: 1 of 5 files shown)


src/monitoring_service/states.py
```diff
@@ -253,6 +253,7 @@ class MonitorRequest(UnsignedMonitorRequest):
 
     def __post_init__(self) -> None:
         super(MonitorRequest, self).__post_init__()
+        self.token_network_address = to_checksum_address(self.token_network_address)
```
palango: Please make this a check

PR #125, opened 2019-02-19
Limit amount of blocks processed per iteration

palango:
The first sync led to some request timeouts on slow networks.


palango:
Moved it out.

PR #130, opened 2019-02-21
Add more details to channel event logging

palango:


src/pathfinding_service/service.py
```diff
@@ -127,10 +127,6 @@ def _get_token_network(self, token_network_address: Address) -> Optional[TokenNe
         else:
             return self.token_networks[token_network_address]
 
-    def _check_chain_id(self, received_chain_id: int):
-        if not received_chain_id == self.chain_id:
-            raise ValueError('Chain id does not match')
-
     def handle_channel_event(self, event: Dict):
         event_name = event['event']
```
karlb: We could also do
```python
    log.info('Received event', name=event['event'], args=event['args'])
```
here instead of printing nearly every arg in every case separately. But we don't have that many cases, so the individual handling is ok and gives us a bit more flexibility to print the messages more nicely.

palango: Good point, but I think it's clearer for now.

PR #131, opened 2019-02-21
Adds periodic balance check for monitoring service

### LINKED ISSUE #44: MS should check account balance at startup

Dominik1999:
Closes #44


src/monitoring_service/cli.py
```diff
@@ -4,20 +4,55 @@
...
+    """ Check periodically for gas reserve in the account """
+    gas_price = web3.eth.gasPrice
+    gas_limit = GAS_REQUIRED_FOR_MS_MONITOR
+    estimated_required_balance = gas_limit * gas_price * DEFAULT_GAS_BUFFER_FACTOR
+    estimated_required_balance_eth = Web3.fromWei(estimated_required_balance, 'ether')
+    current_balance = web3.eth.getBalance(private_key_to_address(private_key))
+    if current_balance < estimated_required_balance:
+        log.info('Missing gas reserve', required_wei=estimated_required_balance)
+        log.error(
+            (
```
palango: you can put this in the last line

src/monitoring_service/cli.py
```diff
@@ -4,20 +4,55 @@
...
+                "WARNING\n"
```
palango: no need for this, the logger will print that this is an error

src/monitoring_service/cli.py
```diff
@@ -4,20 +4,55 @@
...
+                "Your account's balance is below the estimated gas reserve of "
+                f"{estimated_required_balance_eth} Eth. You will be be unable"
```
palango: space missing at the end.

src/monitoring_service/cli.py
```diff
@@ -4,20 +4,55 @@
...
+                "to perform on-chain transactions and cannot monitor any channels."
```
palango: space missing at the end.

tests/monitoring/monitoring_service/test_cli.py
```diff
@@ -36,6 +36,11 @@ def test_wrong_password(keystore_file, default_cli_args_ms):
     assert result.exit_code == 1
 
 
+def test_check_gas(web3):
+    private_key = '3a1076bf45ab87712ad64ccb3b10217737f7faacbf2872e88fdd9a537d8fe266'
+    check_gas_reserve(web3, private_key)
```
palango: this isn't checking anything. you should at least assert that the warning message is displayed.

src/monitoring_service/cli.py
Dominik1999: I am not sure if I understand. The `log.error` is already in the last line of that function, right?

src/monitoring_service/cli.py
Dominik1999: thanks, done

src/monitoring_service/cli.py
Dominik1999: thank, added

src/monitoring_service/cli.py
Dominik1999: thank, added

src/monitoring_service/cli.py
palango: you can just move up the `(` in the last liune

tests/monitoring/monitoring_service/test_cli.py
Dominik1999: I see. I can't assert the log message since it is changing all the time according to the current gas estimate. Should we just remove that test or maybe test only that the log level is 'Warning'

src/monitoring_service/cli.py
Dominik1999:
```
        log.error(
            (
                "Your account's balance is below the estimated gas reserve of "
                f"{estimated_required_balance_eth} Eth. You will be be unable "
                "to perform on-chain transactions and cannot monitor any channels. "
                "Please add funds to your account as soon as possible."
            ),
            fg='red',)
```

Dominik1999: sorry ich steh grad auf dem Schlauch tbh ...

palango: exactly what I meant, even though i wonder if it's actually better now that I see it.

tests/monitoring/monitoring_service/test_cli.py
palango: You can also only assert only on a part or create private key for an address with a fixed amount of eth.

Dominik1999: Yes good idea, but the gas_estimate always changes independently from the current balance. This gas estimate is then shown in the message. 
I could assert the len of the log message though, I will try

palango: Good point. For the log message see https://docs.pytest.org/en/latest/reference.html#caplog

Dominik1999: skipped the test, for now, fails with my setup at the moment.

src/monitoring_service/cli.py

src/monitoring_service/cli.py
```diff
@@ -4,20 +4,53 @@
...
+    current_balance = web3.eth.getBalance(private_key_to_address(private_key))
+    if current_balance < estimated_required_balance:
+        log.info('Missing gas reserve', required_wei=estimated_required_balance)
+        log.error((
+            "Your account's balance is below the estimated gas reserve of "
+            f"{estimated_required_balance_eth} Eth. You will be be unable "
+            "to perform on-chain transactions and cannot monitor any channels. "
+            "Please add funds to your account as soon as possible."
+        ),
+            fg='red',
```
palango: this is not needed when using logging

src/monitoring_service/cli.py
```diff
@@ -4,20 +4,53 @@
...
 
+def check_gas_reserve(web3: Web3, private_key: str) -> None:
+    """ Check periodically for gas reserve in the account """
+    gas_price = web3.eth.gasPrice
+    gas_limit = GAS_REQUIRED_FOR_MS_MONITOR
+    estimated_required_balance = gas_limit * gas_price * DEFAULT_GAS_BUFFER_FACTOR
+    estimated_required_balance_eth = Web3.fromWei(estimated_required_balance, 'ether')
+    current_balance = web3.eth.getBalance(private_key_to_address(private_key))
+    if current_balance < estimated_required_balance:
+        log.info('Missing gas reserve', required_wei=estimated_required_balance)
```
palango: why have this here when we log again afterwards

Dominik1999: deleted

src/monitoring_service/cli.py
Dominik1999: agreed

PR #132, opened 2019-02-21
Use POST for path requests in preparation of fees

karlb:


tests/pathfinding/test_rest.py
```diff
@@ -23,57 +23,45 @@ def test_get_paths_validation(
...
+    default_params = {
+        'from': initiator_address,
+        'to': target_address,
+        'value': 5,
+        'max_paths': 3,
+    }
 
-    url = base_url
-    response = requests.get(url)
+    def get_path(**kwargs):
```
palango: a name like `request_path_with` would fit the function better in my eyes.

src/pathfinding_service/api/rest.py
```diff
@@ -79,7 +79,7 @@ def _validate_args(args):
         return None
 
     # url parameters are used because json bodies for GET requests are uncommon
```
BOR4: does this comment make sense now that request is POST?

karlb: You're right, it doesn't.

PR #133, opened 2019-02-21
adds a test to balance check

Dominik1999:


tests/monitoring/monitoring_service/test_cli.py
```diff
@@ -12,6 +12,13 @@
 )
 
 
+def test_account_check(web3):
+    private_key = '0F951D6EAF7685D420AACCA3900127E669892FE5CA6C8E4C572A59B0609AAE6B'
+    check_gas_reserve(web3, private_key)
+    # This test assers the logging level error
+    assert logging.getLevelName(40)
```
karlb: You don't actually test for any log output here, the assert is always true:
```
>>> import logging
>>> logging.getLevelName(40)
'ERROR'
```

See https://docs.python.org/3/library/logging.html#logging.getLevelName

You might want to use `pytest.caplog` (https://docs.pytest.org/en/latest/logging.html#caplog-fixture) to have a look at the log output.

Dominik1999: Yes, I dediced to not assert for the output, since the output always changes. It logs the current gas price estimate of the repective chain the MS runs on. So I just wanted to log the error. Does that make sense?

Dominik1999: ah ok, but when this is always true, the test does not make sense. I will check

Dominik1999: changed the test to a more meaningful one

Issue #108, opened 2019-02-15
Unify logging


palango:
## Problem Definition

Currently we use the standard lib `logging` and `structlog` for different parts. We should decide on one tool and use it consistently. (keep dependencies in mind, the imported matrix stuff uses structlog)

## Tasks
- [x] Decide on whether `logging` or `structlog`
- [x] Port all subprojects to the decided logging lib.

karlb:
Testing structlog for logging with `pytest.caplog` does not work. Use `pytest.capsys` or https://github.com/wimglenn/pytest-structlog, instead.

palango:
Good point, I|ll have to fix this.

PR #135, opened 2019-02-25
Fix cli tests with structlog

palango:
Fixes https://github.com/raiden-network/raiden-services/issues/108

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -16,6 +16,7 @@ mypy==0.660
 pytest>=4.0.1
 pytest-runner
 pytest-cov
+pytest-structlog==0.1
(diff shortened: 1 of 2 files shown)


karlb:
Why didn't the test fail before this change?

palango:
> Why didn't the test fail before this change?

`caplog` works after all, it just ugly:

```
'\x1b[2m2019-02-25 17:21:15.879212\x1b[0m [\x1b[31m\x1b[1merror    \x1b[0m] \x1b[1mCan not connect to the Ethereum client. Please check that it is running and that your settings are correct.\x1b[0m [\x1b[34m\x1b[1mpathfinding_service.cli\x1b[0m] '
```

PR #138, opened 2019-02-26
run gas check inside the main loop

palango:
@Dominik1999 Small fix for the issue we found in the debugging session.


src/monitoring_service/cli.py
```diff
@@ -154,9 +154,15 @@ def main(
...
 
     ms.start()
-    gevent.spawn(check_gas, web3, private_key)
+    check_gas_task = gevent.spawn(check_gas, web3, private_key)
+
+    # this blocks until the service is stopped
     ms.start(gevent.sleep)
 
+    # stop the remaining task
+    check_gas_task.kill()
```
karlb: Do we kill the Greenlet right after starting the MS? In that case, it would only be executed once (because the first execution is scheduled before the kill).

palango:
I updated this to the in-loop approach we discussed. Please have a look again. We could make the default `false` for running the check.

src/monitoring_service/service.py
```diff
@@ -82,9 +104,24 @@ def __init__(
...
+        wait_function: Callable = time.sleep,
+        check_account_gas_reserve: bool = True,
+    ) -> None:
         while True:
             last_confirmed_block = self.web3.eth.blockNumber - self.required_confirmations
+
+            # check gas reserve
+            do_gas_reserve_check = (
+                check_account_gas_reserve and
+                last_confirmed_block % DEFAULT_GAS_CHECK_BLOCKS == 0 and
```
karlb: We can process multiple blocks at once, so this might be skipped. Why not use
```
last_confirmed_block >= self.last_gas_check_block + DEFAULT_GAS_CHECK_BLOCKS
```

Dominik1999:
I found another solution together with Ulo

```
def check_gas_reserve(web3: Web3, private_key: str) -> None:
    """ Check periodically for gas reserve in the account """
    gas_price = web3.eth.gasPrice
    gas_limit = GAS_REQUIRED_FOR_MS_MONITOR
    estimated_required_balance = gas_limit * gas_price * DEFAULT_GAS_BUFFER_FACTOR
    estimated_required_balance_eth = Web3.fromWei(estimated_required_balance, 'ether')
    current_balance = web3.eth.getBalance(private_key_to_address(private_key))
    if current_balance < estimated_required_balance:
        log.error(
            "Your account's balance is below the estimated gas reserve of "
            f"{estimated_required_balance_eth} Eth. You will be be unable "
            "to perform on-chain transactions and cannot monitor any channels. "
            "Please add funds to your account as soon as possible.",
        )


def check_gas(stop_event: Event, web3: Web3, private_key: str) -> None:
    while not stop_event.is_set():
        check_gas_reserve(web3, private_key)
        gevent.sleep(DEFAULT_GAS_CHECK_WAIT)


def main():
[...]
   gas_check_stop_event = Event()
    ms.start()
    gas_check = gevent.spawn(check_gas, gas_check_stop_event, web3, private_key)
    ms.start(gevent.sleep)
    gas_check_stop_event.set()

```

Dominik1999:
I could push my solution for comparison?

palango:
I'm not sure which approach we should go for.
- The gevent one is noce because it decouples the two things but adds some synchronization overhead
- The in-loop approach is easier to reason about, but mixes the domains

karlb:
While I agree with @palango's reasoning, I'm in favor of the gevent-free solution. We already lost too much time debugging gevent problems and I don't trust us to use it well enough that it won't happen again too soon.

Dominik1999:
perfect let's get rid of it. But we use gevent in the pfs anyway, right? The matrix listener is a gevent greenlet?

palango:
Ok, added a commit with Karls remarks.

> perfect let's get rid of it. But we use gevent in the pfs anyway, right? The matrix listener is a gevent greenlet?

That's true, but we try(ed) to keep the usage in the MS at a minimum (we might need at some point though).

karlb:
> perfect let's get rid of it. But we use gevent in the pfs anyway, right? The matrix listener is a gevent greenlet?

Yes and yes. We probably won't be able to remove it entirely, but the more we can remove it, the better. It might even make it possible to use other concurrency approaches in the future if most of the code base does not deal with gevent.

PR #143, opened 2019-02-27
Working version of matrix integration pfs

Dominik1999:


src/pathfinding_service/cli.py
```diff
@@ -150,6 +176,8 @@ def main(
             registry_address=registry_address,
             sync_start_block=start_block,
             required_confirmations=confirmations,
+            private_key=private_key,
+            poll_interval=10,
```
palango: I'd like to not define this here or use a default

src/pathfinding_service/service.py
```diff
@@ -13,13 +13,17 @@
     create_channel_event_topics,
     create_registry_event_topics,
 )
+from raiden.constants import PATH_FINDING_BROADCASTING_ROOM
+# from matrix_client.errors import MatrixRequestError
```
palango: what about this

src/pathfinding_service/service.py
```diff
@@ -100,16 +121,17 @@ def _setup_token_networks(self):
...
         self.token_network_registry_listener.start()
-
         self.is_running.wait()
 
     def stop(self):
         self.token_network_registry_listener.stop()
         for task in self.token_network_listeners:
             task.stop()
         self.is_running.set()
+        self.matrix_listener.stop()
```
palango: I think this should go before the  `self.is_running.set()`

src/pathfinding_service/service.py
```diff
@@ -253,3 +275,13 @@ def create_token_network_for_address(
...
+
+    def handle_message(self, message: SignedMessage):
+        if isinstance(message, UpdatePFS):
+            self.on_pfs_update(message, UpdatePFS)
+            # TODO validate message
+        else:
+            log.info('Ignoring unknown message type')
+
+    def on_pfs_update(self, message, UpdatePFS):
+        print(f"das ist die message: {message}")
```
palango: no german please, and better make this a log message.

src/pathfinding_service/service.py
```diff
@@ -253,3 +275,13 @@ def create_token_network_for_address(
...
         self.token_network_listeners.append(token_network_listener)
+
+    def handle_message(self, message: SignedMessage):
+        if isinstance(message, UpdatePFS):
+            self.on_pfs_update(message, UpdatePFS)
+            # TODO validate message
+        else:
+            log.info('Ignoring unknown message type')
+
+    def on_pfs_update(self, message, UpdatePFS):
```
palango: this should be `message: UpdatePFS`, right?

tests/pathfinding/conftest.py
```diff
@@ -44,12 +44,17 @@ def no_greenlets_left():
     yield
     tasks = _get_running_greenlets()
     # give all tasks the chance to clean themselves up
+    for task in tasks:
```
palango: Puh, this stuff is ugly

tests/pathfinding/test_blockchain_integration.py
```diff
@@ -38,12 +38,13 @@ def test_pfs_with_mocked_client(
...
         required_confirmations=1,
         poll_interval=0,
+        private_key='3a1076bf45ab87712ad64ccb3b10217737f7faacbf2872e88fdd9a537d8fe266',
     )
 
     # greenlet needs to be started and context switched to
     pfs.start()
     wait_for_blocks(1)
-    gevent.sleep(0)
+    gevent.sleep(1)
```
palango: Is this still necessary after the other changes?

tests/pathfinding/test_cli.py
```diff
@@ -16,42 +16,54 @@
...
-def test_success():
+def test_success(default_cli_args):
     """ Calling the pathfinding_service with default args should succeed after heavy mocking """
     runner = CliRunner()
     with patch.multiple(**patch_args) as mocks:
         mocks['get_default_registry_and_start_block'].return_value = Mock(), Mock()
-        result = runner.invoke(main, [])
+        result = runner.invoke(
+            main,
+            default_cli_args + [],
```
palango: the `+ []` doesn't do anything

tests/pathfinding/test_rest.py
```diff
@@ -64,6 +66,11 @@ def request_path_with(**kwargs):
     response = request_path_with(max_paths=-1)
     assert response.json()['errors'] == 'Number of paths must be positive: -1'
 
+    # killen aller greenlets
```
palango: english comments, and we have to find a better solution for this. I can have a look.

palango:
Just a comment: this will also break the deployment. But this is no reason to not merge this.

src/pathfinding_service/cli.py
Dominik1999: I defined a `DEFAULT_POLL_INTERVALL = 10` in the pfs config

src/pathfinding_service/service.py
Dominik1999: this was meant for the next debugging release. deleted it for now

src/pathfinding_service/service.py
Dominik1999: changed to English, will be deleted soon. This is just for debugging

src/pathfinding_service/service.py
Dominik1999: yes, changed it

tests/pathfinding/conftest.py
Dominik1999: yes it is, let's change it next week with Ulo :)

tests/pathfinding/test_cli.py
Dominik1999: True, changed it

tests/pathfinding/test_blockchain_integration.py
Dominik1999: tests fail when it is removed. So it seems necessary. I suggest to keep it

src/raiden_libs/matrix.py
```diff
@@ -56,12 +57,12 @@ def __init__(
 
     def _run(self):
         self.client.start_listener_thread()
-        self.client.sync_thread.get().wait()
```
palango: Are you sure about this? I am not sure about the effect of this

Dominik1999: Ehm it works like that. But I don't really know what it does either. I will put it on the open questions list

PR #152, opened 2019-03-11
Add Continuous Deployment code to this repo

### LINKED ISSUE #22: Continuous deployment

karlb:
Merge https://github.com/palango/services-dev into this repo.

* add containers for MS and RC
* use single image for all containers
* provide `docker-compose.override.yml` for local testing
* reuse shared args via yaml merge command
* use keystore (which is now required for all services)

Closes #22

### DIFF
--- a/deployment/.gitignore
+++ b/deployment/.gitignore
@@ -1,0 +1,6 @@
+venv/
+.mypy_cache/
+
+*.pyc
+
+.vscode/
--- a/deployment/builder/app/main.py
+++ b/deployment/builder/app/main.py
@@ -65,7 +65,7 @@ def build(branch, container_names, source, deployment, **kw):
             _print(f'docker down: {name}')
             subprocess.check_output(["docker-compose", "stop", name])
         _print('docker up')
-        subprocess.check_output(["docker-compose", "up", "-d"])
+        subprocess.check_output(["docker-compose", "-f", "docker-compose.yml", "up", "-d"])
     except Exception as e:
         _print(str(e))
         return False
--- a/deployment/builder/app/main.py
+++ b/deployment/builder/app/main.py
@@ -1,0 +1,72 @@
+import os
+import subprocess
+import sys
+from pprint import pprint
+
+from flask import Flask, request
+
+
+REPOS = {
+    'raiden-network/raiden-services': {
+        'master': {
+            'source': '/data/raiden-services',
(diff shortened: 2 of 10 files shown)


deployment/.gitignore
```diff
@@ -0,0 +1,6 @@
+venv/
```
palango: should we have this file here?

deployment/builder/app/main.py
```diff
@@ -0,0 +1,72 @@
...
+        os.chdir(source)
+        _print('git fetch')
+        subprocess.check_output(["git", "fetch", "--all"])
+        _print('git reset')
+        subprocess.check_output(["git", "reset", "--hard", f"origin/{branch}"])
+
+        _print(f'Switching to {deployment}')
+        os.chdir(deployment)
+        _print('docker build')
+        subprocess.check_output(["docker-compose", "build"])
```
palango: Do we need to update this?

palango: nvm

deployment/docker-compose.yml
```diff
@@ -0,0 +1,121 @@
...
+  ms-rinkeby:
+    << : *defaults
+    command: ["python3", "-m", "monitoring_service.cli"]
+    environment:
+      - MS_ETH_RPC=http://geth.rinkeby.ethnodes.brainbot.com:8545
+      - MS_REGISTRY_ADDRESS=0xE4769320245E4DeEAE5e816C9A827098B2988ef8
+      - MS_MONITOR_CONTRACT_ADDRESS=0xA0ED03b276cA581e8F120b5DcFe152DaAb0162d9
+      - MS_STATE_DB=/state/ms-rinkeby.db
+
+  # TODO: wrong addresses!
```
palango: We should do this automatically, but will this even start?

karlb: Yes, we will only get problems once we do transactions on those contracts.

deployment/.gitignore
karlb: No, that only made sense in the separate repo.

PR #155, opened 2019-03-11
Check transactions receipts for errors

### LINKED ISSUE #154: Check on-chain tx state after call
## Problem Definition

Currently the MS doesn't check if a transaction succeeded or failed. 

## Actual Outcome

The MS assumed the tx went through properly, even though it reverts.

## Expected Outcome

The tx receipt should be checked and a proper error message should be shown in case something went wrong.

palango:
Fixes #154

### DIFF
--- a/src/monitoring_service/handlers.py
+++ b/src/monitoring_service/handlers.py
@@ -38,6 +38,7 @@ class Context:
     ms_state: MonitoringServiceState
     db: Database
     scheduled_events: List[ScheduledEvent]
+    waiting_transactions: List[str]
     w3: Web3
     contract_manager: ContractManager
     last_known_block: int
@@ -371,12 +372,15 @@ def action_monitoring_triggered_event_handler(event: Event, context: Context) ->
             ).transact({'from': context.ms_state.address})
 
             log.info(
-                'Calling `monitor` on channel',
+                'Sent transaction calling `monitor` for channel',
                 token_network_address=channel.token_network_address,
                 channel_identifier=channel.identifier,
                 transaction_hash=encode_hex(tx_hash),
             )
             assert tx_hash is not None
+
+            # Add tx hash to list of waiting transactions
+            context.waiting_transactions.append(encode_hex(tx_hash))
 
             channel.closing_tx_hash = tx_hash
             context.db.upsert_channel(channel)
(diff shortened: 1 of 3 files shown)


src/monitoring_service/handlers.py
```diff
@@ -371,13 +372,16 @@ def action_monitoring_triggered_event_handler(event: Event, context: Context) ->
...
-                'Calling `monitor` on channel',
+                'Sent transaction calling `monitor` for channel',
                 token_network_address=channel.token_network_address,
                 channel_identifier=channel.identifier,
                 transaction_hash=encode_hex(tx_hash),
             )
             assert tx_hash is not None
 
+            # Add tx hash to list of waiting transactions
+            context.waiting_transactions.append(encode_hex(tx_hash))
```
karlb: Isn't this already hex encoded?

src/monitoring_service/service.py
```diff
@@ -163,16 +163,41 @@ def _process_new_blocks(self, last_block: BlockNumber) -> None:
 
         # check triggered events
         # TODO: create a priority queue for this
```
karlb: Not for this PR, but now that I see it: I don't think we want a priority queue here, but instead we want to write these into the db, so that the events are not lost on crash. We also have to extend our crash test to cover such a scenario.

src/monitoring_service/service.py
```diff
@@ -163,16 +163,41 @@ def _process_new_blocks(self, last_block: BlockNumber) -> None:
...
+            self.context.scheduled_events.remove(event)
 
         if self.context.scheduled_events:
-            log.info('Scheduled_events', events=self.context.scheduled_events)
+            log.debug('Scheduled_events', events=self.context.scheduled_events)
+
+        # check pending transactions
+        # this is done here so we don't have to block waiting for receipts in the state machine
+        hash_to_remove = []
+        for tx_hash in self.context.waiting_transactions:
```
karlb: Wouldn't it be easier to iterate over `self.context.waiting_transactions.copy()` and remove the elements right away instead of tracking the elements you want to remove?

src/monitoring_service/service.py
```diff
@@ -163,16 +163,41 @@ def _process_new_blocks(self, last_block: BlockNumber) -> None:
...
+            receipt = self.web3.eth.getTransactionReceipt(tx_hash)
```
karlb: Will this return soon in all cases, or might there be cases where we wait for the receipt for many blocks, because the transaction is not mined quickly?

palango: This might take a while. Do you have some improvement in mind?

src/monitoring_service/service.py
palango: Yes, both this and the other queue needs to be persisted to the DB

palango: See #157

src/monitoring_service/handlers.py
palango: No, it's `HexBytes`. That should be fine as well but a string feels cleaner.

src/monitoring_service/service.py
palango: Good point, that looks a lot cleaner.

src/monitoring_service/service.py
karlb: I mixed it up with `waitForTransactionReceipt`. This should be ok as it is now.

PR #162, opened 2019-03-13
Update contracts to 0.14

palango:
We need to do this to follow raiden: https://travis-ci.com/raiden-network/raiden-services/builds/104211941#L921


palango:
No, but I think it's better to wait for Karl to fix this. He did the changes in the contracts.

karlb:
> The contracts according to the contracts repo are at version 0.10. Is this important?

The version numbers for the `raiden-contracts` python package and for the deployed contracts differ. This is confusing, but necessary because not all python package changes warrant a new deployment.

PR #169, opened 2019-03-13
Store scheduled events in DB as well

### LINKED ISSUE #157: Save required internal state to DB
## Problem Definition

The is some internal state that we need to persist, like `Context.scheduled_events`  and ` Context.waiting_transactions`. 

## Actual Outcome

Entries in these lists are lost after restarts.

## Expected Outcome

Items should still be available after a restart, with those deleted that do not make sense any more.

palango:
Merge after #159 .

Store scheduled events in DB as well

This commit moves scheduled events from memory to the database.
This makes them persistent, so that they are available after restarts
and makes sure that the DB is the single source of truth.

Fixes #157

### DIFF
--- a/src/monitoring_service/database.py
+++ b/src/monitoring_service/database.py
@@ -51,6 +51,7 @@ class SharedDatabase:
             f'file:{filename}?mode={mode}',
             detect_types=sqlite3.PARSE_DECLTYPES,
             uri=True,
+            isolation_level=None,  # Disable sqlite3 module’s implicit transaction management
         )
         self.conn.row_factory = sqlite3.Row
         self.conn.execute("PRAGMA foreign_keys = ON")
@@ -125,11 +126,11 @@ class SharedDatabase:
             ]
         else:
             values += [None, None]
-        with self.conn:
-            upsert_sql = "INSERT OR REPLACE INTO channel VALUES ({})".format(
-                ', '.join('?' * len(values)),
-            )
-            self.conn.execute(upsert_sql, values)
+
+        upsert_sql = "INSERT OR REPLACE INTO channel VALUES ({})".format(
+            ', '.join('?' * len(values)),
+        )
+        self.conn.execute(upsert_sql, values)
 
     def get_channel(self, token_network_address: str, channel_id: int) -> Optional[Channel]:
         row = self.conn.execute(
@@ -166,11 +167,10 @@ class SharedDatabase:
             hex(contained_event.channel_identifier),
             contained_event.non_closing_participant,
         ]
-        with self.conn:
-            upsert_sql = "INSERT OR REPLACE INTO scheduled_events VALUES ({})".format(
-                ', '.join('?' * len(values)),
-            )
-            self.conn.execute(upsert_sql, values)
+        upsert_sql = "INSERT OR REPLACE INTO scheduled_events VALUES ({})".format(
+            ', '.join('?' * len(values)),
+        )
+        self.conn.execute(upsert_sql, values)
 
     def get_scheduled_events(
             self,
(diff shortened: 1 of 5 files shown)


src/monitoring_service/database.py
```diff
@@ -143,6 +157,94 @@ def get_channel(self, token_network_address: str, channel_id: int) -> Optional[C
...
+    def upsert_scheduled_event(self, event: ScheduledEvent) -> None:
+        contained_event: SubEvent = cast(SubEvent, event.event)
+        values = [
+            hex(event.trigger_block_number),
+            EVENT_TYPE_ID_MAP[type(contained_event)],
+            contained_event.token_network_address,
+            hex(contained_event.channel_identifier),
+            contained_event.non_closing_participant,
+        ]
+        with self.conn:
```
karlb: Transaction should be outside the Database class or you will lose the possibility to group different writes into the same transaction. Even grouping some reads along with a single write can be necessary for consistency when having concurrent db access.

Apparently, we want one transaction be write in many cases and the transaction handling is getting tedious. We could us autocommit on our connection by setting `isolation_level = None` when creating the connection. See https://docs.python.org/3.7/library/sqlite3.html#controlling-transactions.

src/monitoring_service/database.py
```diff
@@ -13,11 +18,20 @@
...
 )
 
+SubEvent = Union[ActionMonitoringTriggeredEvent, ActionClaimRewardTriggeredEvent]
+
 log = structlog.get_logger(__name__)
 SCHEMA_FILENAME = os.path.join(
     os.path.dirname(os.path.realpath(__file__)),
     'schema.sql',
 )
+EVENT_ID_TYPE_MAP = {
```
karlb: Maybe overkill for this simple case, but and `Enum` could be used here (along with `@unique`).

src/monitoring_service/database.py
```diff
@@ -143,6 +157,94 @@ def get_channel(self, token_network_address: str, channel_id: int) -> Optional[C
...
+        rows = self.conn.execute(
+            """
+                SELECT * FROM scheduled_events
+                WHERE trigger_block_number <= ?
+            """,
+            [hex(max_trigger_block)],
+        ).fetchall()
+
+        def create_scheduled_event(row: sqlite3.Row) -> ScheduledEvent:
+            klass = EVENT_ID_TYPE_MAP[row['event_type']]
```
karlb: The variable could be name `EventType`, which is more descriptive and gets around the spelling-mistake-to-avoid-keyword uglyness.

src/monitoring_service/database.py
```diff
@@ -143,6 +157,94 @@ def get_channel(self, token_network_address: str, channel_id: int) -> Optional[C
...
+
+    def get_waiting_transactions(self) -> List[str]:
+        return [
+            row[0] for row in self.conn.execute(
+                "SELECT transaction_hash FROM waiting_transactions",
+            )
+        ]
+
+    def add_waiting_transaction(self, waiting_tx_hash: str) -> None:
+        with self.conn:
```
karlb: Same as above.

src/monitoring_service/database.py
```diff
@@ -143,6 +157,94 @@ def get_channel(self, token_network_address: str, channel_id: int) -> Optional[C
...
+            self.conn.execute(
+                "INSERT INTO waiting_transactions VALUES (?)",
+                [waiting_tx_hash],
+            )
+
+    def remove_waiting_transaction(self, tx_hash: str) -> None:
+        with self.conn:
```
karlb: Same as above.

src/monitoring_service/schema.sql
```diff
@@ -52,3 +52,18 @@ CREATE TABLE monitor_request (
...
+
+CREATE TABLE scheduled_events (
+    trigger_block_number    HEX_INT     NOT NULL,
+    event_type              INT NOT NULL CHECK (event_type >= 0 AND event_type <=1),
+
+    token_network_address   CHAR(42)    NOT NULL,
+    channel_identifier      HEX_INT     NOT NULL,
+    non_closing_participant CHAR(42)    NOT NULL,
+
+    PRIMARY KEY (trigger_block_number, channel_identifier, token_network_address)
```
karlb: Can only a single event be scheduled for same block and channel?

src/monitoring_service/database.py
palango: I'd rather leave it for now. We can look into this again when we touch it.

palango:
Updated PR to the feedback.

PR #170, opened 2019-03-13
Idempotent scheduled events

### LINKED ISSUE #161: Creation of scheduled events is not idempotent
## Problem Definition

When the MS crashes before a `UpdatedHeadBlockEvent` has been processed but after a `ScheduledEvent` has been created. This `ScheduledEvent` will be duplicated on the next run.

## Actual Outcome

Duplicated `ScheduledEvent` 

## Expected Outcome

`ScheduledEvent`  should not be duplicated

### Related
#157

palango:
Merge after #169 

Fixes #161

### DIFF
--- a/tests/monitoring/monitoring_service/test_database.py
+++ b/tests/monitoring/monitoring_service/test_database.py
@@ -37,7 +37,27 @@ def test_scheduled_events(ms_database):
     assert len(ms_database.get_scheduled_events(23)) == 1
     assert len(ms_database.get_scheduled_events(24)) == 2
 
+    ms_database.upsert_scheduled_event(event=event1)
+    assert ms_database.scheduled_event_count() == 2
+
+    assert len(ms_database.get_scheduled_events(22)) == 0
+    assert len(ms_database.get_scheduled_events(23)) == 1
+    assert len(ms_database.get_scheduled_events(24)) == 2
+
     ms_database.remove_scheduled_event(event2)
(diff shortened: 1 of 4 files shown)


palango:
Changes stuff as per our discussion and added more tests.

tests/monitoring/monitoring_service/test_database.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def test_waiting_transactions(ms_database):
+    assert ms_database.get_waiting_transactions() == []
+
+    ms_database.add_waiting_transaction('0xA')
+    assert ms_database.get_waiting_transactions() == ['0xA']
+
+    ms_database.add_waiting_transaction('0xA')
+    assert ms_database.get_waiting_transactions() == ['0xA', '0xA']
```
karlb: Having two waiting transactions with the same hash is s weird case. It won't do any harm having it in there twice, but it also should not actually happen.

tests/monitoring/monitoring_service/test_handlers.py
```diff
@@ -186,6 +186,30 @@ def test_channel_closed_event_handler_closes_existing_channel(
     assert_channel_state(context, ChannelState.CLOSED)
 
 
+def test_channel_closed_event_handler_idempotency(
```
karlb: Maybe we can find a good way to test idempotency for all events without writing a separate test for each event in the future.

PR #172, opened 2019-03-15
Add payment handling to PFS

### LINKED ISSUE #149: Payments for PFS
## Description

pay the PFS

**So that it has**

an economic incentive to provide services.

## Acceptance Criteria

- [x] Provide interface defined in https://raiden-network-specification.readthedocs.io/en/latest/pathfinding_service.html#post-api-v1-token-network-address-paths
  - [x] Just accept one session for now, block all others from PFS side
  - [x] Happy case tests
  - [x] Test failure cases
- [x] Provide interface for https://raiden-network-specification.readthedocs.io/en/latest/pathfinding_service.html#get-api-v1-token-network-address-payment-iou
  - [x] Tests

## Related story / issue
- https://github.com/raiden-network/raiden/issues/3461

**13 SP**

karlb:
In addition to the main feature, two things in this PR need review/discussion:

## Using marshmallow for serialization/deserialization/validation
I tried to consolidate the following tasks into a single dataclass:
* validating data passed via HTTP
* serializing data for HTTP responses
* serializing into basic types for db storage
* providing detailed types for typechecking
Have a look at the `IOU` and `IOURequest` classes and check if this is a good way to do this. We might also compare with raiden.

## Handling exceptions with error codes
All subclasses of `APIException` will be automatically converted into proper HTTP error responses when raised. This allows raising errors in all functions instead of only allowing a return from the `Resource`. Right now I created a separate `exceptions.py` for the PFS, but if we keep this approach we probably want to move all or part of the exceptions into `raiden_libs`. Are there any approaches to error code in other raiden projects?

Required followups:
* Add error codes to spec
* Check deposit in UDC

closes #149

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,3 +24,5 @@ networkx==2.1
 requests==2.20.0
 
 jsonschema==2.6.0
+marshmallow-dataclass==0.5
+#marshmallow-dataclass==6.0.0b7
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,5 +24,5 @@ networkx==2.1
 requests==2.20.0
 
 jsonschema==2.6.0
-marshmallow-dataclass==0.5
+marshmallow-dataclass==0.5.5
 #marshmallow-dataclass==6.0.0b7
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,4 +24,5 @@ networkx==2.1
 requests==2.20.0
 
 jsonschema==2.6.0
+marshmallow==2.15.4
 marshmallow-dataclass==0.5.5
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,4 +25,3 @@ requests==2.20.0
 
 jsonschema==2.6.0
 marshmallow-dataclass==0.5.5
-#marshmallow-dataclass==6.0.0b7
--- a/src/pathfinding_service/api/rest.py
+++ b/src/pathfinding_service/api/rest.py
@@ -144,6 +144,7 @@ def process_payment(iou_dict, pathfinding_service):
     if iou_dict is None:
         raise exceptions.MissingIOU
 
+    # Basic IOU validity checks
     iou, errors = IOU.Schema().load(iou_dict)
     if errors:
         raise exceptions.InvalidRequest(**errors)
@@ -152,22 +153,24 @@ def process_payment(iou_dict, pathfinding_service):
     if not iou.is_signature_valid():
         raise exceptions.InvalidIOUSignature
 
-    # TODO:
-    # * does sender have other IOUs?
-    # * is deposit large enough?
-
-    last_iou = pathfinding_service.database.get_iou(
-        iou.sender,
-        iou.expiration_block,
+    # Compare with known IOU
+    active_iou = pathfinding_service.database.get_iou(
+        sender=iou.sender,
+        claimed=False,
     )
-    if last_iou:
-        if last_iou.claimed:
-            raise exceptions.IOUAlreadyClaimed
-        else:
-            iou.claimed = False
-        expected_amount = last_iou.amount + pathfinding_service.service_fee
+    if active_iou:
+        if active_iou.expiration_block != iou.expiration_block:
+            raise exceptions.UseThisIOU(iou=active_iou)
+
+        expected_amount = active_iou.amount + pathfinding_service.service_fee
     else:
-        iou.claimed = False
+        if pathfinding_service.database.get_iou(
+            sender=iou.sender,
+            expiration_block=iou.expiration_block,
+            claimed=True,
+        ):
+            raise exceptions.IOUAlreadyClaimed
+
         min_expiry = pathfinding_service.web3.eth.blockNumber + MIN_IOU_EXPIRY
         if iou.expiration_block < min_expiry:
             raise exceptions.IOUExpiredTooEarly(min_expiry=min_expiry)
@@ -175,6 +178,10 @@ def process_payment(iou_dict, pathfinding_service):
     if iou.amount < expected_amount:
         raise exceptions.InsufficientServicePayment(expected_amount=expected_amount)
 
+    # TODO: deposit large enough?
+
+    # Save latest IOU
+    iou.claimed = False
     pathfinding_service.database.upsert_iou(
         IOU.Schema().dump(iou)[0],
     )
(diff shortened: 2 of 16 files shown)


src/pathfinding_service/api/rest.py
```diff
@@ -120,6 +147,102 @@ def post(self, token_network_address: str):
         return {'result': paths}, 200
 
 
+def process_payment(iou_dict, pathfinding_service):
```
palango: Please add type annotations

requirements.txt
```diff
@@ -24,3 +24,4 @@ networkx==2.1
 requests==2.20.0
 
 jsonschema==2.6.0
+marshmallow-dataclass==0.5.5
```
palango: `marshmaollow` itself should be added as well.

src/pathfinding_service/api/rest.py
```diff
@@ -120,6 +147,102 @@ def post(self, token_network_address: str):
...
+        sender=iou.sender,
+        claimed=False,
+    )
+    if active_iou:
+        if active_iou.expiration_block != iou.expiration_block:
+            raise exceptions.UseThisIOU(iou=active_iou)
+
+        expected_amount = active_iou.amount + pathfinding_service.service_fee
+    else:
+        if pathfinding_service.database.get_iou(
```
palango: Can you extract the variable from the call, this is hard to read.

src/pathfinding_service/api/rest.py
```diff
@@ -120,6 +147,102 @@ def post(self, token_network_address: str):
...
+            raise exceptions.IOUAlreadyClaimed
+
+        min_expiry = pathfinding_service.web3.eth.blockNumber + MIN_IOU_EXPIRY
+        if iou.expiration_block < min_expiry:
+            raise exceptions.IOUExpiredTooEarly(min_expiry=min_expiry)
+        expected_amount = pathfinding_service.service_fee
+    if iou.amount < expected_amount:
+        raise exceptions.InsufficientServicePayment(expected_amount=expected_amount)
+
+    # TODO: deposit large enough?
```
palango: Can you create a follow-up issue for this?

src/pathfinding_service/api/rest.py
```diff
@@ -120,6 +147,102 @@ def post(self, token_network_address: str):
...
+        packed_data = (
+            Web3.toBytes(hexstr=self.sender) +
+            Web3.toBytes(hexstr=self.receiver) +
+            Web3.toBytes(text=self.timestamp_str)
+        )
+        try:
+            recovered_address = eth_recover(packed_data, self.signature)
+        except InvalidSignature:
+            return False
+        return to_checksum_address(recovered_address) == to_checksum_address(self.sender)
```
palango: There's also `eth_utils.is_same_address` which could make this clearer.

src/pathfinding_service/api/rest.py
```diff
@@ -120,6 +147,102 @@ def post(self, token_network_address: str):
...
+
+class IOUResource(PathfinderResource):
+
+    def get(self, token_network_address: TokenNetworkAddress):
+        iou_request, errors = IOURequest.Schema().load(request.args)
+        if errors:
+            raise exceptions.InvalidRequest(**errors)
+        if not iou_request.is_signature_valid():
+            raise exceptions.InvalidSignature
+        if iou_request.timestamp < datetime.utcnow() - timedelta(hours=1):
```
palango: Can we make the time delta a constant.

src/pathfinding_service/api/rest.py
```diff
@@ -120,6 +147,102 @@ def post(self, token_network_address: str):
...
+        last_iou = self.pathfinding_service.database.get_iou(
+            sender=iou_request.sender,
+            claimed=False,
+        )
+        if last_iou:
+            last_iou = IOU.Schema(strict=True, exclude=['claimed']).dump(last_iou)[0]
+
+        return {
+            'last_iou': last_iou,
+        }, 200
```
palango: I wonder if we should return a 404 Not Found here, if no IOU is available.

src/pathfinding_service/database.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+class PFSDatabase:
+    """ Store data that needs to persist between PFS restarts """
+
+    def __init__(self, filename: str, pfs_address: Address):
+        log.info('Opening database at ' + filename)
+        self.conn = sqlite3.connect(
+            filename,
+            detect_types=sqlite3.PARSE_DECLTYPES,
```
palango: Should we disable implicit tx handling here as well?

src/pathfinding_service/config.py
```diff
@@ -15,3 +15,7 @@
 DEFAULT_SETTLE_TO_REVEAL_TIMEOUT_RATIO = 2
 
 DEFAULT_POLL_INTERVALL = 10
+
+# When a new IOU session is started, this is the minimum number of blocks
+# between the current block and `expiration_block`.
+MIN_IOU_EXPIRY = 1000
```
palango: I'd say we should increase this, especially in the beginning. What about a day, i.e. 24*60*4 = 5760?

palango: I'd say we should increase this, especially in the beginning. What about a day, i.e. 24*60*4 = 5760?

requirements.txt
karlb: It's already in raiden, so I add it only with a major version number?

src/pathfinding_service/api/rest.py
karlb: Good to know. That's better, of course.

src/pathfinding_service/api/rest.py
karlb: https://github.com/raiden-network/raiden-services/issues/177

requirements.txt
karlb: Pip is unable to properly resolve dependency versions. It will install a slightly more recent version than requested by raiden. The code still works with the exception of the `pkg_resources.require('raiden-services')[0].version` call which raises
```
pkg_resources.ContextualVersionConflict: (marshmallow 2.19.1 (/home/karl/tools/venv-services/lib/python3.7/site-packages), Requirement.parse('marshmallow==2.15.4'), {'raiden'})
```
It also causes a warning during install.

Do you know a clean solution to this?

src/pathfinding_service/api/rest.py
karlb: Yes, I that's an even better fit than 200.

src/pathfinding_service/database.py
karlb: Yes, good point!

PR #175, opened 2019-03-18
Balance Update Integration

Dominik1999:


src/pathfinding_service/model/channel_view.py
```diff
@@ -23,6 +23,7 @@ def __init__(
         settle_timeout: int,
         deposit: int = 0,
         reveal_timeout: int = DEFAULT_REVEAL_TIMEOUT,
+        balance_update_nonce: int = 0,
```
palango: Is this ever set in the constructor, otherwise just set it to 0 by default

src/pathfinding_service/model/channel_view.py
```diff
@@ -31,46 +32,35 @@ def __init__(
...
         self._deposit = deposit
-        self._transferred_amount = 0
-        self._received_amount = 0
-        self._locked_amount = 0
         self._capacity = deposit
         self.state = ChannelView.State.OPEN
         self.channel_id = channel_id
         self.settle_timeout = settle_timeout
         self.reveal_timeout = reveal_timeout
+        self.balance_update_nonce = balance_update_nonce
```
palango: Is there another `nonce`? Otherwise I'd prefer the shorter name.

src/pathfinding_service/model/token_network.py
```diff
@@ -56,14 +56,18 @@ def handle_channel_opened_event(
             participant1=participant1,
             participant2=participant2,
             settle_timeout=settle_timeout,
-            deposit=0)
+            deposit=0,
+            balance_update_nonce=0,
```
palango: See comment above, could be removed.

src/pathfinding_service/model/token_network.py
```diff
@@ -118,30 +122,59 @@ def handle_channel_closed_event(self, channel_identifier: ChannelIdentifier):
...
             )
+            return
+
+        # Check nonces of the balance update against current state
+        if (
+            updating_nonce <= channel_view_to_partner.balance_update_nonce and
+            other_nonce <= channel_view_from_partner.balance_update_nonce
+        ):
+            log.debug(
+                "Balance Update already received",
```
palango: Some additional infor here would be nice, like channel id.

src/pathfinding_service/model/token_network.py
```diff
@@ -118,30 +122,59 @@ def handle_channel_closed_event(self, channel_identifier: ChannelIdentifier):
...
+            )
+            return
+
+        # Call function update_capacity
```
palango: Does this comment help to understand better what happens?

tests/pathfinding/fixtures/network_service.py
```diff
@@ -204,12 +214,12 @@ def populate_token_network(
         for channel_id, (
             p1_index,
             p1_deposit,
-            _p1_transferred_amount,
+            p1_transferred_amount,
```
palango: We could change this to `capacity` to not have to do the calculations below.

tests/pathfinding/test_graphs.py
```diff
@@ -101,13 +99,9 @@ def test_routing_result_order(
         max_paths=5,
         hop_bias=1,
     )
-    # 5 paths requested, but only 3 are available
-    assert len(paths) == 2
+    # 5 paths requested, but only 1 ist available
```
palango: `ist` -> `is`

src/pathfinding_service/model/channel_view.py
Dominik1999: not yet, but there might be another nonce when we have withdraw and new deposits?!

src/pathfinding_service/model/channel_view.py
Dominik1999: Isn't that the constructor where it is set to 0 by default? Or what does `: int = 0` mean?

src/pathfinding_service/model/token_network.py
Dominik1999: the `=0` could be removed, correct?

tests/pathfinding/fixtures/network_service.py
Dominik1999: will do in the next PR

PR #176, opened 2019-03-18
Replace sign and recover functions with raiden equivalents

### LINKED ISSUE #36: Remove/replace deprecated functions
We use some deprectated functions from raiden-libs. These usages should be replaced.

palango:
@andrevmatos I just replaced the `raiden-libs` functions with the `raiden` equivalents, but some tests break. Do you have any hints/ideas before I dive into this tomorrow? Ref: https://travis-ci.com/palango/raiden-services/jobs/185743069#L959

Fixes #36


andrevmatos:
I'm not sure about coincurve's algorithm, but seems like the invalid signature you're trying to use on `test_invalid_request_signatures` (`b'\x00' * 65`) isn't a valid signature, so it raises an `InvalidSignature` exception that you're not catching. `'\x01' * 65` is a valid signature, which will yield some valid address, you can both test with it, and also properly handle said exception on your method and return False if it's raised, to ensure no one crashes your app with an invalid signature.

palango:
Thanks @andrevmatos . It was easy, I just forgot to update the `InvalidSignature` exception from `raiden-libs` to `raiden.exceptions`.

src/monitoring_service/states.py
```diff
@@ -106,7 +107,8 @@ def __init__(
 
         if signature is None:
             assert priv_key
-            self.signature = encode_hex(eth_sign(priv_key, self.serialize_bin()))
+            signer = LocalSigner(private_key=to_bytes(hexstr=priv_key))
```
karlb: What is the difference between `to_bytes(hexstr=priv_key)` and `decode_hex(priv_key)`?

src/monitoring_service/states.py
```diff
@@ -178,6 +180,7 @@ def from_balance_proof(
         )
 
     def sign(self, priv_key: str) -> 'MonitorRequest':
+        signer = LocalSigner(private_key=to_bytes(hexstr=priv_key))
```
karlb: Now `signer` is a `LocalSigner` instance in some places and an Ethereum address of the signing party in other places. This can cause confusion or name clashes. But I also don't have a good suggestion, right now.

palango:
> I actually prefer the old eth_sign function signature a bit for our use case, but the consistency and code deduplication is clearly worth it.

We can also remove the deprecation warnings and continue using these functions.

src/monitoring_service/states.py
palango: `decode_hex`  is nicer ;)

andrevmatos:
@karlb this usecase of creating efemeral LocalSigners to sign in a single point and discarding it works but isn't the intended usage of signers.
You may like it a little more if you imagine that the Signer can be created once in a place, and then passed around, instead of the raw private key. The selling point here is that Signer is the interface for any signer (provides `address`, `sign`, soon `signTransaction`, etc), and `LocalSigner` is just a `Signer` which takes a raw private key, and could be easily swapped by e.g. a RemoteSIgner which does remote signing on an unlocked or interactive account in ETH node through JSON Rpc, without changing any code that uses the signer, just the initial, instantiated object, or a hardwallet Signer, etc. This isn't possible at all with `eth_signer`.

karlb:
@andrevmatos, thanks for the details, it's nice to have that context.

PR #187, opened 2019-03-22
Disable gevent `thread` resolver for PFS

karlb:
It was causing `socket.gaierror: [Errno -2] Name or service not known`
errors when run inside docker.


karlb:
Suggestions why this breaks on travis are welcome!

palango:
Do we maybe have to set this during test setup (`conftest.py`) as well?

palango:
Mh, apparently not -.-

karlb:
There are multiple places where we do the monkey patching. Maybe I have to do it in all of them. Or move the patching up to the top level of the tests. I didn't think thoroughly about the execution order and the effects of the money patching, yet.

PR #189, opened 2019-03-22
Adds option for minimum required service fee for ms

### LINKED ISSUE #185: Add option for MS minimum reward
## Problem Definition

As the MS I'd like to define a minimal reward that has to be given in order to provide my services.

### ToDos
- [x] Decide where to filter MRs based on reward (I'd argue that we want to do this in the MS, and still collect everything in the RC)
- [x] Add a command line flag for the minimal reward
- [x] Only call `monitor` when the reward of the MR is bigger than the threshold.

Dominik1999:
Closes #185

### DIFF
--- a/src/monitoring_service/cli.py
+++ b/src/monitoring_service/cli.py
@@ -92,6 +92,12 @@ def validate_address(_ctx: Any, _param: Any, value: Optional[str]) -> Optional[s
     type=str,
     help='Path to SQLite3 db which stores the application state',
 )
+@click.option(
+    '--min-reward',
+    default=0,
+    type=click.IntRange(min=0),
+    help='Minimum reward which is required before processing requests',
+)
 def main(
     keystore_file: str,
     password: str,
@@ -103,6 +109,7 @@ def main(
     confirmations: int,
     log_level: str,
     state_db: str,
+    min_reward: int,
 ) -> None:
     setup_logging(log_level)
 
@@ -153,6 +160,7 @@ def main(
         sync_start_block=contract_infos[START_BLOCK_ID],
         required_confirmations=confirmations,
         db_filename=state_db,
+        min_reward=min_reward,
     )
     ms.start()
 
--- a/src/monitoring_service/handlers.py
+++ b/src/monitoring_service/handlers.py
@@ -45,6 +45,7 @@ class Context:
     last_known_block: int
     monitoring_service_contract: Contract
     user_deposit_contract: Contract
+    min_reward: int
 
 
 def channel_opened_event_handler(event: Event, context: Context) -> None:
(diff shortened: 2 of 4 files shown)


src/monitoring_service/cli.py
```diff
@@ -92,6 +92,12 @@ def validate_address(_ctx: Any, _param: Any, value: Optional[str]) -> Optional[s
     type=str,
     help='Path to SQLite3 db which stores the application state',
 )
+@click.option(
+    '--min-service-fee',
```
palango: I wonder if we should call this fee or reward...

src/monitoring_service/handlers.py
```diff
@@ -383,10 +384,26 @@ def action_monitoring_triggered_event_handler(event: Event, context: Context) ->
     user_address = monitor_request.non_closing_signer
     user_deposit = context.user_deposit_contract.functions.effectiveBalance(user_address).call()
 
+    if monitor_request.reward_amount < context.min_service_fee:
+        log.info('Monitor request not executed due to insufficient reward amount',
```
palango: Formatting is off here

src/monitoring_service/handlers.py
```diff
@@ -383,10 +384,26 @@ def action_monitoring_triggered_event_handler(event: Event, context: Context) ->
...
+                 monitor_request_signer=monitor_request.signer,
```
palango: instead of having all fields, you can also just write: `monitor_request=monitor_request`

tests/monitoring/monitoring_service/test_handlers.py
```diff
@@ -548,6 +549,41 @@ def test_action_monitoring_triggered_event_handler_with_sufficient_balance_does_
     assert context.monitoring_service_contract.functions.monitor.called is True
 
 
+def test_action_monitoring_triggered_event_handler_with_insufficient_reward_amount_does_not_trigger_monitor_call(  # noqa
+        context: Context,
+):
+    """ Tests that `monitor` is called when the ActionMonitoringTriggeredEvent is triggered and
```
palango: This needs to be updated or removed

tests/monitoring/monitoring_service/test_handlers.py
```diff
@@ -548,6 +549,41 @@ def test_action_monitoring_triggered_event_handler_with_sufficient_balance_does_
...
+    )
+
+    channel = context.db.get_channel(
+        trigger_event.token_network_address,
+        trigger_event.channel_identifier,
+    )
+    assert channel
+    assert channel.closing_tx_hash is None
+
+    context.user_deposit_contract.functions.effectiveBalance(
```
palango: Is this necessary for the test?

src/monitoring_service/cli.py
Dominik1999: We are not quite consistent here all over the place. In the spec we call it both. Maybe define it once and for all? For the monitoring service I tend to "reward" or even "bounty"

src/monitoring_service/handlers.py
Dominik1999: ok done

src/monitoring_service/handlers.py
Dominik1999: good point and the min_service_fee

tests/monitoring/monitoring_service/test_handlers.py
Dominik1999: done

tests/monitoring/monitoring_service/test_handlers.py
Dominik1999: seems necessary, test fails without it, which is indeed strange. Error message
```
E       TypeError: '>=' not supported between instances of 'Mock' and 'int'

src/monitoring_service/handlers.py:397: TypeError
```

src/monitoring_service/cli.py
Dominik1999: I will call it reward now

Dominik1999: `reward` and `min_reward` seems consistent and for the pathfinding service we will call it fee

src/monitoring_service/cli.py
```diff
@@ -92,6 +92,12 @@ def validate_address(_ctx: Any, _param: Any, value: Optional[str]) -> Optional[s
     type=str,
     help='Path to SQLite3 db which stores the application state',
 )
+@click.option(
+    '--min-reward',
+    default=0,
+    type=click.IntRange(min=0),
+    help='Minimum service fee which is required before processing requests',
```
palango: replace fee here as well?

Dominik1999: good catch

PR #197, opened 2019-03-26
Fix db path creation for local dbs

karlb:
In accd17fd7b298057260848a604da321e28676855 a bug has been introduced
which causes failure on plain db names without specifying a path.

I thought `os.makedirs(os.path.dirname(filename), exist_ok=True)` would
work in all cases, but if not path segment is in `filename`, the
`dirname(filename) == ''` which is not a valid path for `makedirs`.

### DIFF
--- a/src/monitoring_service/database.py
+++ b/src/monitoring_service/database.py
@@ -47,7 +47,7 @@ class SharedDatabase:
 
     def __init__(self, filename: str, allow_create: bool = False):
         log.info('Opening database', filename=filename)
-        if filename != ':memory:':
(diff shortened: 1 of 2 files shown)


karlb:
> Didn't you say we could use some click utility?

For those cases where we want a filehandle, using http://click.palletsprojects.com/en/5.x/arguments/#file-arguments is a great choice. But sqlite wants a file name, so we could only use http://click.palletsprojects.com/en/5.x/arguments/#file-path-arguments. That won't take care of the path creation, though.

Issue #167, opened 2019-03-13
Fix type annotations


palango:
## Problem Definition

When we switched to the type definitions from raiden we missed that some of them doen't match the types we use (for example the `Address` is `bytes` in raiden but `str` in the services).

We need to fix this.

PR #203, opened 2019-03-28
Fix problems in on_pfs_update

palango:
It's really strange that this stuff wasn't found by mypy. I guess we should look into #167 soon.

### DIFF
--- a/src/raiden_libs/matrix.py
+++ b/src/raiden_libs/matrix.py
@@ -47,6 +47,7 @@ def message_from_dict(data: dict) -> Message:
             raise InvalidProtocolMessage(
                 'Invalid message data. Can not find the data type',
             ) from None
+
     return klass.from_dict(data)
 
 
@@ -185,13 +186,14 @@ class MatrixListener(gevent.Greenlet):
                     _exc=ex,
                 )
                 continue
-            except InvalidProtocolMessage as ex:
+            except (InvalidProtocolMessage, KeyError) as ex:
                 logger.warning(
-                    "Message data JSON are not a valid message",
+                    "Message data JSON is not a valid message",
                     message_data=line,
                     _exc=ex,
                 )
                 continue
+
             if not isinstance(message, SignedMessage):
                 logger.warning(
                     'Received invalid message',
--- a/src/pathfinding_service/service.py
+++ b/src/pathfinding_service/service.py
@@ -317,8 +317,8 @@ class PathfindingService(gevent.Greenlet):
             channel_identifier=message.canonical_identifier.channel_identifier,
         )
 
-        assert is_checksum_address(message.updating_participant)
-        assert is_checksum_address(message.other_participant)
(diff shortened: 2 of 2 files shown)


src/raiden_libs/matrix.py
```diff
@@ -178,7 +179,7 @@ def _handle_message(self, room, event) -> bool:
             try:
                 message_dict = json.loads(line)
                 message = message_from_dict(message_dict)
-            except (UnicodeDecodeError, json.JSONDecodeError) as ex:
+            except (UnicodeDecodeError, json.JSONDecodeError, KeyError) as ex:
                 logger.warning(
                     "Can't parse message data JSON",
```
karlb: The error message is a bit misleading in case of a `KeyError`. Still somewhat ok, though.

palango: True, I'll add another case for this.

palango:
I still don't get why mypy doesn't complain about this:

- Our `Address` type is defined as `Address = NewType('Address', str)`
- This is used in the tests like this, where `addresses` has the type `List[Address]`:
```
    message = get_updatepfs_message(
        updating_participant=addresses[0],
        other_participant=addresses[1],
        updating_capacity=TokenAmount(UINT256_MAX + 1),
    )
```
- In there a `UpdatePFS` message is created, where `Address = NewType('Address', bytes)`
```
class UpdatePFS(SignedMessage):
    def __init__(
            self,
            *,
            canonical_identifier: CanonicalIdentifier,
            updating_participant: typing.Address,
            other_participant: typing.Address,
```

For me mypy should be able to see through this.

karlb:
Maybe two `NewType`s with the same name are treated as the same type?

karlb:
> Maybe two `NewType`s with the same name are treated as the same type?

No, that's not generally the case. The following program correctly complains:

```python
from typing import NewType


class C1:
    Address = NewType('Address', str)


class C2:
    Address = NewType('Address', bytes)


def f(x: C1.Address):
    return x


byte_addr = C2.Address(bytes([1, 2]))
f(byte_addr)
```

palango:
Seems like the problem is the `--ignore-missing-imports` flag. When I take it out of the mypy invocation I get (amoing other) the following warning:
```
src/monitoring_service/service.py:19: error: Cannot find module named 'raiden.utils.typing'
```

karlb:
> No the question is why that is.

I think mypy ignores that package because it doesn't declare itself as properly typed via PEP561:
https://mypy.readthedocs.io/en/latest/installed_packages.html#using-pep-561-compatible-packages-with-mypy

karlb:
See https://github.com/raiden-network/raiden/pull/3698. I'll try to fix as many of the type failures caused by this as possible before that PR is merged.

src/raiden_libs/matrix.py
palango: Moved it to the next except block which fits better.

PR #205, opened 2019-03-28
Fix types after declaring raiden as typed

### LINKED ISSUE #167: Fix type annotations

karlb:
I had to exclude one error for flake8 which prevented using
`varname: Type = Type(constant)` assignments for defaults in the
constructor. If you know how to get the types right without excluding
this error, let me know!

Fixes #167

### DIFF
--- a/tests/pathfinding/test_capacity_updates.py
+++ b/tests/pathfinding/test_capacity_updates.py
@@ -7,7 +7,7 @@ The Capacity Updates show different correct and incorrect values to test all edg
 from typing import List
 
 import pytest
-from eth_utils import decode_hex, to_checksum_address
+from eth_utils import decode_hex
 
 from pathfinding_service import PathfindingService
 from pathfinding_service.exceptions import InvalidCapacityUpdate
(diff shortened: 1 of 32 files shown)


tests/pathfinding/test_capacity_updates.py
```diff
@@ -34,26 +33,26 @@
...
 ) -> UpdatePFS:
     return UpdatePFS(
         canonical_identifier=CanonicalIdentifier(
             chain_identifier=chain_identifier,
             channel_identifier=channel_identifier,
             token_network_address=token_network_address,
         ),
-        updating_participant=updating_participant,
-        other_participant=other_participant,
+        updating_participant=to_checksum_address(updating_participant),
```
palango: Shouldn't this be `to_canonical_identifier` because the `UpdatePFS` message takes a `bytes` based `Address`  representation?

palango:
Ah, and it fixes #167 as well.

tests/pathfinding/test_capacity_updates.py
karlb: I was confused due to #203 not being merged, yet. I don't see where `to_canonical_identifier` should come from. I'll just use `decode_hex` for now.

Issue #151, opened 2019-03-11
Claim PFS IOUs


karlb:
## Description

**As Josh I want to be able to**

I want to claim the PFS fee IOUs at some point

**So that I can**

my token (from fees) into my wallet.

## Acceptance Criteria

- [x] Write a script that can claim certain IOUs
- [x] Create conditions for claiming IOUs (close expiration date, claimed amount should be larger than the transaction fees, ...)

PR #206, opened 2019-04-01
Extract common cli code to shared module

karlb:
When implementing https://github.com/raiden-network/raiden-services/issues/151, we'll have four cli commands with a lot options in common. So we need this to avoid excessive code duplication.


src/raiden_libs/cli.py
```diff
@@ -0,0 +1,73 @@
...
+    """A decorator to be used with all service commands
+
+    It will pass two new args to the given func:
+    * private_key (as a result of `--keystore-file` and `--password`)
+    * state_db
+
+    The `app_name` will be used to determine the state_db location.
+    """
+    def decorator(func: Callable) -> Callable:
+        for option in reversed([
```
palango: Why use `reversed` here?

tests/pathfinding/test_cli.py
```diff
@@ -82,7 +83,10 @@ def test_registry_address(default_cli_args):
 
     # check validation of address format
     def fails_on_registry_check(address):
-        result = runner.invoke(main, ['--registry-address', address], catch_exceptions=False)
+        result = runner.invoke(
+            main,
+            default_cli_args + ['--registry-address', address],
+            catch_exceptions=False)
```
palango: The closing `)` should go to the next line.

We could also think about using black for the repo in order to get rid of this manual formatting.

src/raiden_libs/cli.py
```diff
@@ -71,3 +83,49 @@ def decorator(func: Callable) -> Callable:
         return func
 
     return decorator
+
+
+def blockchain_options(func: Callable) -> Callable:
+    """A decorator to be used with all service commands
```
palango: docstring needs to be updated.

src/raiden_libs/cli.py
```diff
@@ -129,3 +141,59 @@ def blockchain_options(func: Callable) -> Callable:
...
+        )
+        sys.exit(1)
+
+    # Add POA middleware for geth POA chains, no/op for other chains
+    web3.middleware_stack.inject(geth_poa_middleware, layer=0)
+
+    # give web3 some time between retries before failing
+    provider.middlewares.replace(
+        'http_retry_request',
+        http_retry_with_backoff_middleware,
```
palango: Not sure if we can use this one with the MS right now, as it used `gevent.sleep` internally.

tests/pathfinding/test_cli.py
```diff
@@ -30,6 +30,17 @@
 }
 
 
+@pytest.fixture
+def provider_mock(monkeypatch):
+    provider_mock = Mock()
+    monkeypatch.setattr('raiden_libs.cli.HTTPProvider', provider_mock)
+    web3_mock = Mock()
+    web3_mock.return_value.net.version = 1
```
palango: Shouldn't this be `web3_mock.net.version.return_value` or do both work?

src/pathfinding_service/middleware.py
```diff
@@ -1,25 +1,26 @@
...
-):
+) -> Callable:
     """ Retry requests with exponential backoff
 
     Creates middleware that retries failed HTTP requests and exponentially
     increases the backoff between retries. Meant to replace the default
     middleware `http_retry_request_middleware` for HTTPProvider.
     """
-    def middleware(method, params):
+    def middleware(method: str, params: dict) -> Any:
```
palango: I prefer the capitalized version `Dict` for type annotations, but have no strong objections.

src/raiden_libs/cli.py
karlb: So that the order or options is handled in the same way as when using them directly on the command.

tests/pathfinding/test_cli.py
karlb: This is a mock for the `Web3` class, so the return value of this mock is the `Web3` instance. So we need the `return_value` in that place. The `version` is a property, so it's not called but accessed like an attribute. So we must not use `return_value` for it.

src/pathfinding_service/middleware.py
karlb: Since there are no uppercase variants of `str`, `int`, etc we will have a mixture of upper and lowercase, anyway. So I don't see and advantage in always using the uppercase when available. Using the lowercase saves adding and removing a few imports on code changes. No strong opinions here either, but I'm slightly on the side of using lowercase when the additional features of the uppercase variants are not needed. I couldn't find any official recommendations for either.

src/raiden_libs/cli.py
karlb: My quick manual test indicates that `gevent.sleep` behaves just like `time.sleep` when gevent is not actively used.

palango: But this would mean that we block for 204.6 s with the default settings in the worst case. On the other hand we cannot do anything without connecting to the node anyways.

karlb: Retrying for a few minutes seems like a better alternative than crashing to me.

palango: Makes sense 👍

src/pathfinding_service/middleware.py
palango: 👍

PR #209, opened 2019-04-01
Adds signature test to Capacity Update

Dominik1999:


tests/pathfinding/test_capacity_updates.py
```diff
@@ -58,27 +56,31 @@ def get_updatepfs_message(
...
+    pathfinding_service_mocked_listeners: PathfindingService,
 ):
 
     message = get_updatepfs_message(
         chain_identifier=ChainID(121212),
-        updating_participant=addresses[0],
-        other_participant=addresses[1],
+        updating_participant=private_key_to_address(PRIVAT_KEY_EXAMPLE_1),
+        other_participant=private_key_to_address(PRIVAT_KEY_EXAMPLE_2),
+        privkey_signer=PRIVAT_KEY_EXAMPLE_1,
```
palango: `PRIVAT_KEY_EXAMPLE_1` is the default, so you don't strictly need this here. But it might be nice for readability. (same is true some more times below.)

Dominik1999: you are right. Let's keep it for readability? I strongly suggest that :)

PR #218, opened 2019-04-03
Use sync_start_block in MS

### LINKED ISSUE #186: Use start_block CLI option in MS
## Problem Definition

The `MonitoringService` class has a parameter `sync_start_block` which gets set when using predeployed contracts. However it is currently not used.

Related to #168

palango:
Fixes #186

### DIFF
--- a/src/monitoring_service/database.py
+++ b/src/monitoring_service/database.py
@@ -219,14 +219,17 @@ class SharedDatabase:
     def remove_waiting_transaction(self, tx_hash: str) -> None:
         self.conn.execute("DELETE FROM waiting_transactions WHERE transaction_hash = ?", [tx_hash])
 
-    def load_state(self) -> MonitoringServiceState:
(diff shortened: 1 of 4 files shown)


src/monitoring_service/database.py
```diff
@@ -219,14 +219,17 @@ def add_waiting_transaction(self, waiting_tx_hash: str) -> None:
...
+    def load_state(self, sync_start_block: BlockNumber) -> MonitoringServiceState:
         """ Load MS state from db or return a new empty state if not saved one is present
         """
         blockchain = self.conn.execute("SELECT * FROM blockchain").fetchone()
         token_network_addresses = [
             row[0] for row in self.conn.execute("SELECT address FROM token_network")
         ]
+
         latest_known_block = blockchain['latest_known_block']
+        if latest_known_block == -1:
```
karlb: If we do such a check, we really should use `NULL`/`None`

PR #229, opened 2019-04-05
Run pylint with many checks turned off

karlb:
The category "good checks that could be enabled in the future" in
pylintrc should be worked through at some point.

### DIFF
--- a/src/monitoring_service/blockchain.py
+++ b/src/monitoring_service/blockchain.py
@@ -100,7 +100,8 @@ def query_blockchain_events(
     ]
 
 
-class BlockchainListener:
+# TODO: maybe pylint is right and this class should just be a `get_events` function?
+class BlockchainListener:  # pylint: disable=too-few-public-methods
     """ This is pull-based blockchain listener. """
 
     def __init__(self, web3: Web3, contract_manager: ContractManager):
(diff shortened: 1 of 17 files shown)


src/monitoring_service/blockchain.py
```diff
@@ -100,7 +100,8 @@ def query_blockchain_events(
     ]
 
 
-class BlockchainListener:
+# TODO: maybe pylint is right and this class should just be a `get_events` function?
+class BlockchainListener:  # pylint: disable=too-few-public-methods
```
palango: Created https://github.com/raiden-network/raiden-services/issues/230

src/pathfinding_service/claim_fees.py
```diff
@@ -59,7 +59,7 @@ def main(
     contract_manager = ContractManager(contracts_precompiled_path())
     one_to_n_contract = web3.eth.contract(
         abi=contract_manager.get_contract_abi(CONTRACT_ONE_TO_N),
-        address=user_deposit_contract_address,
+        address=contract_infos['user_deposit_contract_address'],
```
palango: Nice catch!

src/pathfinding_service/model/token_network.py
```diff
@@ -10,7 +10,7 @@
     DIVERSITY_PEN_DEFAULT,
     MAX_PATHS_PER_REQUEST,
 )
-from pathfinding_service.model import ChannelView
+from pathfinding_service.model.channel_view import ChannelView
```
palango: How could this work before?

karlb: `pathfinding_service/model/__init__.py` does an `import *` from it's submodules. The main reason I changed the import is that it caused a circular import, before. It did work, but I prefer to avoid circular imports.

PR #231, opened 2019-04-05
Check chain_id when accepting MRs

karlb:


src/request_collector/server.py
```diff
@@ -82,6 +83,11 @@ def on_monitor_request(self, request_monitoring: RequestMonitoring) -> None:
             log.info('Ignore MR with invalid signature', monitor_request=request_monitoring)
             return
 
+        # Validate MR
+        if monitor_request.chain_id != self.chain_id:
+            log.debug('Bad chain_id', monitor_request=monitor_request, expected=self.chain_id)
```
palango: I wonder if we want to increase the log level for this one?

karlb: I'm generally a bit unsure about which loglevel to choose in which cases. I assume this gets easier when getting some experience with debugging problems in production.

palango: Yes, let's leave it like this then.

PR #232, opened 2019-04-05
 Make order of `keystore` and `password` irrelevant 

### LINKED ISSUE #228: `password` and `keystore` parameters depend on order
These two parameters only work if passed in the correct order on the command line, right now. But the order should be independent.

This can probably be solved by moving the `_open_keystore` function from the callback into the `common_options` decorator.

karlb:
I also think this is nicer to read and more flexible than before.

Closes #228.

### DIFF
--- a/src/pathfinding_service/claim_fees.py
+++ b/src/pathfinding_service/claim_fees.py
@@ -9,7 +9,7 @@ from web3.contract import Contract
 from pathfinding_service.database import PFSDatabase
 from pathfinding_service.model import IOU
 from raiden.utils.typing import BlockNumber, TokenAmount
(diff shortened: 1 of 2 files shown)


src/pathfinding_service/claim_fees.py
```diff
@@ -59,7 +59,7 @@ def main(
     contract_manager = ContractManager(contracts_precompiled_path())
     one_to_n_contract = web3.eth.contract(
         abi=contract_manager.get_contract_abi(CONTRACT_ONE_TO_N),
-        address=contract_infos['user_deposit_contract_address'],
+        address=contract_infos[CONTRACT_USER_DEPOSIT],
```
palango: Damn, I should have seen this.

PR #236, opened 2019-04-05
Small logging improvements for PFS update messages

palango:
Log more infos when `UpdatePFS` messages are received.

### DIFF
--- a/src/pathfinding_service/service.py
+++ b/src/pathfinding_service/service.py
@@ -281,25 +281,13 @@ class PathfindingService(gevent.Greenlet):
             try:
                 self.on_pfs_update(message)
             except InvalidCapacityUpdate as x:
-                log.info(
-                    str(x),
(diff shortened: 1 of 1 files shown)


src/pathfinding_service/service.py
```diff
@@ -352,13 +343,18 @@ def on_pfs_update(self, message: UpdatePFS) -> None:
             other_participant=other_participant,
         )
 
-        valid_nonces = (
+        known_nonces = (
             message.updating_nonce <= view_to_partner.update_nonce
             and message.other_nonce <= view_from_partner.update_nonce
         )
-        if valid_nonces:
+        if known_nonces:
```
karlb: I would expect `known_nonces` to be a list of nonces. Maybe `are_nonces_known` or `is_any_nonce_new` (with the opposite value)?

PR #238, opened 2019-04-08
Switch blockchain listener into freestanding function

palango:
There's no need to have a `BlockchainListener` class when a freestanding funcion works as well. Even `pylint` tells us that 😄

### DIFF
--- a/src/monitoring_service/blockchain.py
+++ b/src/monitoring_service/blockchain.py
@@ -100,159 +100,177 @@ def query_blockchain_events(
     ]
 
 
-# TODO: maybe pylint is right and this class should just be a `get_events` function?
-class BlockchainListener:  # pylint: disable=too-few-public-methods
-    """ This is pull-based blockchain listener. """
+def _get_token_network_registry_events(
+    web3: Web3,
+    contract_manager: ContractManager,
+    registry_address: Address,
+    from_block: BlockNumber,
+    to_block: BlockNumber,
+) -> List[Dict]:
+    return query_blockchain_events(
+        web3=web3,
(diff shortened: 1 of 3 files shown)


src/monitoring_service/blockchain.py
```diff
@@ -100,159 +100,177 @@ def query_blockchain_events(
...
-        new_chain_state = deepcopy(chain_state)
-        log.info('Querying new block(s)', from_block=from_block, end_block=to_block)
 
-        # first check for new token networks and add to state
-        registry_events = self._get_token_network_registry_events(
-            registry_address=new_chain_state.token_network_registry_address,
-            from_block=from_block,
-            to_block=to_block,
-        )
+def _get_monitoring_service_events(
```
karlb: Is it worth it to have these three `_get_..._events` functions? Their signature is about as long as their implementation.

src/monitoring_service/blockchain.py
```diff
@@ -100,159 +100,177 @@ def query_blockchain_events(
...
                 events.append(
-                    ReceiveMonitoringRewardClaimedEvent(
-                        ms_address=event['args']['ms_address'],
-                        amount=event['args']['amount'],
-                        reward_identifier=encode_hex(event['args']['reward_identifier']),
+                    ReceiveChannelSettledEvent(
+                        token_network_address=event['address'],
+                        channel_identifier=event['args']['channel_identifier'],
                         block_number=block_number,
                     )
```
karlb: You could pull `token_network_address`, `channel_identifier` and `block_number` into a `kwargs` dict used for all events.

PR #240, opened 2019-04-08
Refactor handling of blockchain args in CLI and constructors

karlb:
I didn't get as far as I wanted, but I have to wrap it up for now to get some of the "real" tickets done.

### DIFF
--- a/src/monitoring_service/cli.py
+++ b/src/monitoring_service/cli.py
@@ -17,7 +17,9 @@ from raiden_libs.contract_info import START_BLOCK_ID
 log = structlog.get_logger(__name__)
 
 
-@blockchain_options()
+@blockchain_options(
+    contracts=[CONTRACT_TOKEN_NETWORK_REGISTRY, CONTRACT_USER_DEPOSIT, CONTRACT_MONITORING_SERVICE]
+)
 @click.command()
 @click.option(
     '--min-reward',
--- a/src/monitoring_service/cli.py
+++ b/src/monitoring_service/cli.py
@@ -1,5 +1,6 @@
 import click
 import structlog
+from web3 import Web3
 
 from monitoring_service.constants import DEFAULT_REQUIRED_CONFIRMATIONS
 from monitoring_service.service import MonitoringService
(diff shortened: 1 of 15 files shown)


src/monitoring_service/cli.py
```diff
@@ -10,57 +11,38 @@
...
     private_key: str,
     state_db: str,
-    eth_rpc: str,
-    registry_address: Address,
-    user_deposit_contract_address: Address,
-    start_block: BlockNumber,
-    monitor_contract_address: Address,
+    web3: Web3,
+    contract_infos: dict,
+    confirmations: BlockNumber,
```
palango: Not for this PR, but I think we should start to differentiate between absolute `BlockNumber`s and relative `BlockInterval`s. I think there's something similar in Raiden.

src/pathfinding_service/claim_fees.py
```diff
@@ -9,15 +9,15 @@
...
+from raiden_contracts.constants import CONTRACT_ONE_TO_N
 from raiden_contracts.contract_manager import ContractManager, contracts_precompiled_path
 from raiden_libs.cli import blockchain_options, common_options
 from raiden_libs.utils import private_key_to_address
 
 log = structlog.get_logger(__name__)
 
 
-@blockchain_options()
+@blockchain_options(contracts=[CONTRACT_ONE_TO_N])
```
palango: Don't we need the user deposit contract as well? Otherwise we should check the deposit before sending the transaction. Could you open a ticket in that case?

src/raiden_libs/cli.py
```diff
@@ -133,23 +116,42 @@ def blockchain_options(contracts_version: str = None) -> Callable:
         ),
     ]
 
+    arg_for_contract = {
+        CONTRACT_TOKEN_NETWORK_REGISTRY: 'registry',
```
palango: Maybe we should make the name more specific: `token-network-registry` and use the `contract`-suffix as well

src/monitoring_service/cli.py
```diff
@@ -37,7 +36,8 @@ def main(
     private_key: str,
     state_db: str,
     web3: Web3,
-    contract_infos: dict,
+    contracts: dict,
```
palango: Can we specify this more, below you use `Dict[str, Contract]`

src/pathfinding_service/claim_fees.py
```diff
@@ -36,17 +35,15 @@ def main(
     private_key: str,
     state_db: str,
     web3: Web3,
-    contract_infos: dict,
+    contracts: dict,
```
palango: Same here: `Dict[str, Contract]`

src/pathfinding_service/cli.py
```diff
@@ -45,7 +45,8 @@ def main(
     private_key: str,
     state_db: str,
     web3: Web3,
-    contract_infos: dict,
+    contracts: dict,
```
palango: And here: `Dict[str, Contract]`

src/pathfinding_service/claim_fees.py
karlb: The UDC is used by the OneToN contract. There's no need to use it directly, since the claim function  returns the actually claimable amount. That's the reason why we do a `call` first only when the returned value is large enough a `transact`.

src/raiden_libs/cli.py
karlb: I didn't want to break compatibility in this PR. We would not only need doc changes, but also break existing scenarios. Do we still specify contract addresses in the scenarios we use, or do we already use the default addresses of the official deployments?

palango: I'm not sure, but I'd say better do it now than later.

palango: But we can also do this in a separate PR.

PR #242, opened 2019-04-09
Rework PFS blockchain listening

### LINKED ISSUE #230: Replace PFS BlockchainListener with stateless variant
## Problem Definition

Currently we have stateful `BlockchainListener`s in the PFS. This makes it hard to debug problems (as seen in #219, #223 ) but also complicates persistence of the state.

We need to adopt the stateless model used in the MS. 

Part of #10 

Related to #84

palango:
This PR moves the PFS to a blockchain listener based on a functional
approach that is also used in the MS.

Besides simplifying the PFS code it also removes a lot of complexity
from the BlockchainListener and paves to way for PFS state persistency.

This is still rough around some edges, but I think we can improve that in follow-up PRs.

Fixes #230


src/pathfinding_service/service.py
```diff
@@ -114,162 +106,152 @@ def __init__(
...
             self.handle_channel_new_deposit(event)
-        elif event_name == ChannelEvent.CLOSED:
+        elif isinstance(event, ReceiveChannelClosedEvent):
             self.handle_channel_closed(event)
         else:
             log.debug('Unhandled event', evt=event)
 
-    def handle_channel_opened(self, event: Dict) -> None:
-        token_network = self._get_token_network(event['address'])
+    def handle_token_network_created(self, event: ReceiveTokenNetworkCreatedEvent) -> None:
```
palango: This is something we might want to change:
Currently we create the `TokenNetwork` when we receive the according event. However, until the first channel this doesn't make to much sense, so it might be nice to create the network on demand when the first channel is opened.

src/monitoring_service/blockchain.py
```diff
@@ -179,40 +201,41 @@ def get_blockchain_events(
...
-        contract_address=new_chain_state.monitor_contract_address,
-        contract_name=CONTRACT_MONITORING_SERVICE,
-        topics=[None],
-        from_block=from_block,
-        to_block=to_block,
-    )
-    for event in monitoring_service_events:
-        event_name = event['event']
-        block_number = event['blockNumber']
+    if query_ms:
```
karlb: This branch would be a good candidate for a separate function.

src/pathfinding_service/service.py
```diff
@@ -114,162 +106,152 @@ def __init__(
...
-
     def _run(self) -> None:  # pylint: disable=method-hidden
         register_error_handler(error_handler)
         self.matrix_listener.start()
-        self.token_network_registry_listener.start()
-        self.is_running.wait()
+        while not self.is_running.is_set():
+            last_confirmed_block = self.web3.eth.blockNumber - self.required_confirmations
+
+            last_query_interval_block = (
```
karlb: I had problems with this name because I read it as "last (query_interval_block)" and didn't know what a query interval block should be. Might just be me, though.

src/pathfinding_service/service.py
```diff
@@ -114,162 +106,152 @@ def __init__(
...
-            channel_identifier=channel_identifier,
-            participant1=participant1,
-            participant2=participant2,
-            settle_timeout=settle_timeout,
-            event_block_number=event_block_number,
+            channel_identifier=event.channel_identifier,
+            participant1=event.participant1,
+            participant2=event.participant2,
+            settle_timeout=event.settle_timeout,
+            event_block_number=event.block_number,
```
karlb: Why not just `log.info('Received ChannelOpened event', event=event)`?

tests/pathfinding/test_capacity_updates.py
```diff
@@ -14,17 +14,16 @@
 from pathfinding_service.model import TokenNetwork
 from raiden.constants import UINT256_MAX
 from raiden.messages import UpdatePFS
-from raiden.utils import CanonicalIdentifier
+from raiden.utils import CanonicalIdentifier, typing
```
karlb: Do we want to decide on using either
```
from raiden.utils import typing
```
or 
```
from raiden.utils.typing import ...
```
? Or is there a benefit of doing it on a case by case basis?

palango: The problem in this file is that we need both `raiden.utils.typing.TokenNetworkAddress`  and `raiden_libs.types.TokenNetworkAddress`. I'll change the import name to ` raiden_typing`  to make this clearer.

karlb: Hmm, we could also use `from raiden.utils.typing import TokenNetworkAddress as TokenNetworkAddressBytes`.

palango: even nicer

PR #245, opened 2019-04-09
Receive and store fee info in PFS

### LINKED ISSUE #233: Receive channel fee infos
## Description

**As Josh I want the PFS to be able to**

read channel fee updates from the broadcasting room

**So that it can**

take that information into account when calculating routes and return it to me.

## Acceptance Criteria

- [ ] TBD

## Related story / issue
- Raiden mirror issue: https://github.com/raiden-network/raiden/issues/3754

palango:
Updates raiden commit and store the mediation fee info.

Fixes #233

### DIFF
--- a/src/pathfinding_service/model/channel_view.py
+++ b/src/pathfinding_service/model/channel_view.py
@@ -2,8 +2,8 @@ from enum import Enum
 
 from eth_utils import is_checksum_address
 
-from pathfinding_service.config import DEFAULT_PERCENTAGE_FEE, DEFAULT_REVEAL_TIMEOUT
-from raiden.utils.typing import ChannelID
+from pathfinding_service.config import DEFAULT_REVEAL_TIMEOUT
+from raiden.utils.typing import ChannelID, FeeAmount
 from raiden_libs.types import Address
 
 
@@ -34,6 +34,7 @@ class ChannelView:
 
         self._deposit = deposit
         self._capacity = deposit
+        self.mediation_fee = FeeAmount(0)
         self.state = ChannelView.State.OPEN
         self.channel_id = channel_id
         self.settle_timeout = settle_timeout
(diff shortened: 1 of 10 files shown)


src/pathfinding_service/model/channel_view.py
```diff
@@ -2,8 +2,8 @@
 
 from eth_utils import is_checksum_address
 
-from pathfinding_service.config import DEFAULT_PERCENTAGE_FEE, DEFAULT_REVEAL_TIMEOUT
-from raiden.utils.typing import ChannelID
+from pathfinding_service.config import DEFAULT_REVEAL_TIMEOUT
+from raiden.utils.typing import ChannelID, FeeAmount
```
karlb: I feel like using `FeeAmount` instead of using `TokenAmount` is going a bit too far regarding the detailed typing. We don't need a separate type for each variable. And since we can't use inheritance on `NewType`s, a function like `format_token_amount` will have an ugly signature or require type casts when called on `FeeAmount`. Not that much of a problem, but we should remember to not just think that more different types is always better.

src/pathfinding_service/model/channel_view.py
```diff
@@ -63,8 +71,8 @@ def capacity(self) -> int:
         return self._capacity
 
     @property
-    def relative_fee(self) -> int:
-        return DEFAULT_PERCENTAGE_FEE
+    def mediation_fee(self) -> FeeAmount:
+        return self._mediation_fee
```
karlb: We don't need a property, we can just use `self.mediation_fee` directly.

src/pathfinding_service/model/channel_view.py
palango: This is something coming from Raiden. I think there it makes a lot more sense to keep stuff correct in the core.

As for `format_token_amount`, a `Union` should be fine, right?

karlb: Yes, but if we go too far it will be a union of
* TokenAmount
* MediationFeeAmount
* PFSFeeAmount
* MSRewardAmount
* ...

and steadily growing.

Not much of a problem for this PR, I just wanted to mention it before we've invented even more detailed types. Should we also switch to FeeAmount for PFS fees, or is it for mediation fees only?

palango: At least in the client the `FeeAmount` is special, and there are (currently) no types for `[MS|PFS]FeeAmount`.

PR #251, opened 2019-04-10
add initial pfs happy path scenario

czepluch:
This PR add a minimal viable happy path scenario for the PFS. 

Once this scenario passes, I think we have a some working version of the PFS.

Currently it doesn't work since the `assert_pfs_history` task doesn't work yet.

### DIFF
--- a/scenarios/simple_pfs_integration_without_reward.yml
+++ b/scenarios/simple_pfs_integration_without_reward.yml
@@ -1,0 +1,84 @@
+version: 2
+
+settings:
+  gas_price: "fast"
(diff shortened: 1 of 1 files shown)


czepluch:
Afaik, the `assert_pfs_history` task does not yet work, because it's missing a debug endpoint from the pfs API, so it'd fail. But we can comment those out and see if it works.

Dominik1999:
Exactly, we didn't implement the endpoint yet. But we can use the scenario anyway also when it would fail just to see if the PFS receives and provides everything it needs

PR #255, opened 2019-04-12
Add debug endpoint

Dominik1999:


src/pathfinding_service/api.py
```diff
@@ -33,7 +33,7 @@
 from raiden_libs.types import Address, TokenNetworkAddress
 
 log = structlog.get_logger(__name__)
-
+g.last_request = []
```
karlb: The `g` object is only usable after the flask application has been initialized, because it belongs to the application context. Instead use
```python
if 'last_request' no in g:
    g.last_request = []
```
where using that attribute,

src/pathfinding_service/api.py
```diff
@@ -122,6 +122,18 @@ def post(self, token_network_address: str) -> Tuple[dict, int]:
                 **optional_args,
             )
         except (NetworkXNoPath, NodeNotFound):
+
+            ## this is for assertion via the scenario player
+            if len(g.last_request) > 20: g.last_request.pop(0)
```
karlb: You might want to use `collections.deque` with `maxlen=20`: https://docs.python.org/2/library/collections.html#collections.deque

Dominik1999: thanks, that is a nice object

src/pathfinding_service/api.py
Dominik1999: done

src/pathfinding_service/api.py
```diff
@@ -33,12 +34,14 @@
...
 log = structlog.get_logger(__name__)
+# change maxlen in case more than 20 paths are requested
+last_requests: collections.deque = collections.deque([], maxlen=20)
 
 
 class ApiWithErrorHandler(Api):
     def handle_error(self, e: exceptions.ApiException) -> Response:
         return self.make_response(
-            {'errors': e.msg, 'error_code': e.error_code, 'error_details': e.error_details},
+            {"errors": e.msg, "error_code": e.error_code, "error_details": e.error_details},
```
czepluch: According to the Raiden contributor's guidelines this should be single quotes and not double quotes: https://github.com/raiden-network/raiden/blob/master/CONTRIBUTING.md

karlb: I heard raiden is going to use black. If that's true, that part of the code style will probably change, since black prefers double quotes: https://github.com/ambv/black#strings .

I mostly don't mind, so don't count this as a vote for any of the choices.

Dominik1999: Yes this is black preferring the double quotes. I don't care tbh. can we maybe change the black settings somehow @karlb ?

karlb: Right now, we are running black with `--skip-string-normalization` which means it does not change the quotes. I don't think there is a switch to prefer single quotes due to black's general "any color you want -- as long as it's black"-attitude.

palango: Given these changes in PRs, should we just switch to "real" black and remove the `--skip-string-normalization` option?

Dominik1999:
@palango it now works, could you do a final review?

src/pathfinding_service/api.py
```diff
@@ -33,12 +34,14 @@
 from raiden_libs.types import Address, TokenNetworkAddress
 
 log = structlog.get_logger(__name__)
+# change maxlen in case more than 20 paths are requested
+last_requests: collections.deque = collections.deque([], maxlen=20)
```
ulope: 20 could be a bit small for large scenarios

src/pathfinding_service/api.py
```diff
@@ -52,18 +55,18 @@ def _validate_token_network_argument(
...
-            return {'errors': address_error.format(token_network_address)}, 400
+            address_error = "Token network address not checksummed: {}"
+            return {"errors": address_error.format(token_network_address)}, 400
 
         token_network = self.pathfinding_service.get_token_network(
             TokenNetworkAddress(token_network_address)
         )
         if token_network is None:
-            return {'errors': 'Unsupported token network: {}'.format(token_network_address)}, 400
+            return ({"errors": "Unsupported token network: {}".format(token_network_address)}, 400)
```
ulope: Unnecessary parentheses are unnecessary ;)

src/pathfinding_service/api.py
```diff
@@ -225,30 +234,44 @@ def get(self, token_network_address: TokenNetworkAddress) -> Tuple[dict, int]:
...
             },
             200,
         )
 
 
+class DebugEndpoint(PathfinderResource):
+    def get(  # pylint: disable=no-self-use
+        self, token_network_address: TokenNetworkAddress, source_address: Address
+    ) -> Tuple[dict, int]:
+        request_count = 0
```
ulope: The idea of the `request_count` in the SP was to be able to differentiate between failed and successful requests. In this implementation it's basically just `len(responses) - 1`.

src/pathfinding_service/api.py
```diff
@@ -225,30 +234,44 @@ def get(self, token_network_address: TokenNetworkAddress) -> Tuple[dict, int]:
...
+                "operator": operator,
+                "message": message,
             },
             200,
         )
 
 
+class DebugEndpoint(PathfinderResource):
+    def get(  # pylint: disable=no-self-use
+        self, token_network_address: TokenNetworkAddress, source_address: Address
```
ulope: This needs to able to optionally handle `target_address` as well. 
The SP `assert_pfs_history` task passes that as the last path segment if `to: <int>` is given in the task config.

src/pathfinding_service/api.py
```diff
@@ -225,30 +234,44 @@ def get(self, token_network_address: TokenNetworkAddress) -> Tuple[dict, int]:
...
+    ) -> Tuple[dict, int]:
+        request_count = 0
+        responses = []
+        for r in last_requests:
+            log.debug("Last Requests Values:", r=r)
+            if token_network_address == r[0] and source_address == r[1]:
```
ulope: This isn't very pythonic. 

The elements you insert into `last_requests` above could instead already be dicts with with correct signature, then you'd only need to filter them here and not also transform the tuples.

src/pathfinding_service/api.py
```diff
@@ -257,21 +280,22 @@ def __init__(self, pathfinding_service: PathfindingService):
         self.pathfinding_service = pathfinding_service
 
         resources: List[Tuple[str, Resource, Dict]] = [
-            ('/<token_network_address>/paths', PathsResource, {}),
-            ('/<token_network_address>/payment/iou', IOUResource, {}),
-            ('/info', InfoResource, {}),
+            ("/<token_network_address>/paths", PathsResource, {}),
+            ("/<token_network_address>/payment/iou", IOUResource, {}),
+            ("/info", InfoResource, {}),
+            ("/_debug/routes/<token_network_address>/<source_address>", DebugEndpoint, {}),
```
ulope: As mentioned above, this also needs to optionally support `target_address`.

src/pathfinding_service/api.py
Dominik1999: changed it to 200

src/pathfinding_service/api.py
Dominik1999: removed

src/pathfinding_service/api.py
Dominik1999: hmm the failed request would now count as a request with set to `"responses": [ ]`. So the count would not include requests that did not reach the pfs. What do you suggest?

scenarios/simple_pfs_integration_without_reward.yml
```diff
@@ -24,6 +24,7 @@ nodes:
     environment-type: development
     routing-mode: pfs
     pathfinding-max-paths: 5
+    pathfinding-eth-address: 0x308a33011c45bfa0a801fcd16100e80629b25e1f
```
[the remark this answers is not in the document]
Dominik1999: it seemed to work without as well, but I added them

scenarios/simple_pfs_integration_without_reward.yml
```diff
@@ -5,7 +5,7 @@ settings:
   chain: kovan
   services:
     pfs:
-      url: https://pfs-kovan.services-dev.raiden.network
+      url: http://localhost:6000
```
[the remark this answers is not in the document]
Dominik1999: yes that should be the goal. I startet the pfs locally to see the logs

src/pathfinding_service/api.py
```diff
@@ -122,16 +125,22 @@ def post(self, token_network_address: str) -> Tuple[dict, int]:
...
+                    "errors": "No suitable path found for transfer from {} to {}.".format(
                         path_req.from_, path_req.to
                     )
                 },
                 400,
             )
 
-        return {'result': paths}, 200
+        # this is for assertion via the scenario player
+        last_requests.append([token_network_address, path_req.from_, path_req.to, paths])
```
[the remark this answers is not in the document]
Dominik1999: ok maybe you can show me tomorrow how to use that dataclass

src/pathfinding_service/api.py
Dominik1999: good point, changed it

src/pathfinding_service/api.py
Dominik1999: so whenever the pfs throws `NetworkXNoPath` or `NodeNotFound` the debug endpoint adds a request and an empty list [ ] to the request deque. I am not sure how we could track failed requests on the pfs side apart from that

Dominik1999: we take the `request_count` for now as it is and improve later :)

tests/pathfinding/test_rest.py
```diff
@@ -22,6 +22,58 @@
...
+            'target': addresses[2],
+            'routes': [{'path': [addresses[0], addresses[1], addresses[2]], 'estimated_fee': 0}],
+        }
+    ]
+
+    # now there must be a debug endpoint for that specific route and that specific target
+    response_debug_incl_target = requests.get(url_debug_incl_target)
+    assert response_debug_incl_target.status_code == 200
+    request_count = response_debug_incl_target.json()['request_count']
+    assert request_count == 1
```
palango: I'd like to see that the `target_address`  is actually used to filter something out. Can you add the same test with a different `target_address`  below, so we see that `request_count`  becomes 0.

Dominik1999: done

src/pathfinding_service/api.py
```diff
@@ -209,53 +231,91 @@ def get(self, token_network_address: TokenNetworkAddress) -> Tuple[dict, int]:
...
+            log.debug("Last Requests Values:", r=r)
+            matches_params = is_same_address(
+                token_network_address, r['token_network_address']
+            ) and is_same_address(source_address, r['source'])
+            if target_address is not None:
+                matches_params = matches_params and is_same_address(target_address, r['target'])
+
+            if matches_params:
+                request_count += 1
+                responses.append(dict(source=r['source'], target=r['target'], routes=r['routes']))
```
ulope: @Dominik1999 You can just use `responses.append(r)`

PR #262, opened 2019-04-16
Save and load state in PFS

### LINKED ISSUE #10: Persistence
The PFS should store its state when shutting down cleanly. This includes channels, channel balances and fees.

In the case of an unexpected shutdown the PFS can sync from the start.
 
## ToDo
- [x] Decide on datastore (decision: sqlite)
- [x] Store state of blockchain sync (maybe replace the blockchain syncing code with the more functional code from MS, see #230)
- [x] Store token networks and channel infos
- [x] Restore network graph before starting BC listening and matrix client
- [x] ~~Save Matrix syncing state~~ (not possible right now)

karlb:
Closes #10.

There' some refactoring work left for the `BlockchainState`, but that can be done in a different PR.

### DIFF
--- a/src/pathfinding_service/database.py
+++ b/src/pathfinding_service/database.py
@@ -112,7 +112,8 @@ class PFSDatabase(BaseDatabase):
                 printf('0x%064x', :reveal_timeout),
                 printf('0x%064x', :deposit),
                 printf('0x%064x', :update_nonce),
-                :absolute_fee, :relative_fee
+                printf('0x%064x', :absolute_fee),
+                :relative_fee
             )
         """,
             cv_dict,
--- a/src/pathfinding_service/database.py
+++ b/src/pathfinding_service/database.py
@@ -50,6 +50,12 @@ class PFSDatabase(BaseDatabase):
             iou_dict,
         )
 
+    def get_latest_known_block(self) -> BlockNumber:
+        return self.conn.execute("SELECT latest_known_block FROM blockchain").fetchone()[0]
+
+    def update_lastest_known_block(self, latest_known_block: BlockNumber) -> None:
+        self.conn.execute("UPDATE blockchain SET latest_known_block = ?", [latest_known_block])
+
     def get_ious(
         self,
         sender: Address = None,
--- a/src/pathfinding_service/database.py
+++ b/src/pathfinding_service/database.py
@@ -5,6 +5,7 @@ import structlog
 
 from pathfinding_service.model import IOU
 from pathfinding_service.model.channel_view import ChannelView
+from pathfinding_service.model.token_network import TokenNetwork
 from raiden.utils.typing import BlockNumber, ChannelID, TokenAmount
 from raiden_libs.database import BaseDatabase, hex256
 from raiden_libs.types import Address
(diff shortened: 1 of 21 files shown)


tests/pathfinding/test_claim_fees.py
```diff
@@ -96,5 +97,7 @@ def mock_connect_to_blockchain(monkeypatch):
 @pytest.mark.usefixtures('mock_connect_to_blockchain')
 def test_cli(default_cli_args):
     runner = CliRunner()
-    result = runner.invoke(main, default_cli_args, catch_exceptions=False)
+    result = runner.invoke(
+        main, default_cli_args + ['--state-db=:memory:'], catch_exceptions=False
```
palango: Can we put `--state-db=:memory:` into `default_cli_args`

src/pathfinding_service/model/channel_view.py
```diff
@@ -17,30 +19,22 @@ class State(Enum):
...
-        assert is_checksum_address(participant1)
-        assert is_checksum_address(participant2)
-
-        self.self = participant1
-        self.partner = participant2
+    channel_id: ChannelID
+    participant1: Address
+    participant2: Address
+    settle_timeout: int
+    capacity: int = field(init=False)
```
palango: Should be a `TokenAmount`

src/pathfinding_service/model/channel_view.py
```diff
@@ -17,30 +19,22 @@ class State(Enum):
...
+    reveal_timeout: int = DEFAULT_REVEAL_TIMEOUT
+    deposit: TokenAmount = TokenAmount(0)
+    state: State = State.OPEN
+    update_nonce: int = field(default=0, init=False)
```
palango: Should be type `Nonce`

src/pathfinding_service/database.py
```diff
@@ -86,3 +87,35 @@ def get_iou(
             return next(self.get_ious(sender, expiration_block, claimed))
         except StopIteration:
             return None
+
+    def upsert_channel_view(self, channel_view: ChannelView) -> None:
+        cv_dict = ChannelView.Schema(strict=True, exclude=['state']).dump(channel_view)[0]
+        self.conn.execute(
+            """
+            INSERT OR REPLACE INTO channel_view (
+                token_network_address, channel_id, participant1, participant2,
```
palango: I wonder if we should add the `token_network_registry` address here as well, just to be future safe. On the other hand this would request big changes anyways all over the client. so probably nvm.

src/pathfinding_service/model/channel_view.py
```diff
@@ -19,22 +23,25 @@ class State(Enum):
...
     participant2: Address
     settle_timeout: int
-    capacity: int = field(init=False)
+    capacity: int = None  # type: ignore
     reveal_timeout: int = DEFAULT_REVEAL_TIMEOUT
     deposit: TokenAmount = TokenAmount(0)
-    state: State = State.OPEN
-    update_nonce: int = field(default=0, init=False)
-    absolute_fee = FeeAmount(0)
+    state: State = State.OPEN  # TODO: do we care about channels with state != OPEN?
```
palango: Good point, is this used/checked anywhere?

src/pathfinding_service/schema.sql
```diff
@@ -5,12 +5,29 @@ CREATE TABLE blockchain (
...
+    token_network_address   CHAR(42) NOT NULL,
+    channel_id      HEX_INT NOT NULL,
+    participant1    CHAR(42) NOT NULL,
+    participant2    CHAR(42) NOT NULL,
+    settle_timeout  HEX_INT NOT NULL,
+    capacity        HEX_INT NOT NULL,
+    reveal_timeout  HEX_INT NOT NULL,
+    deposit         HEX_INT NOT NULL,
+    update_nonce    HEX_INT,
+    absolute_fee    INT,
```
palango: SHould this be `HEX_INT`  as well? We never know how fees will behave... -.-

src/pathfinding_service/model/channel_view.py
karlb: No, and it's removed in another commit of this PR.

src/pathfinding_service/schema.sql
karlb: I've only used HEX_INT for those values that touch solidity. But since MAXINT can only represent up to ~9 RDN using HEX_INT is probably safer.

tests/pathfinding/test_claim_fees.py
karlb: Yes, actually we already did. There are just two different `default_cli_args`. I'll make a shared one.

PR #273, opened 2019-04-17
Add some logging to Flask requests

palango:
For some reason Flask still logs on it's own, and structlog doesn't pick up that.


karlb:
I don't think structlog is supposed to "pick up" flasks logging. It looks like the usual solution is to configure them in a way that both logging and structlog behave in the same way: http://www.structlog.org/en/stable/standard-library.html#suggested-configurations

Issue #269, opened 2019-04-17
PFS doesn't stop with CTRL-C


palango:
## Problem Definition

The running PFS is currently hard to stop, as CTRL-C work only during a little time interval.

PR #278, opened 2019-04-18
 Do not ignore system errors and unfiy error handling

karlb:
Closes https://github.com/raiden-network/raiden-services/issues/269.

### DIFF
--- a/src/monitoring_service/service.py
+++ b/src/monitoring_service/service.py
@@ -25,7 +25,6 @@ from raiden_contracts.constants import (
 from raiden_libs.blockchain import get_blockchain_events
 from raiden_libs.contract_info import CONTRACT_MANAGER
 from raiden_libs.events import Event
-from raiden_libs.gevent_error_handler import register_error_handler
 from raiden_libs.utils import private_key_to_address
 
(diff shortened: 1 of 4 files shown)


src/raiden_libs/gevent_error_handler.py
```diff
@@ -1,20 +1,33 @@
...
+        log.info("Service termination requested by user.")
+        sys.exit()
+
+    log.critical(
+        "Unhandled exception. Terminating the program..."
+        "Please report this issue at "
+        "https://github.com/raiden-network/raiden-services/issues"
+    )
+    traceback.print_exception(etype=type, value=value, tb=tb)
+    sys.exit(-1)
```
palango: Is a negative value correct?

src/monitoring_service/service.py
```diff
@@ -104,6 +105,7 @@ def __init__(
     def start(
         self, wait_function: Callable = time.sleep, check_account_gas_reserve: bool = True
     ) -> None:
+        register_error_handler()
```
palango: Why do we need this here? There's no gevent at the moment?

PR #281, opened 2019-04-18
Update README.md

palango:
Recommend development version for now

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,10 +26,13 @@ For more technical details see:
 
 The Raiden Services require Python 3.7+.
 
-To install the Raiden services run the following commands:
+Currently we recomment to use the latest development version of the services. To install them run the following commands:
 
 ```sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,10 +26,13 @@ For more technical details see:
...
-To install the Raiden services run the following commands:
+Currently we recomment to use the latest development version of the services. To install them run the following commands:
 
 ```sh
+git clone git@github.com:raiden-network/raiden-services.git
+cd raiden-services
 virtualenv -p python3.7 venv
 source venv/bin/activate
-pip install raiden-services
+pip install -U -r requirements-dev.txt
```
Dominik1999: why do we need the -U ? I always run it without

palango: This allows to upgrade dependencies, but shouldn't be necessary here.

PR #292, opened 2019-04-23
Fix sqlite error on values larger than maxuint64

karlb:
We passed the values into sqlite as int, which caused SQLite to reject
then input values as too large if they were above 2^64. Fixed by passing
the values in as string.


src/pathfinding_service/database.py
```diff
@@ -38,13 +38,14 @@ def __init__(
 
     def upsert_iou(self, iou: IOU) -> None:
         iou_dict = IOU.Schema(strict=True).dump(iou)[0]
+        for key in ('amount', 'expiration_block'):
+            iou_dict[key] = hex(iou_dict[key])
```
palango: hex256 here as well

karlb: Thanks!

PR #293, opened 2019-04-23
Further deployment fixes

palango:
Some stuff that I missed.

### DIFF
--- a/deployment/docker-compose.override.yml
+++ b/deployment/docker-compose.override.yml
@@ -5,9 +5,9 @@ version: '3.4'
 x-defaults: &defaults
   restart: "no"
   volumes:
-    - ~/services-dev/state:/state
+    - ./state:/state
     - ~/.ethereum/keystore:/keystore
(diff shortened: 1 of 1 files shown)


deployment/docker-compose.override.yml
```diff
@@ -5,9 +5,9 @@ version: '3.4'
 x-defaults: &defaults
   restart: "no"
   volumes:
-    - ~/services-dev/state:/state
+    - ~/state:/state
```
karlb: I think we can use
```
- ./state:/state
- ../:/services
```
to make this work regardless of where it is installed and to make sure the `state` directory already exists. It might be considered a bad thing to store the state inside the git clone, though.

palango: Good points, I changed it.

palango:
Other thing: Do we have randomness in tests?

```
- Hits         1590     1584       -6     
- Misses        170      172       +2    
```

PR #294, opened 2019-04-23
Disable debug API by default.

### LINKED ISSUE #285: Disable `_debug` endpoints by default
We added debug endpoints to make system testing possible. However, these endpoints should be disabled by default.

palango:
It can be enabled by running the PFS with `--enable-debug`

Fixes #285


src/pathfinding_service/cli.py
```diff
@@ -43,6 +43,13 @@
     type=click.IntRange(min=0),
     help="Number of block confirmations to wait for",
 )
+@click.option(
+    "--enable-debug",
+    default=False,
+    is_flag=True,
+    hidden=True,
+    help="Number of block confirmations to wait for",
```
ulope: That’s probably not the right help text ;)

PR #301, opened 2019-04-25
uses better address comparison (checksum-agnostic) and adds some logs

Dominik1999:
Closes a minor bug regarding IOU acceptance, addresses were checked but the check failed for same addresses if one was not checksummed.


src/pathfinding_service/api.py
```diff
@@ -100,6 +100,7 @@ class PathsResource(PathfinderResource):
     def post(self, token_network_address: str) -> Tuple[dict, int]:
         token_network = self._validate_token_network_argument(token_network_address)
         path_req = self._parse_post(PathRequest)
+        log.info("Received path request", request=path_req)
```
palango: I'd remove that log again.

Dominik1999: ok

PR #302, opened 2019-04-25
Test more edge cases in PFS service

palango:
Add some tests for the event handlers in PFS service.

### DIFF
--- a/tests/pathfinding/test_service.py
+++ b/tests/pathfinding/test_service.py
@@ -17,6 +17,9 @@ from raiden_libs.types import Address, TokenNetworkAddress
 
 from ..libs.mocks.web3 import ContractMock, Web3Mock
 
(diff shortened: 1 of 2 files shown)


tests/pathfinding/test_service.py
```diff
@@ -106,17 +107,137 @@ def new_service(filename):
...
+        channel_identifier=ChannelID(1),
+        participant1=participant1,
+        participant2=participant2,
+        settle_timeout=20,
+        block_number=BlockNumber(1),
+    )
+
+    pathfinding_service_mock.handle_event(channel_event)
+    assert len(pathfinding_service_mock.token_networks) == 1
+    assert len(token_network_model.channel_id_to_addresses) == 1
```
karlb: This part is repeated three times. Reusing it would make the tests a bit shorter.

PR #303, opened 2019-04-25
Adds debug endpoint for ious for scenario player

Dominik1999:
We want to assert the sent and stored ious in the pfs via the scenario player

### DIFF
--- a/src/pathfinding_service/api.py
+++ b/src/pathfinding_service/api.py
@@ -289,6 +289,17 @@ class DebugEndpoint(PathfinderResource):
         return dict(request_count=request_count, responses=responses), 200
 
 
+class DebugEndpointIOU(PathfinderResource):
+    def get(self, source_address: Address) -> Tuple[dict, int]:
+        iou = self.pathfinding_service.database.get_iou(source_address)
+        if iou:
+            return (
+                dict(sender=iou.sender, amount=iou.amount, expiration_block=iou.expiration_block),
+                200,
+            )
+        return {}, 404
+
+
 class ServiceApi:
     def __init__(self, pathfinding_service: PathfindingService):
         self.flask_app = Flask(__name__)
@@ -318,6 +329,7 @@ class ServiceApi:
                         {},
                         "debug2",
                     ),
+                    ("/_debug/ious/<source_address>", DebugEndpointIOU, {}, "debug3"),
                 ]
             )
 
--- a/tests/pathfinding/test_api.py
+++ b/tests/pathfinding/test_api.py
@@ -5,13 +5,14 @@ from typing import List
 import gevent
 import pkg_resources
 import requests
-from eth_utils import decode_hex, encode_hex, to_normalized_address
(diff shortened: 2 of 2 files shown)


src/pathfinding_service/api.py
```diff
@@ -289,6 +289,20 @@ def get(  # pylint: disable=no-self-use
         return dict(request_count=request_count, responses=responses), 200
 
 
+class DebugEndpointIoU(PathfinderResource):
+    def get(self, source_address: Address) -> Tuple[dict, int]:  # pylint: disable=no-self-use
+        response = []
+        ious_sender = self.pathfinding_service.database.conn.execute(
+            "select * from iou where sender = ?", [source_address]
```
palango: We usually upper case sql keywords.

palango: Oh, and can't you use `get_iou` from the database class?

src/pathfinding_service/api.py
```diff
@@ -289,6 +289,20 @@ def get(  # pylint: disable=no-self-use
         return dict(request_count=request_count, responses=responses), 200
 
 
+class DebugEndpointIoU(PathfinderResource):
```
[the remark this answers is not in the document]
Dominik1999: ok

src/pathfinding_service/api.py
Dominik1999: `"select * from iou where SENDER = ?", [source_address]` ?

Dominik1999: > Oh, and can't you use get_iou from the database class?
nope, because I want to iterate over several IOUs 

```    
for iou in ious_sender:
TypeError: 'IOU' object is not iterable
```

Dominik1999:
@palango pytest passes locally. Do you see the error?

src/pathfinding_service/api.py
```diff
@@ -289,6 +289,18 @@ def get(  # pylint: disable=no-self-use
         return dict(request_count=request_count, responses=responses), 200
 
 
+class DebugEndpointIOU(PathfinderResource):
+    def get(self, source_address: Address) -> Tuple[dict, int]:  # pylint: disable=no-self-use
```
palango: The # pylint: disable=no-self-use might be able to go away.

tests/pathfinding/test_api.py
```diff
@@ -99,6 +100,42 @@ def test_get_paths_via_debug_endpoint(
...
+    addresses: List[Address],
+    token_network_model: TokenNetwork,
+):
+    iou = IOU(
+        sender=addresses[0],
+        receiver=addresses[4],
+        amount=TokenAmount(111),
+        expiration_block=BlockNumber(7619644),
+        signature=Signature(
+            str.encode("118a93e9fd0a3a1c3d6edbad194b5c9d95715c754881d80e23e985793b1e13de")
```
palango: What doesn the `str.encode` do?

PR #309, opened 2019-04-25
Fix two reasons for flaky tests

karlb:
Structlog messes up capturing with `capsys` in some cases when not properly reset. That was hard to find.

### DIFF
--- a/src/raiden_libs/cli.py
+++ b/src/raiden_libs/cli.py
@@ -94,7 +94,7 @@ def common_options(app_name: str) -> Callable:
             func = option(func)
 
         @wraps(func)
-        def call_with_opened_keystore(**params: Any) -> Callable:
+        def call_with_common_options_initialized(**params: Any) -> Callable:
             params["private_key"] = _open_keystore(
                 params.pop("keystore_file"), params.pop("password")
             )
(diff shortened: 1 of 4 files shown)


src/raiden_libs/cli.py
```diff
@@ -100,7 +98,11 @@ def call_with_opened_keystore(**params: Any) -> Callable:
             params["private_key"] = _open_keystore(
                 params.pop("keystore_file"), params.pop("password")
             )
-            return func(**params)
+            try:
```
palango: Should this go into `call_with_opened_keystore` or a new function?

tests/pathfinding/test_api.py
```diff
@@ -162,7 +162,7 @@ def request_path_with(status_code=400, **kwargs):
 
     # bad signature
     bad_iou_dict = good_iou_dict.copy()
-    bad_iou_dict["signature"] = hex(int(bad_iou_dict["signature"], 16) + 1)
+    bad_iou_dict["signature"] = "0x" + "1" * 130
```
palango: Why did this start to bite us again?

karlb: Because it was used in two different places.

src/raiden_libs/cli.py
karlb: I think this is a good place for it. The name `call_with_opened_keystore` could use some updating, though. Something like `call_with_common_options_initialized`. But that's not a great name, either.

palango: Not perfect, but better. And I have nothing better to propose...

Issue #124, opened 2019-02-19
Add matrix testing


palango:
Add basic tests for the matrix listener.

palango:
Fixed with #307 and #310, further tests outweigh the benefit at the moment.

PR #310, opened 2019-04-25
Matrix2

palango:
Refactor and test message deserialization.

Part of #124

### DIFF
--- a/tests/libs/test_matrix.py
+++ b/tests/libs/test_matrix.py
@@ -12,7 +12,7 @@ from raiden.utils.typing import Address as BytesAddress, ChainID, Nonce, TokenAm
 from raiden_contracts.tests.utils import EMPTY_LOCKSROOT
 from raiden_libs.matrix import deserialize_messages, message_from_dict
 
-PEER_ADDRESS = BytesAddress(to_canonical_address("0x" + "1" * 40))
+INVALID_PEER_ADDRESS = BytesAddress(to_canonical_address("0x" + "1" * 40))
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


tests/libs/test_matrix.py
```diff
@@ -47,3 +56,60 @@ def test_message_from_dict(token_network, get_accounts, get_private_key):
...
+    assert len(messages) == 0
+
+
+def test_deserialize_messages_empty_lines():
+    messages = deserialize_messages(data=" \r\n ", peer_address=PEER_ADDRESS)
+    assert len(messages) == 0
+
+
+def test_deserialize_messages_invalid_json():
+    messages = deserialize_messages(data="@@@", peer_address=PEER_ADDRESS)
```
karlb: You could combine these three tests and use
```python
for data in ['', ' \r\n ', '@@@']:
    ...
```
Mostly a matter of taste, though.

tests/libs/test_matrix.py
```diff
@@ -47,3 +56,60 @@ def test_message_from_dict(token_network, get_accounts, get_private_key):
...
+
+    with patch("raiden_libs.matrix.message_from_dict", new=Mock()):
+        messages = deserialize_messages(data=json.dumps(message_json), peer_address=PEER_ADDRESS)
+        assert len(messages) == 0
+
+
+def test_deserialize_messages_invalid_sender(request_monitoring_message):
+    message_json = request_monitoring_message.to_dict()
+
+    messages = deserialize_messages(data=json.dumps(message_json), peer_address=PEER_ADDRESS)
```
karlb: Just from the variable name, I would have expected `PEER_ADDRESS` to be the correct `peer_address`.

PR #313, opened 2019-04-26
Include PFS' eth address in info endpoint

karlb:
Which allows running the Raiden client with a custom PFS without
manually passing the PFS' eth address to the client.

### DIFF
--- a/src/pathfinding_service/api.py
+++ b/src/pathfinding_service/api.py
@@ -260,6 +260,7 @@ class InfoResource(PathfinderResource):
                 "version": self.version,
                 "operator": operator,
                 "message": message,
+                "payment_address": self.pathfinding_service.address,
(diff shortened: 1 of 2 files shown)


src/pathfinding_service/api.py
```diff
@@ -260,6 +260,7 @@ def get(self) -> Tuple[dict, int]:
                 "version": self.version,
                 "operator": operator,
                 "message": message,
+                "eth_address": self.pathfinding_service.address,
```
palango: What about `payment_address`?

karlb: I wanted to put `eth` in there to make the distinction from an IP address. But payment address is better, since it's obviously an eth address and conveys the purpose.

PR #315, opened 2019-04-26
Improve event decoding

palango:
- Remove some unnecessary conditions
- Cache topic_to_event_abi mapping, to safe some cycles

Also tested with parity and geth, both work fine.

### DIFF
--- a/src/raiden_libs/blockchain.py
+++ b/src/raiden_libs/blockchain.py
@@ -2,13 +2,13 @@ from copy import deepcopy
 from typing import Dict, List, Tuple
 
 import structlog
-from eth_utils import decode_hex, encode_hex, to_checksum_address
+from eth_utils import encode_hex, to_checksum_address
 from eth_utils.abi import event_abi_to_log_topic
 from web3 import Web3
 from web3.contract import get_event_data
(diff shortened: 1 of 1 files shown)


src/raiden_libs/blockchain.py
```diff
@@ -43,25 +43,11 @@ def create_registry_event_topics(contract_manager: ContractManager) -> List:
...
-    """
-    if isinstance(log_["topics"][0], str):
-        log_["topics"][0] = decode_hex(log_["topics"][0])
-    elif isinstance(log_["topics"][0], int):
-        log_["topics"][0] = decode_hex(hex(log_["topics"][0]))
-    event_id = log_["topics"][0]
-    events = filter_by_type("event", abi)
-    topic_to_event_abi = {event_abi_to_log_topic(event_abi): event_abi for event_abi in events}
+def decode_event(topic_to_event_abi: Dict[bytes, Dict], log_entry: Dict) -> Dict:
+    event_id = log_entry["topics"][0]
```
karlb: `event_id` and `topic` look like synonyms. If they are, we should use `topic` to be consistent.

palango: Fair point 👀

PR #316, opened 2019-04-26
Claim event

palango:
Log useful infos when a ClaimReward event is received.

### DIFF
--- a/src/monitoring_service/handlers.py
+++ b/src/monitoring_service/handlers.py
@@ -310,7 +310,15 @@ def monitor_new_balance_proof_event_handler(event: Event, context: Context) -> N
 
 def monitor_reward_claim_event_handler(event: Event, context: Context) -> None:
     assert isinstance(event, ReceiveMonitoringRewardClaimedEvent)
-    log.info("Received MSC RewardClaimed event", evt=event)
+
+    if event.ms_address == context.ms_state.address:
+        log.info(
+            "Successfully claimed reward", amount=event.amount, reward_id=event.reward_identifier
+        )
+    else:
+        log.debug(
+            "Another MS claimed reward", amount=event.amount, reward_id=event.reward_identifier
+        )
 
 
 def updated_head_block_event_handler(event: Event, context: Context) -> None:
--- a/tests/monitoring/monitoring_service/test_handlers.py
+++ b/tests/monitoring/monitoring_service/test_handlers.py
@@ -130,7 +130,7 @@ def create_default_token_network(context):
     )
 
 
-def test_event_handler_ignore_other_events(context: Context,):
+def test_event_handler_ignore_other_events(context: Context):
     event = Event()
 
     for handler in [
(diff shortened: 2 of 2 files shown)


src/monitoring_service/handlers.py
```diff
@@ -310,7 +310,17 @@ def monitor_new_balance_proof_event_handler(event: Event, context: Context) -> N
 
 def monitor_reward_claim_event_handler(event: Event, context: Context) -> None:
     assert isinstance(event, ReceiveMonitoringRewardClaimedEvent)
-    log.info("Received MSC RewardClaimed event", evt=event)
+
+    if event.ms_address == context.ms_state.address:
+        log.info(
+            "Successfully claimed reward", amount=event.amount, reward_id=event.reward_identifier
+        )
+    else:
```
Dominik1999: someone told me that we should not use `else` in the Raiden code but smarter `ifs`

tests/monitoring/monitoring_service/test_handlers.py
```diff
@@ -440,6 +441,24 @@ def test_monitor_new_balance_proof_event_handler_idempotency(context: Context,):
...
+        block_number=BlockNumber(23),
+    )
+
+    monitor_reward_claim_event_handler(claim_event, context)
+    assert log.has("Successfully claimed reward")
+
+    claim_event.ms_address = Address("C")
+    monitor_reward_claim_event_handler(claim_event, context)
+    assert log.has("Another MS claimed reward event")
+
```
Dominik1999: Do we need to include a test or a log when the MS claims with the wrong token amount?

src/monitoring_service/handlers.py
```diff
@@ -310,7 +310,17 @@ def monitor_new_balance_proof_event_handler(event: Event, context: Context) -> N
...
+        log.debug(
+            "Another MS claimed reward event",
```
karlb: type: "event" does not belong here.

tests/monitoring/monitoring_service/test_handlers.py
```diff
@@ -440,6 +441,24 @@ def test_monitor_new_balance_proof_event_handler_idempotency(context: Context,):
...
+        reward_identifier="REWARD",
+        block_number=BlockNumber(23),
+    )
+
+    monitor_reward_claim_event_handler(claim_event, context)
+    assert log.has("Successfully claimed reward")
+
+    claim_event.ms_address = Address("C")
+    monitor_reward_claim_event_handler(claim_event, context)
+    assert log.has("Another MS claimed reward event")
```
karlb: `log.has` is nice. Didn't know that exists.

src/monitoring_service/handlers.py
palango: But in this case it would then log both messages which is not what I want

tests/monitoring/monitoring_service/test_handlers.py
palango: This is just the event, the transaction itself is sent in `action_claim_reward_triggered_event_handler`

PR #319, opened 2019-04-26
Pylint: fix unused arguments

karlb:


tests/pathfinding/test_api.py
```diff
@@ -36,11 +38,9 @@ def test_get_paths_via_debug_endpoint_with_debug_disabled(
     assert response_debug.status_code == 404
 
 
+@pytest.mark.usefixtures("api_sut_with_debug")
```
palango: Mh, I'm not the biggest fan of these annotations just to make pylint happy. But adding ignores isn't much nicer either.

karlb: They do convey that this fixture is required to run the test, but it's not explicitly used inside the test. While I'm not super happy with the syntax, I think this is helpful to distinguish these from cases where a fixture is just not used, anymore.

palango: Yeah, I'm mostly converned about my nice IDE integration features not working anymore 😈

PR #323, opened 2019-04-29
change return value of iou debug endpoint to 200

Dominik1999:


palango:
This needs a fix in the tests. Also: Can you take care of the corresponding fix in raiden and are there any docs we need to update?

Dominik1999:
perfect, changed it in the tests too. There is noch change needed in the Raiden repo

PR #326, opened 2019-04-30
Use Matrix presence information

palango:
Get and store presence information from raiden nodes and use them for routing.

### DIFF
--- a/src/pathfinding_service/model/token_network.py
+++ b/src/pathfinding_service/model/token_network.py
@@ -14,12 +14,13 @@ from pathfinding_service.config import (
 from pathfinding_service.model.channel_view import ChannelView
 from raiden.messages import UpdatePFS
 from raiden.utils.typing import Address, ChannelID, TokenAmount, TokenNetworkAddress
+from raiden.network.transport.matrix import AddressReachability
 
 log = structlog.get_logger(__name__)
 
 
 class Path:
-    def __init__(self, G: DiGraph, nodes: List[Address], value: float):
+    def __init__(self, G: DiGraph, nodes: List[Address], value: TokenAmount):
         self.G = G
         self.nodes = nodes
         self.value = value
@@ -67,8 +68,8 @@ class TokenNetwork:
 
         self.address = token_network_address
         self.channel_id_to_addresses: Dict[ChannelID, Tuple[Address, Address]] = dict()
+        self.addresses_to_reachabilities: Dict[Address, AddressReachability] = dict()
         self.G = DiGraph()
-        self.max_relative_fee = 0
 
     def __repr__(self) -> str:
         return (
--- a/src/pathfinding_service/model/token_network.py
+++ b/src/pathfinding_service/model/token_network.py
@@ -13,17 +13,24 @@ from pathfinding_service.config import (
 )
 from pathfinding_service.model.channel_view import ChannelView
 from raiden.messages import UpdatePFS
-from raiden.utils.typing import Address, ChannelID, TokenAmount, TokenNetworkAddress
 from raiden.network.transport.matrix import AddressReachability
+from raiden.utils.typing import Address, ChannelID, TokenAmount, TokenNetworkAddress
 
 log = structlog.get_logger(__name__)
 
 
 class Path:
-    def __init__(self, G: DiGraph, nodes: List[Address], value: TokenAmount):
+    def __init__(
+        self,
+        G: DiGraph,
+        nodes: List[Address],
+        value: TokenAmount,
+        address_to_reachability: Dict[Address, AddressReachability],
+    ):
         self.G = G
         self.nodes = nodes
         self.value = value
+        self.address_to_reachability = address_to_reachability
 
     @property
     def edge_attrs(self) -> Iterable[dict]:
@@ -57,6 +64,13 @@ class Path:
             ratio = edge["view"].settle_timeout / edge["view"].reveal_timeout
             if ratio < DEFAULT_SETTLE_TO_REVEAL_TIMEOUT_RATIO:
                 return False
+
+        # check node reachabilities
+        for node in self.nodes:
+            node_reachability = self.address_to_reachability.get(node, AddressReachability.UNKNOWN)
+            if node_reachability != AddressReachability.REACHABLE:
+                return False
+
         return True
 
 
@@ -248,7 +262,9 @@ class TokenNetwork:
             # skip duplicates and invalid paths
             path = next(
                 p
-                for p in (Path(self.G, nodes, value) for nodes in all_paths)
+                for p in (
+                    Path(self.G, nodes, value, self.address_to_reachability) for nodes in all_paths
+                )
                 if p.is_valid and p.nodes not in disallowed_paths
             )
             return path
--- a/src/pathfinding_service/model/token_network.py
+++ b/src/pathfinding_service/model/token_network.py
@@ -68,7 +68,7 @@ class TokenNetwork:
 
         self.address = token_network_address
         self.channel_id_to_addresses: Dict[ChannelID, Tuple[Address, Address]] = dict()
-        self.addresses_to_reachabilities: Dict[Address, AddressReachability] = dict()
+        self.address_to_reachability: Dict[Address, AddressReachability] = dict()
         self.G = DiGraph()
 
     def __repr__(self) -> str:
--- a/src/pathfinding_service/service.py
+++ b/src/pathfinding_service/service.py
@@ -15,7 +15,7 @@ from pathfinding_service.model import TokenNetwork
 from raiden.constants import PATH_FINDING_BROADCASTING_ROOM, UINT256_MAX
 from raiden.messages import Message, UpdatePFS
 from raiden.network.transport.matrix import AddressReachability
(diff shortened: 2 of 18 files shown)


palango:
@ulope I just tested this again and it seems that the presence is not updated correctly.

I have the PFS and two Raiden nodes running locally, and created a channel between them. This correctly triggers the presence tracking of both addresses, and calls `refresh_address_presence`, but still the state is `AddressReachability.UNKNOWN`:
```
2019-05-02 09:11:34.912704 [info     ] Received ChannelOpened event   [pathfinding_service.service] block_number=4310582 channel_identifier=1 participant1=0x2645bcF95d1c3fE89dB435be8AaEFb887832803b participant2=0x5fBf2dd2032C2b32CBD5909401b610ba3fC8FBDE settle_timeout=500 token_network_address=0x6219a7A3A5a342B2Ae13B399BF353B6332753b45
2019-05-02 09:11:34.912930 [debug    ] Tracking address               [raiden_libs.matrix] address=0x2645bcF95d1c3fE89dB435be8AaEFb887832803b
2019-05-02 09:11:34.913198 [debug    ] Changing address presence state [raiden.network.transport.matrix.utils] address=0x2645bcf95d1c3fe89db435be8aaefb887832803b current_user=@0xbf1dabcf9eca2d89fb24be3e6e45ef422ec39ca0:transport03.raiden.network prev_state=None state=<AddressReachability.UNKNOWN: 3>
2019-05-02 09:11:34.913475 [debug    ] Updating reachability          [pathfinding_service.service] address=0x2645bcF95d1c3fE89dB435be8AaEFb887832803b status=<AddressReachability.UNKNOWN: 3>
2019-05-02 09:11:34.913650 [debug    ] Tracking address               [raiden_libs.matrix] address=0x5fBf2dd2032C2b32CBD5909401b610ba3fC8FBDE
2019-05-02 09:11:34.913891 [debug    ] Changing address presence state [raiden.network.transport.matrix.utils] address=0x5fbf2dd2032c2b32cbd5909401b610ba3fc8fbde current_user=@0xbf1dabcf9eca2d89fb24be3e6e45ef422ec39ca0:transport03.raiden.network prev_state=None state=<AddressReachability.UNKNOWN: 3>
2019-05-02 09:11:34.914154 [debug    ] Updating reachability          [pathfinding_service.service] address=0x5fBf2dd2032C2b32CBD5909401b610ba3fC8FBDE status=<AddressReachability.UNKNOWN: 3>
```

palango:
Needs to wait until https://github.com/raiden-network/raiden/pull/3930 is merged.

palango:
Tested this locally now and I get the correct presence information for the nodes.

src/pathfinding_service/model/token_network.py
```diff
@@ -56,6 +64,13 @@ def is_valid(self) -> bool:
             ratio = edge["view"].settle_timeout / edge["view"].reveal_timeout
             if ratio < DEFAULT_SETTLE_TO_REVEAL_TIMEOUT_RATIO:
                 return False
+
+        # check node reachabilities
+        for node in self.nodes:
+            node_reachability = self.address_to_reachability.get(node, AddressReachability.UNKNOWN)
+            if node_reachability != AddressReachability.REACHABLE:
+                return False
```
karlb: When we start caring about performance, we should exclude unreachable nodes from the graph instead of removing all routes including them from the results. The same is true fro the `settle_timeout / reveal_timeout >= default ratio` condition.

That's harder for the capacity check, as the exact required capacity depends on the route.

src/pathfinding_service/service.py
```diff
@@ -66,15 +69,18 @@ def __init__(  # pylint: disable=too-many-arguments
...
+
+        self.token_networks = self._load_token_networks()
         try:
-            self.matrix_listener = MatrixListener(
-                private_key=private_key,
-                chain_id=self.chain_id,
-                callback=self.handle_message,
-                service_room_suffix=PATH_FINDING_BROADCASTING_ROOM,
-            )
+            self.matrix_listener.start_client()
```
karlb: Why do we need both `matrix_listener.start_client()` and `matrix_listener.start()`?

If we move `matrix_listener.start_client()` to `_run`, can we stop mocking the MatrixListener in all tests, since it does not cause any network connections if the PFS is instantiated without calling `_run`?

src/pathfinding_service/service.py
```diff
@@ -84,6 +90,11 @@ def _load_token_networks(self) -> Dict[TokenNetworkAddress, TokenNetwork]:
         channel_views = self.database.get_channel_views()
         for cv in channel_views:
             network_for_address[cv.token_network_address].add_channel_view(cv)
+
+            # Register channel participants for presence tracking
+            self.matrix_listener.follow_address_presence(cv.participant1)
+            self.matrix_listener.follow_address_presence(cv.participant2)
```
karlb: Is this idempotent? Otherwise, we'll follow most nodes multiple times.

src/pathfinding_service/service.py
```diff
@@ -138,6 +149,15 @@ def follows_token_network(self, token_network_address: TokenNetworkAddress) -> b
...
         return token_network_address in self.token_networks.keys()
 
+    def handle_reachability_change(
+        self, address: Address, reachability: AddressReachability
+    ) -> None:
+        log.debug("Updating reachability", address=address, status=reachability)
+
+        for token_network_model in self.token_networks.values():
+            if address in token_network_model.G.nodes:
+                token_network_model.address_to_reachability[address] = reachability
```
karlb: Iterating through all nodes of all token networks is not really nice. Can't we do
```python
try:
    token_network_model.address_to_reachability[address] = reachability
except KeyError:
    pass
```
?

src/pathfinding_service/service.py
```diff
@@ -264,8 +287,14 @@ def _validate_pfs_update(self, message: UpdatePFS) -> TokenNetwork:
 
     def on_pfs_update(self, message: UpdatePFS) -> None:
         token_network = self._validate_pfs_update(message)
+
         log.info("Received Capacity Update", **message.to_dict())
         self.database.upsert_capacity_update(message)
+
+        # Follow presence for the channel participants
+        self.matrix_listener.follow_address_presence(message.updating_participant, refresh=True)
+        self.matrix_listener.follow_address_presence(message.other_participant, refresh=True)
```
karlb: Why do we have `refresh=True` here, but not in `_load_token_networks`?

palango: `_load_token_networks` is run before the matrix client is started, so just register interest in the address and this is taken care of when the client starts.

When we receive a new channel we have to register interest but also query the current state, which is what `refresh=True` does.

src/pathfinding_service/service.py
palango: Wouldn't the above code assign the reachability for any case and never throw a `KeyError`?

src/pathfinding_service/service.py
palango: Yes:
```py
   def add_address(self, address: Address):
        """ Add ``address`` to the known addresses that are being observed for reachability. """
        # Since _address_to_userids is a defaultdict accessing the key creates the entry
        _ = self._address_to_userids[address]
```

src/pathfinding_service/model/token_network.py
palango: Sure, but then we have to keep a general graph with all edges and a pruned one which is used for routing.

src/pathfinding_service/service.py
karlb: Right, make that
```
if address in token_network_model.address_to_reachability:
    token_network_model.address_to_reachability[address]
```
and set it to unknown when the node is first seen.

tests/pathfinding/test_graphs.py
```diff
@@ -185,3 +186,75 @@ def get_paths(diversity_penalty):
...
+
+
+@pytest.mark.usefixtures("populate_token_network_case_3")
+def test_reachability_mediator(token_network_model: TokenNetwork, addresses: List[Address]):
+    def get_paths():
+        paths = token_network_model.get_paths(
+            addresses[0], addresses[8], value=TokenAmount(10), max_paths=5
+        )
+        index_paths = [addresses_to_indexes(p["path"], addresses) for p in paths]
+        return index_paths
```
karlb: `get_paths` is repeated a few times. We should pull that out into a separate fixture.

src/pathfinding_service/service.py
palango: Then we iterate over the keys of `address_to_reachability` instead of `G.nodes`. Does that improve things?

I'd say the other way is to store presence in the service, not in the token network and use that global map during routing.

karlb: One is a dict, the other is a list, so yes. But we can leave this for later. We'll probably move things around a few times, anyway.

tests/pathfinding/test_graphs.py
palango: Ref https://github.com/raiden-network/raiden-services/issues/349

src/pathfinding_service/service.py
palango: Good point, fixed!

Issue #325, opened 2019-04-30
Use byte encoded addresses


palango:
In contrast to Raiden we use hex-encoded addresses. This complicated interoperability and is source of confusion.

We should switch to use byte-encoded addresses internally.

karlb:
There's also the HexBytes type which behaves like bytes but provides a hex representation when printed. Do we want to use this or not?

palango:
> There's also the HexBytes type which behaves like bytes but provides a hex representation when printed. 

With the advantage being that we wouldn't need to encode all byte-addresses, e.g. for logging?!

But I guess we would need to create the objects ourselves, so wouldn't this lead to the same inconsistencies like we have now?

karlb:
We should probably talk about this with someone from the client team to make sure that we're going into the right direction. See https://github.com/raiden-network/raiden/issues/3051.

@andrevmatos: any comments on this?

andrevmatos:
The decision to use byte-encoded was made long ago on Raiden client, and checksummed-hex-encoded on all REST/HTTP interfaces and JSON communcation (like Matrix), so I'd say services would do good in keeping this pattern as well. `web3.py` initially supported only hex-encoded strings, but some time ago also added full support to bytes addresses everywhere.
The `HexBytes` is just some logging sugar, while my proposal in 3051 Karl pointed above has some more additional fancyness, while keeping full inheritance with `bytes`, so I think it'd be nice if anyone is willing to finish or import the work from there, it could be useful while being transparently handed over to Raiden and libs.

PR #333, opened 2019-05-03
Byte typed addresses

karlb:
This will need some cleanups, but those should be easier to do now that the main change is done. And I don't want to wait with the merge since this will quickly attract merge conflicts during rebases.

Closes https://github.com/raiden-network/raiden-services/issues/325

### DIFF
--- a/src/monitoring_service/database.py
+++ b/src/monitoring_service/database.py
@@ -3,7 +3,7 @@ import sqlite3
 from typing import List, Optional, Union, cast
 
 import structlog
-from eth_utils import is_checksum_address
+from eth_utils import decode_hex, is_checksum_address, to_checksum_address
 
 from monitoring_service.events import (
     ActionClaimRewardTriggeredEvent,
@@ -16,7 +16,7 @@ from monitoring_service.states import (
     MonitorRequest,
     OnChainUpdateStatus,
 )
-from raiden.utils.typing import BlockNumber
+from raiden.utils.typing import BlockNumber, TokenNetworkAddress
 from raiden_libs.database import BaseDatabase
 from raiden_libs.types import Address
 
@@ -35,7 +35,7 @@ class SharedDatabase(BaseDatabase):
     def upsert_monitor_request(self, request: MonitorRequest) -> None:
         values = [
             hex(request.channel_identifier),
-            request.token_network_address,
+            to_checksum_address(request.token_network_address),
             request.balance_hash,
             hex(request.nonce),
             request.additional_hash,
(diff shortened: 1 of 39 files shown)


src/monitoring_service/database.py
```diff
@@ -102,23 +106,29 @@ def upsert_channel(self, channel: Channel) -> None:
...
-            [hex(channel_id), token_network_address],
+            [hex(channel_id), to_checksum_address(token_network_address)],
         ).fetchone()
 
         if row is None:
             return None
         kwargs = {
             key: val for key, val in zip(row.keys(), row) if not key.startswith("update_status")
         }
+        kwargs["token_network_address"] = decode_hex(kwargs["token_network_address"])
```
palango: Mh, I'd perfer to use `to_canonical_address` for hex to bytes conversion. I think it makes the intention clearer.

src/pathfinding_service/api.py
```diff
@@ -42,11 +41,15 @@
 
 class ApiWithErrorHandler(Api):
     def handle_error(self, e: exceptions.ApiException) -> Response:
-        log.debug("Error while handling request", error=e, details=e.error_details, message=e.msg)
-        return self.make_response(
-            {"errors": e.msg, "error_code": e.error_code, "error_details": e.error_details},
-            e.http_code,
-        )
+        if isinstance(e, exceptions.ApiException):
```
palango: Why do we need this check? Following the annotation this should always be true.

src/pathfinding_service/api.py
```diff
@@ -263,24 +266,23 @@ def get(self) -> Tuple[dict, int]:
...
         )
 
 
 class DebugEndpoint(PathfinderResource):
     def get(  # pylint: disable=no-self-use
-        self,
-        token_network_address: TokenNetworkAddress,
-        source_address: Address,
-        target_address: Address = None,
+        self, token_network_address: str, source_address: Address, target_address: Address = None
```
palango: `source_address` and `target_address` should be strings as well.

src/pathfinding_service/api.py
karlb: The annotation should be more generic. We might get any kind of exception in the error handler, AFAICS.

palango:
Another point, should we switch the DB representation of addresses to bytes as well? It would make stuff more consistent, but would be harder to query by hand.

src/monitoring_service/database.py
karlb: But that returns eth_typing.Address, which we are not using, so I expect that we'll get type errors. Do you agree?

And why don't we use eth_typing.Address as address type?

karlb:
> Another point, should we switch the DB representation of addresses to bytes as well? It would make stuff more consistent, but would be harder to query by hand.

I thought about this, too. It would work fine from an sqlite point of view, but I'm still torn due to the readbility and also writeability when manually dealing with the db.

src/monitoring_service/database.py
palango: In a quick check mypy didn't complain, which is a bit surprising.

And I think the `eth_typing` package is quite new, so we might want to use the Raiden types as aliases.

palango: Anyways, we can fix the `decode_hex` stuff later.

palango:
> readbility and also writeability when manually dealing with the db.

Yeah, let's keep it readable for now. The DB code shouldn't change that much anyways.

PR #335, opened 2019-05-03
Adds minimum calculation for capacity updates

### LINKED ISSUE #290: Use `min` to calculate capacity in honest-malicious scenario
As explained in #248 the PFS needs to use `min(C_A, C_B)` when it receives mismatching updates from `A` and `B`.

Dominik1999:
closes #290

### DIFF
--- a/src/pathfinding_service/model/channel_view.py
+++ b/src/pathfinding_service/model/channel_view.py
@@ -42,25 +42,22 @@ class ChannelView:
         if self.capacity is None:
             self.capacity = self.deposit
 
-    # TODO: define another function update_deposit
+    def update_deposit(self, total_deposit: TokenAmount) -> None:
+        if total_deposit > self.deposit:
+            self.capacity = TokenAmount(self.capacity + total_deposit - self.deposit)
+            self.deposit = TokenAmount(total_deposit)
+
     def update_capacity(
         self,
+        capacity: TokenAmount,
         nonce: Nonce = Nonce(0),
-        capacity: TokenAmount = TokenAmount(0),
         reveal_timeout: int = None,
-        deposit: TokenAmount = None,
         mediation_fee: FeeAmount = FeeAmount(0),
     ) -> None:
         self.update_nonce = nonce
         self.capacity = capacity
         if reveal_timeout is not None:
             self.reveal_timeout = reveal_timeout
-        # FIXME: think about edge cases
-        if deposit is not None:
-            self.deposit = deposit
-            if self.capacity is not None:
-                self.capacity = TokenAmount(self.capacity + deposit)
-
         self.absolute_fee = mediation_fee
 
     def fee(self, amount: TokenAmount) -> int:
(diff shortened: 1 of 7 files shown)


src/pathfinding_service/model/channel_view.py
```diff
@@ -42,25 +42,22 @@ def __post_init__(self) -> None:
         if self.capacity is None:
             self.capacity = self.deposit
 
-    # TODO: define another function update_deposit
+    def update_deposit(self, deposit: TokenAmount = TokenAmount(0)) -> None:
```
palango: I'd not make this a optional parameter, and call it `total_deposit` as well.

src/pathfinding_service/model/channel_view.py
```diff
@@ -42,25 +42,22 @@ def __post_init__(self) -> None:
...
+
+        self.capacity = TokenAmount(self.capacity + deposit - self.deposit)
```
palango: I'd also check that `deposit` is bigger than `self.deposit`.

src/pathfinding_service/model/channel_view.py
```diff
@@ -42,25 +42,22 @@ def __post_init__(self) -> None:
...
+        self.deposit = TokenAmount(self.deposit + deposit)
```
palango: This is wrong, `deposit` is a total value.

src/pathfinding_service/model/channel_view.py
```diff
@@ -42,25 +42,22 @@ def __post_init__(self) -> None:
...
+
     def update_capacity(
```
palango: What happens if the PFS receives a update before it can get the `Deposit` event? I think we don't handle this correctly.

src/pathfinding_service/service.py
```diff
@@ -260,22 +260,19 @@ def _validate_pfs_update(self, message: UpdatePFS) -> TokenNetwork:
...
+        self.database.upsert_capacity_update(message)
+        reported_capacities_of_partner = self.database.get_capacity_updates(
+            updating_participant=message.other_participant,
+            token_network_address=TokenNetworkAddress(
+                message.canonical_identifier.token_network_address
+            ),
+            channel_id=message.canonical_identifier.channel_identifier,
+        )
+        token_network.handle_channel_balance_update_message(
+            message, reported_capacities_of_partner
```
palango: Please add keywords to the parameters.

src/pathfinding_service/model/channel_view.py
Dominik1999: ok I give you that

src/pathfinding_service/model/channel_view.py
Dominik1999: what do we do if this is not the case?

palango: Then it wouldn't make sense to call the function

Issue #336, opened 2019-05-03
Routing feedback


palango:
## Description

**As Josh I want to be able to**

give feedback to the PFS

**So that I can**

get better routes in the future.

## Acceptance Criteria

- [x] Client gives feedback about (un)sucessful routes (https://github.com/raiden-network/raiden/issues/3996)
- [x] PFS received feedback (from client that requested routes before)
- [x] PFS stores feedback for further usage
- [ ] Update spec

palango:
This feedback is not broadcasted, but sent to a specific PFS with a token received when requesting the route.

- `GET` /paths: returns a token, valid for x time, format???
- `POST` /feedback: new endpoint

    JSON body:

    ```json
    {
        "token": "a6c340e8-73cb-495e-9ce0-52e460411299",
        "status": "success",  # or: "failure"
        "path": ["0xalice", "0xcharlie", "0xbob"],
    }
    ```

PR #342, opened 2019-05-06
Add FeedbackToken

palango:
This adds the type FeedbackToken which will be used for the feedback
mechanism. It also adds code to save and load it from the database.

Part of #336

### DIFF
--- a/src/pathfinding_service/database.py
+++ b/src/pathfinding_service/database.py
@@ -1,10 +1,11 @@
 import os
 from typing import Iterator, Optional, Tuple
+from uuid import UUID
 
 import structlog
 from eth_utils import decode_hex, to_checksum_address
 
-from pathfinding_service.model import IOU
+from pathfinding_service.model import IOU, FeedbackToken
 from pathfinding_service.model.channel_view import ChannelView
 from pathfinding_service.model.token_network import TokenNetwork
 from raiden.messages import UpdatePFS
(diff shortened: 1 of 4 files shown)


src/pathfinding_service/database.py
```diff
@@ -203,3 +204,27 @@ def delete_channel_views(self, channel_id: ChannelID) -> None:
     def get_token_networks(self) -> Iterator[TokenNetwork]:
         for row in self.conn.execute("SELECT address FROM token_network"):
             yield TokenNetwork(token_network_address=decode_hex(row[0]))
+
+    def insert_feedback_token(self, token: FeedbackToken) -> None:
+        token_dict = dict(token_id=token.id.hex, expiry=token.expiry)
+        self.conn.execute(
+            """
+            INSERT INTO feedback_token (
+                token_id, expiry
```
karlb: I would rather store the time at which the token is generated. This seems like the more meaningful time to me. Should work either way, though.

src/pathfinding_service/schema.sql
```diff
@@ -46,5 +46,11 @@ CREATE TABLE capacity_update (
     PRIMARY KEY (updating_participant, token_network_address, channel_id)
 );
 
+CREATE TABLE feedback_token (
+    token_id CHAR(32) NOT NULL,
+    expiry TIMESTAMP NOT NULL,
+    PRIMARY KEY (token_id)
+);
+
 CREATE UNIQUE INDEX one_active_session_per_sender
```
karlb: This constraint belongs to the `IOU` table, so better keep it right below it.

tests/pathfinding/test_database.py
```diff
@@ -0,0 +1,15 @@
+from datetime import datetime
+from uuid import uuid4
+
+from pathfinding_service.model import FeedbackToken
+
+
+def test_save_and_load_feedback_token(pathfinding_service_mock):
+    token = FeedbackToken(id=uuid4(), expiry=datetime.now())
```
karlb: Using `utcnow` usually leads to less problems down the road.

palango:
All comments made a lot of sense and are fixed now.

PR #343, opened 2019-05-07
Add feedback API

palango:
Part of #336 

This adds the feedback API to the PFS, I'd like to wait with merging this until I have the client changes done and tested.

Basically this PR adds a new ` /<token_network_address>/feedback` endpoint, which can be used to send feedback about routes to the PFS.

Some points for discussion:
- I don't store the sender of the feedback, as we can't be sure the person requesting a route is the actual source in the request. We'd need a signature in the request for that.
- I decided to return HTTP 200, even if the feedback was not accepted, the reasons for this are that the client's doesn't need to know and
- The time interval for feedback is currently set to 10mins
- Feedback is not yet persisted to the DB

### DIFF
--- a/src/pathfinding_service/api.py
+++ b/src/pathfinding_service/api.py
@@ -28,7 +28,7 @@ from pathfinding_service.config import (
     UDC_SECURITY_MARGIN_FACTOR,
 )
 from pathfinding_service.model import IOU
-from pathfinding_service.model.feedback import FeedbackToken, RouteFeedback
+from pathfinding_service.model.feedback import FeedbackToken
 from pathfinding_service.model.token_network import TokenNetwork
 from pathfinding_service.service import PathfindingService
 from raiden.exceptions import InvalidSignature
@@ -156,19 +156,24 @@ class PathsResource(PathfinderResource):
             )
 
         # Create a feedback token and store it to the DB
-        feedback_token = create_and_store_feedback_token(
+        feedback_token = create_and_store_feedback_tokens(
             pathfinding_service=self.pathfinding_service,
             token_network_address=token_network.address,
+            routes=paths,
         )
 
         return {"result": paths, "feedback_token": feedback_token.id.hex}, 200
 
 
-def create_and_store_feedback_token(
-    pathfinding_service: PathfindingService, token_network_address: TokenNetworkAddress
+def create_and_store_feedback_tokens(
+    pathfinding_service: PathfindingService,
+    token_network_address: TokenNetworkAddress,
+    routes: List[Dict],
 ) -> FeedbackToken:
     feedback_token = FeedbackToken(token_network_address=token_network_address)
-    pathfinding_service.database.insert_feedback_token(feedback_token)
+
+    for route in routes:
+        pathfinding_service.database.prepare_feedback(token=feedback_token, route=route["path"])
 
     return feedback_token
 
@@ -297,7 +302,9 @@ class FeedbackResource(PathfinderResource):
         token_network = self._validate_token_network_argument(token_network_address)
         feedback_request = self._parse_post(FeedbackRequest)
         feedback_token = self.pathfinding_service.database.get_feedback_token(
-            token_id=feedback_request.token, token_network_address=token_network.address
+            token_id=feedback_request.token,
+            token_network_address=token_network.address,
+            route=feedback_request.path,
         )
 
         # The client doesn't need to know whether the feedback was accepted or not,
@@ -306,13 +313,16 @@ class FeedbackResource(PathfinderResource):
             return {}, 400
 
         log.info("Received feedback", feedback=feedback_request)
-        token_network.feedback.append(
-            RouteFeedback(
-                status=feedback_request.status,
-                received_time=datetime.utcnow(),
-                path=feedback_request.path,
-            )
+
+        has_feedback_for_request = self.pathfinding_service.database.has_feedback_for(
+            token=feedback_token, route=feedback_request.path
         )
+
+        if not has_feedback_for_request:
+            success = feedback_request.status == "successful"
+            self.pathfinding_service.database.update_feedback(
+                token=feedback_token, route=feedback_request.path, successful=success
+            )
 
         return {}, 200
 
--- a/src/pathfinding_service/api.py
+++ b/src/pathfinding_service/api.py
@@ -161,7 +161,7 @@ class PathsResource(PathfinderResource):
             token_network_address=token_network.address,
         )
 
-        return {"result": {"paths": paths, "feedback_token": feedback_token.id.hex}}, 200
+        return {"result": paths, "feedback_token": feedback_token.id.hex}, 200
 
 
 def create_and_store_feedback_token(
@@ -301,11 +301,9 @@ class FeedbackResource(PathfinderResource):
         )
 
         # The client doesn't need to know whether the feedback was accepted or not,
-        # so in case the token is invalid we still return HTTP 200
-        if not feedback_token:
-            return {}, 200
-        if not feedback_token.is_valid():
-            return {}, 200
+        # so in case the token is invalid we return HTTP 400 without further infos
+        if not feedback_token or not feedback_token.is_valid():
+            return {}, 400
 
         log.info("Received feedback", feedback=feedback_request)
         token_network.feedback.append(
--- a/src/pathfinding_service/api.py
+++ b/src/pathfinding_service/api.py
@@ -2,7 +2,7 @@ import collections
 from dataclasses import dataclass, field
 from datetime import datetime
 from typing import Any, ClassVar, Dict, List, Optional, Tuple, Type, TypeVar, cast
-from uuid import UUID, uuid4
+from uuid import UUID
 
 import marshmallow
 import pkg_resources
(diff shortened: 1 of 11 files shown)


src/pathfinding_service/api.py
```diff
@@ -152,7 +155,22 @@ def post(self, token_network_address: str) -> Tuple[dict, int]:
...
             )
 
-        return {"result": paths}, 200
+        # Create a feedback token and store it to the DB
+        feedback_token = create_and_store_feedback_token(
+            pathfinding_service=self.pathfinding_service,
+            token_network_address=token_network.address,
+        )
+
+        return {"result": {"paths": paths, "feedback_token": feedback_token.id.hex}}, 200
```
palango: We could undo this change and merge the PR, and apply this change as soon as the client changes are ready.

karlb: If we have one token for multiple paths, we'll have to be able to store different results for different token+path combinations. That's ok, it was just something I didn't think about before.

karlb: We could also keep all keys on the top level and have `{"result": ..., "token": ...}`. Then we would save one level of nesting and be able to merge it without breaking compatibility.

src/pathfinding_service/api.py
```diff
@@ -253,6 +271,54 @@ def get(
...
+        feedback_request = self._parse_post(FeedbackRequest)
+        feedback_token = self.pathfinding_service.database.get_feedback_token(
+            token_id=feedback_request.token, token_network_address=token_network.address
+        )
+
+        # The client doesn't need to know whether the feedback was accepted or not,
+        # so in case the token is invalid we still return HTTP 200
+        if not feedback_token:
+            return {}, 200
+        if not feedback_token.is_valid():
```
karlb: Since both of these will do the same thing, you could fold them into a single if `if not feedback_token or not feedback_token.is_valid()`

src/pathfinding_service/api.py
```diff
@@ -253,6 +271,54 @@ def get(
...
+            return {}, 200
+
+        log.info("Received feedback", feedback=feedback_request)
+        token_network.feedback.append(
+            RouteFeedback(
+                status=feedback_request.status,
+                received_time=datetime.utcnow(),
+                path=feedback_request.path,
+            )
```
karlb: We should take care that only one response per token+route will be stored. But we can do that when we add persistence.

src/pathfinding_service/model/feedback.py
```diff
@@ -0,0 +1,25 @@
...
+    def is_valid(self) -> bool:
+        """ Checks if the token is valid."""
+        return self.creation_time + MAX_AGE_OF_FEEDBACK_REQUESTS > datetime.utcnow()
+
+
+@dataclass
+class RouteFeedback:
+    status: str
+    received_time: datetime
+    path: List[Address]
```
karlb: When we want to deduplicate token+path combinations, we'll have to add the token, here.

src/pathfinding_service/schema.sql
```diff
@@ -52,5 +52,6 @@ CREATE TABLE capacity_update (
 CREATE TABLE feedback_token (
     token_id CHAR(32) NOT NULL,
     creation_time TIMESTAMP NOT NULL,
+    token_network_address CHAR(42) NOT NULL,
```
karlb: We won't be able to make sure that the client only provides feedback for the routes we've sent to him unless we save those routes. But we can do that at a later time.

tests/pathfinding/test_api.py
```diff
@@ -328,3 +324,58 @@ def make_params(timestamp: datetime):
...
+    assert response.json()["error_code"] == exceptions.InvalidRequest.error_code
+
+    # Test valid IOU, but not in PFS DB
+    token = FeedbackToken(token_network_address=token_network_model.address)
+
+    response = make_request(token_id=token.id.hex)
+    assert response.status_code == 200
+    assert len(token_network_model.feedback) == 0
+
+    # Test old IOU
```
karlb: IOU?

tests/pathfinding/test_api.py
```diff
@@ -328,3 +324,58 @@ def make_params(timestamp: datetime):
...
+    response = make_request(status="abc")
+    assert response.status_code == 400
+    assert response.json()["error_code"] == exceptions.InvalidRequest.error_code
+
+    # Request with invalid path
+    response = make_request(path="abc")  # type: ignore
+    assert response.status_code == 400
+    assert response.json()["error_code"] == exceptions.InvalidRequest.error_code
+
+    # Test valid IOU, but not in PFS DB
```
karlb: IOU?

palango:
Fixed the easy review comments.

For the other ones, this PR (or a follow-up) should also:
- Create one token for each returned path
- Store the path with the token in the DB
- When feedback is received, store the feedback with a reference to the original token

@karlb Anything to add?

karlb:
> For the other ones, this PR (or a follow-up) should also:
>     * Create one token for each returned path

This is one solution, but not necessary. The other way would be use token+route as primary key. You could even combine the token table with the result table and have the following columns

* token
* token_created_at
* route
* successful (true, false or null)
* transfer_duration
PRIMARY KEY (token, route)

The first three columns would be set during insertion when the routes are sent to the client, the last two will be updated when the response is received. This fits nicely with the requirements:
* When no entry for this token+route combination is present, you can't update the row to include the results. Which is what we want
* If the results are already set, you'll overwrite them instead of adding a second feedback entry. Which is also what we want.

src/pathfinding_service/api.py
```diff
@@ -297,7 +302,9 @@ def post(
         token_network = self._validate_token_network_argument(token_network_address)
         feedback_request = self._parse_post(FeedbackRequest)
         feedback_token = self.pathfinding_service.database.get_feedback_token(
```
karlb: Instead of doing
* select
* check if feedback is present
* update

you could do all in one step:
```
UPDATE feedback
SET successful = ?, feedback_time = ? 
WHERE token_id = ? AND route = ? AND successful IS NULL
```
The results `rowcount` attribute will tell you how many rows have been changed.

src/pathfinding_service/database.py
```diff
@@ -206,31 +208,84 @@ def get_token_networks(self) -> Iterator[TokenNetwork]:
...
+                token_id, creation_time, token_network_address, route
             ) VALUES (
                 :token_id,
                 :creation_time,
-                :token_network_address
+                :token_network_address,
+                :route
             )
         """,
             token_dict,
```
karlb: Not for this PR, but it's getting time that we create an insert helper like
```python
database.insert('feedback', token_dict)
```

src/pathfinding_service/database.py
```diff
@@ -206,31 +208,84 @@ def get_token_networks(self) -> Iterator[TokenNetwork]:
...
+            """,
+            [
+                token.id.hex,
+                to_checksum_address(token.token_network_address),
+                json.dumps(hexed_route),
+            ],
+        ).fetchone()
+
+        if feedback:
+            return feedback["successful"] is not None
```
karlb: You can also do `SELECT successful IS NOT NULL` right in the query. To me that is more straighforward, but I'm also used to have logic in SQL.

src/pathfinding_service/schema.sql
```diff
@@ -49,9 +49,12 @@ CREATE TABLE capacity_update (
...
 
-CREATE TABLE feedback_token (
+CREATE TABLE feedback (
     token_id CHAR(32) NOT NULL,
     creation_time TIMESTAMP NOT NULL,
     token_network_address CHAR(42) NOT NULL,
-    PRIMARY KEY (token_id)
+    route TEXT NOT NULL,
+    successful BOOLEAN CHECK (successful IN (0,1)),
+    feedback_time TIMESTAMP,
```
karlb: Is this the time at which we received the feedback or the time it took to do the transfer? It sounds like the former, but IMO the latter would be more useful. Or do we assume `feedback_time - creation_time = transfer_duration`?

src/pathfinding_service/api.py
```diff
@@ -156,19 +156,24 @@ def post(self, token_network_address: str) -> Tuple[dict, int]:
...
+def create_and_store_feedback_tokens(
+    pathfinding_service: PathfindingService,
+    token_network_address: TokenNetworkAddress,
+    routes: List[Dict],
 ) -> FeedbackToken:
     feedback_token = FeedbackToken(token_network_address=token_network_address)
-    pathfinding_service.database.insert_feedback_token(feedback_token)
+
+    for route in routes:
+        pathfinding_service.database.prepare_feedback(token=feedback_token, route=route["path"])
```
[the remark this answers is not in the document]
palango: added a TODO

src/pathfinding_service/schema.sql
palango: This is the time the feedback is received, my thinking was that it's better to save the raw information, the duration is then easy to calculate from them.

src/pathfinding_service/api.py
palango: That's a nice improvement!

src/pathfinding_service/schema.sql
karlb: If we can easily calculate the duration, I agree. But it might get complicated if the first request does not succeed. The client might even wait for locks to expire before trying another route. But maybe I'm too pessimistic here and deriving the data will be fine.

palango: So you say we should take the latest feedback for a given token as the start, and the current time as the end?

karlb: I think this is the only way to get timings out of this data. This will only work if the client tries each route immediately after the previous one, submits feedback immediately and the latency to the PFS is small compared to the duration of a payment.

That's why I thought making the client submit the duration explicitly would make data more robust and easier to handle.

PR #347, opened 2019-05-08
Fix public_key_to_address typing

### LINKED ISSUE #346: Fix public_key_to_address typing
## Problem Definition

The return type `Address` and `to_checksum_address` don't match.
```py
def public_key_to_address(public_key: Union[PublicKey, bytes]) -> Address:
    """ Converts a public key to an Ethereum address. """
    if isinstance(public_key, PublicKey):
        public_key = public_key.format(compressed=False)
    assert isinstance(public_key, bytes)
    return to_checksum_address(keccak(public_key[1:])[-20:])
```

Dominik1999:
Closes #346

### DIFF
--- a/src/monitoring_service/service.py
+++ b/src/monitoring_service/service.py
@@ -3,7 +3,6 @@ import time
 from typing import Callable, Dict
 
 import structlog
-from eth_utils import decode_hex
 from web3 import Web3
 from web3.contract import Contract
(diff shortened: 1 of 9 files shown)


karlb:
We have different definitions for `public_key_to_address` in `raiden-services` and `raiden-contracts`, now. If anyone stumbles across a good opportunity to unify those two, please think of this!

Issue #341, opened 2019-05-06
Spec out Mediation Fees communication between client and PFS


karlb:
- [x] Read up on previous mediation fees discussions:
  - [x] https://github.com/raiden-network/spec/issues/10
  - [x] https://github.com/raiden-network/team/issues/225
- [x] Talk to trustlines about their imbalance fees
- [x] Discuss with client team
- [x] Write ADR

karlb:
Results for older discussions:
* Mediation fees are paid in the token that is being transferred
* Mediation fees can not be set globally. They have to be set either per token network or per channel.

General points regarding general mediation fees:
* Mediation fees will not be enforced for now. If we add onion routing, that can and should be used to enforce mediation fees.

General points regarding the imbalance fees:
* Imbalance fees don't work for nodes which are only sources of tokens or sinks. For those, `deposit` and `withdraw` will have to be used to regain capacity.
* Imbalance fees should depend on the transferred amount. Otherwise you can't even tell if the channel is getting more or less balanced.
* It depends on the node which state is desired (and thus considered "balanced"). Clients which mostly use their channel to pay will want as much free capacity as possible, while mediating nodes might want the capacity evenly split between both parties to facilitate mediations in both directions.
* Imbalance fees should grow non-linearly with the transferred amount to disincentivize fully exhausting a channel.
* For the network balance, it's not only relevant over which channel a mediating node sends a payment but also over which channel the payment is received. For optimal balancing results, it is necessary to include this information in the overall mediation fee.

karlb:
To evaluate different approaches, let's go through some basic examples and rate which transfers are desirable for a node. Then we can evaluate which imbalance fees can be used to properly express the node's intention. Plus signs show desired transfers, minus signs those which are to be avoided.

Example 1:

A mediating node has three channels. The channel to C is nearly exhausted, while the other channels have a free capacity of 9 tokens left (of 10 tokens which are deposited in that channel). This node wants to have 
a similar amount of free capacity for all channel, so that it has more mediation opportunities. To achieve this, it must receive tokens over channel C.

| from\to | A | B | C |
| --- | --- | --- | --- |
| A 1/10 | | - | -- |
| B 1/10 | - | | -- |
| C 9/10 | ++ | ++ | |

Example 2:

A customer uses channel A only to pay and therefore would like to mediate in the reverse direction.

| from\to | A | B |
| --- | --- | --- |
| A 5/10 | | -- |
| B 5/10 | ++ | |

[...]

PR #348, opened 2019-05-09
Add mediation fees ADR

karlb:
I've changed the proportional fee to a float value again, since we won't be 100% accurate anyway due to the interpolation of the imbalance penalty function.

Did I forget anything? Do any points need more clarification? Did I assume consensus on something we didn't agree on?

Closes https://github.com/raiden-network/raiden-services/issues/341.

### DIFF
--- a/adr/003-mediation-fees.md
+++ b/adr/003-mediation-fees.md
@@ -1,0 +1,146 @@
+# Mediation Fees calculation and message format
+
+* Status: proposed
+* Deciders: Augusto, Dominik, Karl, Paul
+* Date: 2019-05-07
+
+## Context and Problem Statement
+
+Mediating nodes demand fees for their mediation services. These fees should not only incentivize running a Raiden node, but also help balancing the network. This can be done by reducing the mediation fees for payments which leave channels in a more balanced state than before the payment (possibly even to a negative fee).
+
+## Decision Drivers
+
+* Balancing the network should be incentivized
+* Raiden nodes will act selfishly
+* Mediation fees should be flexible enough to accommodate arbitrary business models for parties operation Raiden nodes.
+
+## Considered Options
+
+1. A fixed fee + a proportional fee (flat + prop * amount)
+2. A fixed fee + a proportional fee + an imbalance fee based on a fixed imbalance measure
+3. A fixed fee + a proportional fee + submitting points of a function describing the imbalance fee
+4. Submitting points of a single function (dependent on the payment amount) which includes all fees
+
+## Decision Outcome
+
+Chosen option: **Option 3**, but with some important additions (see below).
+
+Option 1 would be sufficient for the short very term, but does not incentivize rebalancing. Option 2 does not work because no objective balance measure could be found that is valid across all business models / user interests. Option 4 would work but require sending a changed fee schedule after every transfer or mediation. Since this information is broadcast, that's a non-negligible amount of traffic.
+
+### Positive Consequences
+
+* Adaptable to different business models
+* Both in- and outgoing channels can be rebalanced
+* Fully exhausting a channel can be punished more strongly than just moving into the wrong direction
+* Fee schedules only have to be changed when the node's strategy changes or the on-chain capacity changes via `deposit` or `withdraw`
+
+### Negative Consequences
+
+* Comparatively complex
+
+## Detailed analysis
+
+One important detail is that fees are not only calculated for the channel over which the mediation payment is sent, but also for the channel over which the payment is received. Without this, a node could not incentivize receiving through a certain channel, which is just as important for rebalancing as sending through the right channel. So the total mediation fee for a node is
+
+`MF_total = MF_in_channel + MF_out_channel`
+
+where for each channel
+
+`MF_channel = flat_fee + proportional_fee * amount + IP(C_before) - IP(C_after)`
+
+where `IP` is the "Imbalance Penalty" function and `C_before/after` are the channel capacities before and after the payment. The channel capacity is the node's free capacity as reported to the PFS with the UpdatePFS messages.
+
+The IP function is a function that describes how much a node is willing to pay to move away from an undesirable channel capacity. If a node prefers to have a channel capacity of 5 while the total capacity of that channel is 10 (so that it could mediate up to 5 tokens in both directions) the IP function might look like
+
+```
+IP
+^
+|X                      X
+|X                      X
+| X                    X
+| X                    X
+|  X                  X
+|   X                X|
+|    X              X |
+|     X            X  |dIP = IP(C_before) - IP(C_after)
+|      XX        XX   |
+|        XX    XX----->
+|          XXXX  amount
++---------------+-----+--> Capacity
+                6     9
+```
+
+If the node currently has a capacity of 6 and is asked to mediate a payment of 3 tokens coming from this channel, it will get into the less desired position of 9 capacity. To compensate for this, it will demand an imbalance fee of `dIP = IP(C_before) - IP(C_after)`. If the situation was reversed and the capacity would go from 9 to 6, the absolute value would be the same, but this time it would be negative and thus incentivize moving towards the preferred state. By viewing the channel balances in this way, the imbalance fee is a zero sum game in the long term. All tokens which are earned by going into a bad state will be spent for moving into a good state again, later.
+
+Only mediating nodes demand mediation fees. The initiator and target could theoretically benefit if their IP was taken into account during routing, but this aspect is left out for now to avoid additional complexity.
+
+### Economic Aspects
+
+In this model, the mediation fees are split into three parts:
+
+1. The *flat fee*. This is meant to represent the base cost of handling a payment, so that even small payment amounts can be profitable to mediate.
+2. The *proportional fee*. Large payments lock up tokens while the payment is in progress. To compensate this, the operator should receive larger fees for large payments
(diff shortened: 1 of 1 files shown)


adr/003-mediation-fees.md
```diff
@@ -0,0 +1,146 @@
...
+1. A fixed fee + a proportional fee (flat + prop * amount)
+2. A fixed fee + a proportional fee + an imbalance fee based on a fixed imbalance measure
+3. A fixed fee + a proportional fee + submitting points of a function describing the imbalance fee
+4. Submitting points of a single function (dependent on the payment amount) which includes all fees
+
+## Decision Outcome
+
+Chosen option: **Option 3**, but with some important additions (see below).
+
+Option 1 would be sufficient for the short very term, but does not incentivize rebalancing. Option 2 does not work because no objective balance measure could be found that is valid across all business models / user interests. Option 4 would work but require sending a changed fee schedule after every transfer or mediation. Since this information is broadcast, that's a non-negligible amount of traffic.
```
palango: Maybe make this a list for each option?

adr/003-mediation-fees.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Positive Consequences
+
```
palango: I'd add that balance fees can also be disabled.

adr/003-mediation-fees.md
```diff
@@ -0,0 +1,146 @@
...
+* Fully exhausting a channel can be punished more strongly than just moving into the wrong direction
+* Fee schedules only have to be changed when the node's strategy changes or the on-chain capacity changes via `deposit` or `withdraw`
+
+### Negative Consequences
+
+* Comparatively complex
+
+## Detailed analysis
+
+One important detail is that fees are not only calculated for the channel over which the mediation payment is sent, but also for the channel over which the payment is received. Without this, a node could not incentivize receiving through a certain channel, which is just as important for rebalancing as sending through the right channel. So the total mediation fee for a node is
```
palango: Maybe the example we had at the whiteboard would help to understand why this is desirable.

adr/003-mediation-fees.md
```diff
@@ -0,0 +1,146 @@
...
+
+`MF_total = MF_in_channel + MF_out_channel`
+
+where for each channel
+
+`MF_channel = flat_fee + proportional_fee * amount + IP(C_before) - IP(C_after)`
```
palango: I thought we said that flat and proportional fees are only applicable for the outgoing channel.

adr/003-mediation-fees.md
```diff
@@ -0,0 +1,146 @@
...
+
+where `IP` is the "Imbalance Penalty" function and `C_before/after` are the channel capacities before and after the payment. The channel capacity is the node's free capacity as reported to the PFS with the UpdatePFS messages.
+
+The IP function is a function that describes how much a node is willing to pay to move away from an undesirable channel capacity. If a node prefers to have a channel capacity of 5 while the total capacity of that channel is 10 (so that it could mediate up to 5 tokens in both directions) the IP function might look like
```
palango: We need to clarify if IP is relative or absolute values, and if it's token or token weis.

adr/003-mediation-fees.md
```diff
@@ -0,0 +1,146 @@
...
+
+In this model, the mediation fees are split into three parts:
+
+1. The *flat fee*. This is meant to represent the base cost of handling a payment, so that even small payment amounts can be profitable to mediate.
+2. The *proportional fee*. Large payments lock up tokens while the payment is in progress. To compensate this, the operator should receive larger fees for large payments
+3. The *imbalance fee*. The direction of payments and the specific channels used are relevant to the operator. Some payments might exhaust the channel capacity and prevent future profitable mediations while others could facilitate future mediations. Due to this, the operator should be compensated (or willing to compensate the initiator, respectively). As a side effect, the whole network will get into a more balanced state.
+
+These three parts roughly map to operating costs, capital costs and opportunity costs.
+
+The flexible shape of the IP function allows to encode many different operator intentions. If I use a channel solely to pay, having more free capacity is always desired and any capacity gained by mediating in the reverse direction is welcome. In that case, my IP function could look like
```
palango: Nice examples! 👍

adr/003-mediation-fees.md
```diff
@@ -0,0 +1,146 @@
...
+
+The message format still has to be defined in detail, but the payload will look roughly like this:
+
+```json
+{
+    'flat': 10,
+    // we have to deal with float values anyway, due to the interpolation
+    'proportional': 0.0001,  // factor of transfer amount
+    'imbalance_penalty': [
+        (0, 1000),
```
palango: These should be lists, there are no tuples in json.

palango:
@hackaugusto Please give this a look.

adr/003-mediation-fees.md
karlb: We could have the flat and proportional fee either
1. set per node
2. set per channel, used only for outgoing channel
3. set per channel, used for both incoming and outgoing channel

Option 2 feels like slightly awkward compromise to me. Option 1 would introduce fee information per node, which we wouldn't have with 2 or 3.

adr/003-mediation-fees.md
karlb: Under "Message Format" I write

> The three mediation fee components are sent separately. Each component is optional and will be assumed to be zero (no cost) if not included in the message.

Do you expect a different statement, or do you just want this information in another place of the document?

adr/003-mediation-fees.md
karlb: I'll add
> The IP function is a function that describes how much a node is willing to pay The IP function is a function that describes how much a node is willing to pay **(as absolute amount or RDN-wei)** to move ...

RDN-wei is not something we've used before, but at least it's not ambiguous.

palango: I think we call it Rei, for RDN wei

adr/003-mediation-fees.md
palango: I would just state it as an advantage of the solution.

adr/003-mediation-fees.md
karlb: I just noticed after commenting that we're not talking about RDN and edited my question.

adr/003-mediation-fees.md
karlb: I've added

> Fee components are optional. A node only using flat fees does not have to submit complicated functions.

adr/003-mediation-fees.md
karlb: Added a section for the example.

adr/003-mediation-fees.md
```diff
@@ -74,6 +77,24 @@ If the node currently has a capacity of 6 and is asked to mediate a payment of 3
 
 Only mediating nodes demand mediation fees. The initiator and target could theoretically benefit if their IP was taken into account during routing, but this aspect is left out for now to avoid additional complexity.
 
+### Why apply fees for the incoming channel?
+
+At a first glance, calculating fees for the incoming channel does not make any sense, since the mediating node itself only sends money over the outgoing channel. But taking part in the mediation will modify both channels channels, so the choice of the incoming channel is relevant to the mediating node. To demonstrate that you can't reliably influence the balancing of your channel if you only apply fees to the outgoing channel, let's look at one example:
+
+A mediating node has three channels. The channel to C is nearly exhausted, while the other channels have a free capacity of 9 tokens left (of 10 tokens which are deposited in that channel). This node wants to have
+a similar amount of free capacity for all channel, so that it has more mediation opportunities. To achieve this, it must receive tokens over channel C.
```
palango: for all channel**s**

adr/003-mediation-fees.md
```diff
@@ -74,6 +77,24 @@ If the node currently has a capacity of 6 and is asked to mediate a payment of 3
...
+++: desirable transfer (point of view of the mediating node)<br/>
+--: undesirable transfer<br/>
+
+| from\to | A   | B   | C   |
+| ---     | --- | --- | --- |
+| A 1/10  |     | -   | --  |
+| B 1/10  | -   |     | --  |
+| C 9/10  | ++  | ++  |     |
+
+To incentivize rebalancing, the node could reduce fees for channel A and B to a negative value. But this might lead to mediating transfers A&rarr;B and B&rarr;A, both of which are not intended! The problem is visible in the table above: by changing fees for outgoing channels, you can only incentivize all transfers in a single column. But all desirable transfers are in a single row, meaning the we must incentivize usage of an incoming channel.
```
palango: channel**s**

adr/003-mediation-fees.md
```diff
@@ -74,6 +77,24 @@ If the node currently has a capacity of 6 and is asked to mediate a payment of 3
 
 Only mediating nodes demand mediation fees. The initiator and target could theoretically benefit if their IP was taken into account during routing, but this aspect is left out for now to avoid additional complexity.
 
+### Why apply fees for the incoming channel?
```
palango: 🥇 I think the example helps understand the problem!

Dominik1999:
Thank you @karl I think this is a very good approach and I like the precise and concise description. Maybe @franzihei can even make a blogpost out of this? I think, when we follow your approach karl, we have a quite nice and unique way compared to the approaches discussed in e.g. Lightning.

franzihei:
Regarding a blog post: Yes, that would be super nice as a piece in out "tech deep dive" series! @karlb let's sit together and briefly discuss this next week?

PR #352, opened 2019-05-10
Enable auto restart for builder

palango:
The builder sometimes crashes and didn't have `restart: always` enabled.


palango:
No clue.

PR #356, opened 2019-05-16
Start RequestCollector greenlet

palango:
Currently the RequestCollector doesn't start, as the RC greenlet is not started and thus the MatrixListener doesn't log in properly.

~This is not a nice solution yet, but I haven't found a nice way to wait until the startup process is finished properly. @ulope any hints? Probably some event that is set once the startup is finished would help?~

### DIFF
--- a/src/request_collector/cli.py
+++ b/src/request_collector/cli.py
@@ -20,7 +20,10 @@ def main(private_key: str, state_db: str) -> int:
 
     database = SharedDatabase(state_db)
 
-    RequestCollector(private_key=private_key, state_db=database).listen_forever()
+    service = RequestCollector(private_key=private_key, state_db=database)
+
+    service.start()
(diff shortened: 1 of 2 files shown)


src/request_collector/cli.py
```diff
@@ -20,7 +20,12 @@ def main(private_key: str, state_db: str) -> int:
 
     database = SharedDatabase(state_db)
 
-    RequestCollector(private_key=private_key, state_db=database).listen_forever()
+    service = RequestCollector(private_key=private_key, state_db=database)
+
+    service.start()
+    service.matrix_listener.event.wait()
```
karlb: If we move the `wait` into `listen_forever`, we could avoid exposing this event and keep it as an implementation detail.

palango: Nice, I like that much better!

PR #358, opened 2019-05-16
Fix types in MonitorRequest

palango:
I think this still needs database changes, but we should also see why this is not tested.


src/monitoring_service/schema.sql
```diff
@@ -38,14 +38,14 @@ CREATE TABLE monitor_request (
...
-    additional_hash         CHAR(32)    NOT NULL,
-    closing_signature       CHAR(34)    NOT NULL,
+    additional_hash         CHAR(66)    NOT NULL,
+    closing_signature       CHAR(132)   NOT NULL,
 
-    non_closing_signature   CHAR(160)   NOT NULL,
+    non_closing_signature   CHAR(132)   NOT NULL,
     reward_amount           HEX_INT     NOT NULL,
-    reward_proof_signature  CHAR(42)    NOT NULL,
+    reward_proof_signature  CHAR(132)   NOT NULL,
```
karlb: Those were from the old code base. They looked suspicious, but I never got around to check them. Thanks for fixing this!

tests/monitoring/monitoring_service/factories.py
```diff
@@ -0,0 +1,77 @@
...
+
+    # Some signature correctness checks
+    assert monitor_request.signer == private_key_to_address(closing_privkey)
+    assert monitor_request.non_closing_signer == private_key_to_address(nonclosing_privkey)
+    assert monitor_request.reward_proof_signer == private_key_to_address(nonclosing_privkey)
+
+    return monitor_request
+
+
+def get_random_channel_identifier() -> ChannelID:
```
karlb: Is this used anywhere?

tests/monitoring/monitoring_service/test_database.py
```diff
@@ -37,24 +47,24 @@ def test_scheduled_events(ms_database):
...
-    assert len(ms_database.get_scheduled_events(24)) == 2
+    assert len(ms_database.get_scheduled_events(BlockNumber(22))) == 0
+    assert len(ms_database.get_scheduled_events(BlockNumber(23))) == 1
+    assert len(ms_database.get_scheduled_events(BlockNumber(24))) == 2
 
-    ms_database.remove_scheduled_event(event2)
-    assert len(ms_database.get_scheduled_events(22)) == 0
-    assert len(ms_database.get_scheduled_events(23)) == 1
-    assert len(ms_database.get_scheduled_events(24)) == 1
+    ms_database.remove_scheduled_event(event=event2)
```
karlb: No need to use kwargs here, IMO.

tests/monitoring/monitoring_service/factories.py
```diff
@@ -0,0 +1,77 @@
...
+        token_network_address=DEFAULT_TOKEN_NETWORK_ADDRESS,
+        identifier=DEFAULT_CHANNEL_IDENTIFIER,
+        participant1=DEFAULT_PARTICIPANT1,
+        participant2=DEFAULT_PARTICIPANT2,
+        settle_timeout=random.randint(0, UINT256_MAX),
+        state=random.choice(list(ChannelState)),
+        closing_block=BlockNumber(random.randint(0, UINT256_MAX)),
+        closing_participant=DEFAULT_PARTICIPANT1,
+        closing_tx_hash=TransactionHash("%d" % random.randint(0, UINT64_MAX)),
+        claim_tx_hash=TransactionHash("%d" % random.randint(0, UINT64_MAX)),
```
karlb: Aren't transaction hashes hex encoded?
(At least until we do https://github.com/raiden-network/raiden-services/issues/364)

src/monitoring_service/schema.sql
palango: Strange that is worked anyways, shouldn't the values be truncated?

tests/monitoring/monitoring_service/factories.py
palango: Yes, the `TransactionHash` is our type from `raiden_libs.types` if that's what you wonder.

tests/monitoring/monitoring_service/factories.py
palango: Nope :)

src/monitoring_service/schema.sql
karlb: Not on SQLite. It takes types more as a suggestion rather than anything strict, which is really odd for and SQL database. See https://www.sqlite.org/datatype3.html for details.

tests/monitoring/monitoring_service/factories.py
karlb: The we should use `%x` instead of `%d`.

PR #367, opened 2019-05-17
Log more infos about presence updates

palango:
Log the current presence as well.

### DIFF
--- a/src/raiden_libs/matrix.py
+++ b/src/raiden_libs/matrix.py
@@ -168,11 +168,17 @@ class MatrixListener(gevent.Greenlet):
 
     def follow_address_presence(self, address: Address, refresh: bool = False) -> None:
         if self.user_manager:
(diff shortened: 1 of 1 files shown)


src/raiden_libs/matrix.py
```diff
@@ -168,12 +168,18 @@ def _start_client(self) -> None:
...
-            log.debug("Tracking address", address=to_checksum_address(address))
             self.user_manager.add_address(address)
 
             if refresh:
                 self.user_manager.refresh_address_presence(address)
 
+            log.debug(
+                "Tracking address",
+                address=to_checksum_address(address),
+                current_presence=self.user_manager.get_address_reachability(address),
```
karlb: Does this have a meaningful value at this time?

palango: At least when refresh is true it should have. But I guess this will not stay and is only to get some more infos to debug the presence problems

Issue #368, opened 2019-05-20
Implement flexible mediation fees in PFS


karlb:
According to https://github.com/raiden-network/raiden-services/blob/master/adr/003-mediation-fees.md

- [x] Create a message type to set new meditation fees (PFS part only)
- [x] Store mediation fees in db (and retrieve then on startup)
- [x] Add code to evaluate mediation fees (based on direction and current capacity; add interpolation)
- [x] Use mediation fees during routing
- [x] Add tests
- [x] Add compounding effect in mediation fee calculation
- [x] Add tests for the interpolation
- [x] Accept FeeUpdate message via network
- [x] Make sure that FeeUpdate message is compatible with Raiden's message handling
- [x] Use `Tuple` typing after updating Raiden
- [x] Handle cases when imbalance fee calculation is out of bounds
- [x] Avoid replay attacks for fees (probably signature + nonce)

PR #369, opened 2019-05-21
Flexible mediation fee calculation

karlb:
This is the main work for https://github.com/raiden-network/raiden-services/issues/368. Accepting the UpdateFee messsage over the network is still left out. I'll add that in the next PR after having a look at Raiden's message handling after the dataclass PR and updating to the latest Raiden version.

### DIFF
--- a/src/pathfinding_service/model/channel_view.py
+++ b/src/pathfinding_service/model/channel_view.py
@@ -76,8 +76,12 @@ class ChannelView:
         if reveal_timeout is not None:
             self.reveal_timeout = reveal_timeout
 
-    def fee(self, amount: TokenAmount) -> int:
+    def fee_out(self, amount: TokenAmount) -> int:
         """Return the mediation fee for this channel when transferring the given amount"""
+        return int(self.fee_schedule.flat + amount * self.fee_schedule.proportional)
+
+    def fee_in(self, amount: TokenAmount) -> int:
+        """Return the mediation fee for this channel when receiving the given amount"""
         return int(self.fee_schedule.flat + amount * self.fee_schedule.proportional)
 
     def __repr__(self) -> str:
--- a/src/pathfinding_service/model/channel_view.py
+++ b/src/pathfinding_service/model/channel_view.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass, field
-from typing import ClassVar, Type
+from typing import ClassVar, List, Optional, Type
 
 import marshmallow
 from marshmallow_dataclass import add_schema
@@ -14,6 +14,26 @@ from raiden.utils.typing import (
     TokenNetworkAddress,
 )
 from raiden_libs.marshmallow import ChecksumAddress
+
+
+@dataclass
+class FeeSchedule:
+    flat: FeeAmount
+    proportional: float
+    imbalance_penalty: Optional[List[List[FeeAmount]]] = None
+
+    def reversed(self) -> "FeeSchedule":
+        # pylint: disable=not-an-iterable
+        if not self.imbalance_penalty:
+            return self
+        max_x = max(x for x, penalty in self.imbalance_penalty)
+        return FeeSchedule(
+            flat=self.flat,
+            proportional=self.proportional,
+            imbalance_penalty=[
+                [FeeAmount(max_x - x), penalty] for x, penalty in self.imbalance_penalty
+            ],
+        )
 
 
 @add_schema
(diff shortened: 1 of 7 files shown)


src/pathfinding_service/model/channel_view.py
```diff
@@ -16,6 +16,26 @@
 from raiden_libs.marshmallow import ChecksumAddress
 
 
+@dataclass
+class FeeSchedule:
+    flat: FeeAmount
+    proportional: float
```
palango: Below (line 58) `proportional` is of type `FeeAmount`  as well.

src/pathfinding_service/model/token_network.py
```diff
@@ -76,6 +78,14 @@ def is_valid(self) -> bool:
         return True
 
 
+@dataclass
+class FeeUpdate:
```
palango: I think we need a signature and nonce here, otherwise this can be replayed.

src/pathfinding_service/model/channel_view.py
```diff
@@ -16,6 +16,26 @@
...
+    imbalance_penalty: Optional[List[List[FeeAmount]]] = None
```
palango: I don't understand this type. Based on what happens below, shouldn't this be `Optional[List[Tuple[TokenAmount, FeeAmount]]]`, a list of pairs `(balance, penalty)`?

src/pathfinding_service/model/channel_view.py
```diff
@@ -76,10 +76,14 @@ def update_capacity(
         if reveal_timeout is not None:
             self.reveal_timeout = reveal_timeout
 
-    def fee(self, amount: TokenAmount) -> int:
+    def fee_out(self, amount: TokenAmount) -> int:
```
palango: This should return a `FeeAmount`.

src/pathfinding_service/model/channel_view.py
```diff
@@ -16,6 +17,60 @@
...
+            return self.y_list[-1]
+        i = bisect_right(self.x_list, x) - 1
+        return self.y_list[i] + self.slopes[i] * (x - self.x_list[i])
+
+
+@dataclass
+class FeeSchedule:
+    # pylint: disable=not-an-iterable
+    flat: FeeAmount = FeeAmount(0)
+    proportional: float = FeeAmount(0)
```
palango: Something doesn't fit here, `FeeAmount` is an int.

src/pathfinding_service/model/channel_view.py
```diff
@@ -16,6 +17,60 @@
...
+        if self.imbalance_penalty:
+            # Total channel capacity - node capacity = balance (used as x-axis for the penalty)
+            balance = self._penalty_func.x_list[-1] - capacity
+            imbalance_fee = self._penalty_func(balance + amount) - self._penalty_func(balance)
+        else:
+            imbalance_fee = 0
+        return FeeAmount(round(self.flat + amount * self.proportional + imbalance_fee))
+
+    def reversed(self) -> "FeeSchedule":
+        if not self.imbalance_penalty:
```
palango: Should we check for explicitely for `None`  here? Not sure what would happen if the client would send an empty list!?

src/pathfinding_service/model/channel_view.py
karlb: marshmallow2 has problems with tuples. I'll update this after updating Raiden.

src/pathfinding_service/model/channel_view.py
karlb: For empty lists we could either ignore them (as done now) or throw an error. Since it is not a useful input, I would be fine with either.

karlb:
I've added the remaining tasks to the todo list in https://github.com/raiden-network/raiden-services/issues/368.

src/pathfinding_service/model/channel_view.py
```diff
@@ -16,6 +17,60 @@
 from raiden_libs.marshmallow import ChecksumAddress
 
 
+class Interpolate:  # pylint: disable=too-few-public-methods
```
palango: Can you add some explicit tests for this class.

src/pathfinding_service/model/channel_view.py
```diff
@@ -16,6 +17,60 @@
...
+    def __init__(self, x_list: Sequence, y_list: Sequence):
+        if any(y - x <= 0 for x, y in zip(x_list, x_list[1:])):
+            raise ValueError("x_list must be in strictly ascending order!")
+        self.x_list = x_list
+        self.y_list = y_list
+        intervals = zip(x_list, x_list[1:], y_list, y_list[1:])
+        self.slopes = [(y2 - y1) / (x2 - x1) for x1, x2, y1, y2 in intervals]
+
+    def __call__(self, x: float) -> float:
```
palango: Why is this a float?

src/pathfinding_service/model/token_network.py
```diff
@@ -31,14 +33,20 @@ def __init__(
...
 
     @property
     def edge_attrs(self) -> Iterable[dict]:
         return (self.G[node1][node2] for node1, node2 in zip(self.nodes[:-1], self.nodes[1:]))
 
     def to_dict(self) -> dict:
-        fee = sum(edge["view"].fee(self.value) for edge in self.edge_attrs)
-        return dict(path=[to_checksum_address(node) for node in self.nodes], estimated_fee=fee)
+        return dict(
+            path=[to_checksum_address(node) for node in self.nodes], estimated_fee=sum(self.fees)
```
palango: I think `sum` is wrong here, as the fees later in a given route have to be included in the earlier ones.

PR #371, opened 2019-05-23
Use clients TransactionHash type

### LINKED ISSUE #364: Convert tx hash variables to bytes
Currently transaction hashes are stored as `str`. Raiden uses:
```
T_TransactionHash = bytes
TransactionHash = NewType("TransactionHash", T_TransactionHash)
```

With that we can also remove the last type definition from `raiden_libs.types`.

palango:
Fixes #364

### DIFF
--- a/src/monitoring_service/database.py
+++ b/src/monitoring_service/database.py
@@ -3,7 +3,7 @@ import sqlite3
 from typing import List, Optional, Union, cast
 
 import structlog
-from eth_utils import decode_hex, to_canonical_address, to_checksum_address, to_hex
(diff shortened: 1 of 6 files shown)


palango:
> I hope that there will be a clear consensus on all these types across all python ethereum projects, soon. Not getting the "right" types from web3.py is a bit annoying. Or do they return hash32 (https://github.com/ethereum/eth-typing/blob/ac4c7f40385c23dcf34e86f5d7df1178b5be8dba/docs/types.rst#hash32) and it's just us not sticking to the conventions?

I think stuff is moving in the right direction, it's just taking time with all these little packages...

PR #374, opened 2019-05-24
Mediation fee improvements

karlb:
Relates to https://github.com/raiden-network/raiden-services/issues/368.

### DIFF
--- a/src/pathfinding_service/model/channel_view.py
+++ b/src/pathfinding_service/model/channel_view.py
@@ -6,6 +6,7 @@ import marshmallow
 from marshmallow_dataclass import add_schema
 
 from pathfinding_service.config import DEFAULT_REVEAL_TIMEOUT
+from pathfinding_service.exceptions import UndefinedFee
 from raiden.utils.typing import (
     Address,
     ChannelID,
@@ -58,7 +59,10 @@ class FeeSchedule:
         if self.imbalance_penalty:
             # Total channel capacity - node capacity = balance (used as x-axis for the penalty)
             balance = self._penalty_func.x_list[-1] - capacity
-            imbalance_fee = self._penalty_func(balance + amount) - self._penalty_func(balance)
+            try:
+                imbalance_fee = self._penalty_func(balance + amount) - self._penalty_func(balance)
+            except ValueError:
+                raise UndefinedFee()
         else:
             imbalance_fee = 0
         return FeeAmount(round(self.flat + amount * self.proportional + imbalance_fee))
--- a/src/pathfinding_service/model/channel_view.py
+++ b/src/pathfinding_service/model/channel_view.py
@@ -1,6 +1,6 @@
 from bisect import bisect_right
 from dataclasses import dataclass, field
-from typing import ClassVar, List, Optional, Sequence, Type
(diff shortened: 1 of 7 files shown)


src/pathfinding_service/model/channel_view.py
```diff
@@ -18,6 +19,11 @@
 
 
 class Interpolate:  # pylint: disable=too-few-public-methods
```
palango: Just a general question: might it be helpful to use numpy here? I'm not sure if it's worth the dependency, but interpolations should be implemented there and probably more performant.

tests/pathfinding/test_fee_schedule.py
```diff
@@ -54,35 +57,34 @@ def a(int_addr) -> Address:  # pylint: disable=invalid-name
...
+            cv.capacity = 100
 
-    def set_fee(node1, node2, fee_schedule: FeeSchedule):
-        channel_id = network.G[a(node1)][a(node2)]["view"].channel_id
-        network.handle_channel_fee_update(
+        # make nodes reachable
+        for node in self.G.nodes:
+            self.address_to_reachability[node] = AddressReachability.REACHABLE
+
+    def set_fee(self, node1, node2, fee_schedule: FeeSchedule):
```
palango: Can you type `node1` and `node2` here?

tests/pathfinding/test_fee_schedule.py
```diff
@@ -91,31 +93,91 @@ def set_fee(node1, node2, fee_schedule: FeeSchedule):
             )
         )
 
-    def estimate_fee(initator, target, value=TA(10), max_paths=1):
-        result = network.get_paths(a(initator), a(target), value=value, max_paths=max_paths)
+    def estimate_fee(self, initator, target, value=TA(10), max_paths=1):
```
palango: type annotation here as well.

tests/pathfinding/test_fee_schedule.py
```diff
@@ -91,31 +93,91 @@ def set_fee(node1, node2, fee_schedule: FeeSchedule):
...
     # Now let's try imbalance fees
-    set_fee(2, 3, FeeSchedule(imbalance_penalty=[[TA(0), TA(0)], [TA(200), TA(200)]]))
-    assert estimate_fee(1, 3) == 10
-    assert estimate_fee(3, 1) == -10
+    tn.set_fee(2, 3, FeeSchedule(imbalance_penalty=[(TA(0), FA(0)), (TA(200), FA(200))]))
+    assert tn.estimate_fee(1, 3) == 10
+    assert tn.estimate_fee(3, 1) == -10
+
+    # When the range covered by the imbalance_penalty does include the
+    # necessary belance values, the route should be considered invalid.
```
palango: belance -> balance

src/pathfinding_service/model/channel_view.py
karlb: It's certainly faster and more mature. I just wanted to avoid adding a problematic dependency. Numpy needs quite a lot of build tools if there's not prebuilt wheel for your system.

palango: Yeah, we can still add this later.

PR #375, opened 2019-05-24
Allow running tests in parallel with pytest-xdist

karlb:
The blocker for this was that gevent monkeypatching broke xdist. By
skipping `subprocess` and `thread` monkeypatching, this can be avoided.

We also need to make sure that not all processes choose the same port.
At least in my tests, choosing a free port by binding to port 0 did not
cause problematic race conditions, although I'm not sure that it always
works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,5 +10,5 @@ script:
   - pip install -r requirements-dev.txt
   - pip install -e .
   - make lint
-  - pytest -v --cov=./ tests/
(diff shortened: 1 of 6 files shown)


karlb:
> How does the client handle port selection? Maybe we can copy that approach?

The client's approach does not look any better, as far as I understand it. I'm currently copying over our new approach to make the tests run in parallel there, too.

Issue #270, opened 2019-04-17
Improve logging


palango:
Currently logging in the API is course, so it's hard to debug problems. We should have better logging, especially for errors.

palango:
Also more/better logging for the routing process is important for debugging failures.

palango:
For example the capacity update log is still ugly:

```
2019-06-06 12:14:39.775763 [info     ] Received Capacity Update       [pathfinding_service.service] canonical_identifier={'chain_identifier': 4, 'token_network_address': b'\x7f[\xf6\xc9To\xa8\x185w\xe4\x9f\x15\xbc\xef@mr\xd5\xd9', 'channel_identifier': 1} mediation_fee=0 other_capacity=950000000000000000000 other_nonce=2 other_participant=b'&E\xbc\xf9]\x1c?\xe8\x9d\xb45\xbe\x8a\xae\xfb\x88x2\x80;' reveal_timeout=50 signature=b'\xf1M\xe6\xb2\x08\x944\xef/\x8dk_\x8a`\x9c\xb8\xd2\x87kI\xc0j\xa2Q\x8cxy\xa9I\xd3/fp\xbe\xe5\xcaP\xd2\x92\xcar\xe0\x12\xa0c\x88D\x7fs\xe4\x85bf\xd9p\xa4:\xb7\x98@\x9cs\xde\xf7\x1c' updating_capacity=50000000000000000000 updating_nonce=0 updating_participant=b'_\xbf-\xd2\x03,+2\xcb\xd5\x90\x94\x01\xb6\x10\xba?\xc8\xfb\xde'
```

PR #377, opened 2019-05-28
Small logging improvements

palango:
Small logging additions

Ref #270


palango:
> nice that is helpful ... there are more logs where we show bytes nowadays, right?

Yes, that's the sad side effect of switching the internal representation. Whenever you see that you can create a small PR.

PR #382, opened 2019-05-29
Use a global presence mapping

### LINKED ISSUE #350: More efficient presence tracking
See discussion https://github.com/raiden-network/raiden-services/pull/326#discussion_r282775643

Presence tracking is currently done on a per network basis and involves iteration over a list of nodes. This is potentially slow and can be improved.

### LINKED ISSUE #380: Wrong presence for newly found token networks
## Problem Definition

The scenario player found an issue with how presence is handled in the PFS.

Currently there is a mapping `address_to_reachability` for each token network. This mapping is updated when presence of a node changes.

However, when a new token network is found it is not properly updated and therefore valid routes are not found.

palango:
Keeping a per token network mapping of addresses to presences results in
more complicated code, bugs (#380) and inefficiencies (#350).

This commit changes the presence tracking to be handles in the service
itself.

Fixes #380
Fixes #350

### DIFF
--- a/tests/pathfinding/fixtures/__init__.py
+++ b/tests/pathfinding/fixtures/__init__.py
@@ -1,7 +1,8 @@
 import pytest
 
 from pathfinding_service.model.token_network import TokenNetwork
-from raiden.utils.typing import TokenNetworkAddress
+from raiden.network.transport.matrix import AddressReachability
+from raiden.utils.typing import Address, Dict, TokenNetworkAddress
 
 from .accounts import *  # noqa
 from .api import *  # noqa
@@ -12,3 +13,8 @@ from .network_service import *  # noqa
 @pytest.fixture
 def token_network_model() -> TokenNetwork:
     return TokenNetwork(TokenNetworkAddress(bytes([1] * 20)))
+
+
+@pytest.fixture
+def address_to_reachability() -> Dict[Address, AddressReachability]:
+    return {}
--- a/tests/pathfinding/test_fee_schedule.py
+++ b/tests/pathfinding/test_fee_schedule.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import Dict, List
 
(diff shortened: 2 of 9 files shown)


tests/pathfinding/fixtures/__init__.py
```diff
@@ -12,3 +13,8 @@
 @pytest.fixture
 def token_network_model() -> TokenNetwork:
     return TokenNetwork(TokenNetworkAddress(bytes([1] * 20)))
+
+
+@pytest.fixture
+def address_to_reachability() -> Dict[Address, AddressReachability]:
+    return {}
```
karlb: Does this fixture really help? Just putting `{}` inside the test would both shorter and one indirection less.

tests/pathfinding/test_fee_schedule.py
```diff
@@ -74,7 +74,8 @@ def __init__(self, channels: List[dict]):
         for _, _, cv in self.G.edges(data="view"):
             cv.capacity = 100
 
-        # make nodes reachable
+        # create reachability mapping for testing
+        self.address_to_reachability: Dict[Address, AddressReachability] = {}
         for node in self.G.nodes:
             self.address_to_reachability[node] = AddressReachability.REACHABLE
```
karlb: Now that we construct the dict here, we could use a dict comprehension
```python
self.address_to_reachability: Dict[Address, AddressReachability] = {
    node: AddressReachability.REACHABLE
    for node in self.G.nodes
}
```
Not much of a difference, but a bit easier to read IMO.

tests/pathfinding/fixtures/__init__.py
palango: The thing is that it gets populated by the `populate_token_network_case_X` fixtures, so I think it's necessary (but we should improve that setup system)

PR #388, opened 2019-06-04
Update `raiden` and `raiden-contracts`

karlb:
Also change the requirements substitution in setup.py to regexes, so
that we don't have to put the raiden revision in two places.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,15 +1,14 @@
 #!/usr/bin/env python3
 import io
 import os
+import re
 from glob import glob
 from os.path import basename, dirname, join, splitext
 from typing import List
 
 from setuptools import find_packages, setup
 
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -1,15 +1,14 @@
...
 from glob import glob
 from os.path import basename, dirname, join, splitext
 from typing import List
 
 from setuptools import find_packages, setup
 
-REQ_REPLACE = {
-    "git+https://github.com/raiden-network/raiden.git@d52a9e65b98af755c31de47c7a121c3172337219": "raiden"  # noqa
-}
+REQ_REPLACE = {re.compile(r"git\+https://github.com/raiden-network/raiden.git@.*"): "raiden"}
```
palango: Nice one!

PR #389, opened 2019-06-04
Calculate basic routing stats from feedback

### LINKED ISSUE #383: Stats debug endpoint
For https://github.com/raiden-network/raiden-services/issues/276 we need a way to access the PFS internal data and statistics. 

I propose we add a debug endpoint that provides the basic data as well as some more advances statistics.

palango:
Closes #383 

This only calculates basic information right now, but more detailed stats can be added when necessary.

### DIFF
--- a/src/pathfinding_service/database.py
+++ b/src/pathfinding_service/database.py
@@ -227,3 +227,14 @@ class PFSDatabase(BaseDatabase):
             )
 
         return None
+
+    def get_num_routes_feedback(
+        self, only_with_feedback: bool = False, only_successful: bool = False
+    ) -> int:
+        where_clause = ""
+        if only_with_feedback:
+            where_clause = "WHERE successful IS NOT NULL"
+        elif only_successful:
+            where_clause = "WHERE successful"
+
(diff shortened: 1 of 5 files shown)


src/pathfinding_service/database.py
```diff
@@ -227,3 +227,14 @@ def get_feedback_token(
...
+    def get_num_routes_feedback(
+        self, only_with_feedback: bool = False, only_successful: bool = False
+    ) -> int:
+        where_clause = ""
+        if only_with_feedback:
+            where_clause = "WHERE successful IS NOT NULL"
+        elif only_successful:
+            where_clause = "WHERE successful"
+
+        return self.conn.execute(f"SELECT COUNT(*) FROM feedback {where_clause};").fetchone()[0]
```
karlb: If you wanted to do all these in a single pass through the table, you could do
```sql
SELECT
    count(*) AS all
    count(CASE WHEN feedback IS NOT NULL THEN 1 END) AS with_feedback,
    count(CASE WHEN feedback THEN 1 END) AS success
FROM feedback
```
Just because you asked before, not because any change is needed, now.

PR #391, opened 2019-06-04
Adds chain-id and contract version to db name

### LINKED ISSUE #390: Add contract version and chain id to db name
We should add the contract version/address and the chain id to the name of the state db, so it's easier to run multiple version along each other.

Dominik1999:
Closes #390

### DIFF
--- a/src/raiden_libs/cli.py
+++ b/src/raiden_libs/cli.py
@@ -20,6 +20,7 @@ from raiden_contracts.constants import (
     CONTRACT_ONE_TO_N,
     CONTRACT_TOKEN_NETWORK_REGISTRY,
     CONTRACT_USER_DEPOSIT,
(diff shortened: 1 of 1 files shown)


src/raiden_libs/cli.py
```diff
@@ -99,6 +99,12 @@ def call_with_common_options_initialized(**params: Any) -> Callable:
             )
             try:
                 setup_logging(params.pop("log_level"))
+                if not params["state_db"]:
+                    params["state_db"] = os.path.join(
+                        "deployment",
+                        "state",
+                        app_name + "-" + str(params["web3"].net.version) + "-" + CONTRACTS_VERSION,
```
palango: Should this have a `.db` in the end?

Dominik1999: yes

PR #392, opened 2019-06-06
Changes logging of MS and PFS to human readable logs

### LINKED ISSUE #270: Improve logging

Dominik1999:
Closes #270


Dominik1999:
I will recheck, good point!

PR #393, opened 2019-06-06
Adds better reading for Capacity Update logs

Dominik1999:
more to #270


Dominik1999:
- we could also add a new function in libs `log_message` analogous to `log_event`
- we could also try to find out how to format in `structlog` in general

palango:
>     * we could also try to find out how to format in `structlog` in general

Good point, I think it would be better to handle this in one place than to have conversions everywhere.  This should be possible by writing a [structlog processor](http://www.structlog.org/en/stable/processors.html)

src/raiden_libs/logging.py
```diff
@@ -34,14 +35,22 @@ def setup_logging(log_level: str) -> None:
     )
 
 
-def log_event(event: Event) -> Dict[str, Any]:
-    event_data = asdict(event)
+def change_bytes(event_dict: Dict, key: Any, val: bytes) -> Dict[str, Any]:
```
palango: Please add a test for this.

src/raiden_libs/logging.py
```diff
@@ -34,14 +35,22 @@ def setup_logging(log_level: str) -> None:
...
 
-    for key, val in event_data.items():
-        if isinstance(val, bytes):
-            if len(val) == 20:
-                event_data[key] = to_checksum_address(val)
-            else:
-                event_data[key] = to_hex(val)
 
-    return event_data
+def format_to_hex(_logger: Any, _log_method: Any, event_dict: Dict) -> Dict[str, Any]:
```
palango: Please add a test as well

src/raiden_libs/logging.py
```diff
@@ -34,14 +35,22 @@ def setup_logging(log_level: str) -> None:
...
+    for key, val in event_dict.items():
+        if isinstance(val, bytes):
+            change_bytes(event_dict, key, val)
+        if isinstance(val, Event):
```
palango: In this case it would be nice to output the event name as well

src/raiden_libs/logging.py
Dominik1999: done

src/raiden_libs/logging.py
Dominik1999: done

src/raiden_libs/logging.py
Dominik1999: Wouldn't that be another logging processor? This should only change bytes to hex and neither delete or add any info, right?

palango: but you're changing it here: `event_data = asdict(val)`

Dominik1999: So when I understand you correctly, then you suggest to add another structlog processor that also includes the Event name. Or would you inlcude that in the `format_to_hex` function?

palango: I'd include it there, as you loose the name when you convert it to a dictionary.

tests/pathfinding/test_service.py
```diff
@@ -333,3 +335,29 @@ def test_invalid_fee_update(pathfinding_service_mock, token_network_model):
     # bad/missing signature
     with pytest.raises(exceptions.InvalidFeeUpdate):
         pathfinding_service_mock.on_fee_update(fee_update)
+
+
+def test_logging_processor():
+    # test if our logging processor changes bytes to checksum addresses
+    # even if bytes-addresses are entangeled into dicts
```
palango: You never test with a dict.

Dominik1999: ok good point!

src/raiden_libs/logging.py
Dominik1999: finally successfully implemented

palango:
@Dominik1999 you need to rebase again!

Dominik1999:
> @Dominik1999 you need to rebase again!

I seem to be able to merge without rebasing. Should I rebase anyway?

palango:
If you can go for it.

PR #394, opened 2019-06-06
 Receive and handle `FeeUpdate` generated by client

karlb:
The messages have to be stored until a ChannelOpened event has been
received, if that is no the case yet when receive the `FeeUpdate`. Still missing:
- Nonce handling
- Scenario to test the client/PFS integration

### DIFF
--- a/src/pathfinding_service/model/token_network.py
+++ b/src/pathfinding_service/model/token_network.py
@@ -1,5 +1,5 @@
+import dataclasses
 from collections import defaultdict
-from dataclasses import dataclass
 from itertools import islice
 from typing import Any, Dict, Iterable, List, Optional, Sequence, Tuple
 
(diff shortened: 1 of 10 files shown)


src/pathfinding_service/model/token_network.py
```diff
@@ -260,12 +251,15 @@ def handle_channel_balance_update_message(
         )
 
     def handle_channel_fee_update(self, message: FeeUpdate) -> None:
+        channel_id = message.canonical_identifier.channel_identifier
+        participants = self.channel_id_to_addresses[channel_id]
+        other_participant = (set(participants) - {message.updating_participant}).pop()
```
palango: Would it make sense to move this into a function? I think we do this a lot in the `service` as well

karlb: I can't see such a case in `pathfinding_service/service.py`. The `UpdatePFS` already includes the other participant, so that logic is not necessary, there.

PR #396, opened 2019-06-11
Add chain id parameter to MSRC

palango:
As discussed before.

### DIFF
--- a/src/raiden_libs/cli.py
+++ b/src/raiden_libs/cli.py
@@ -106,9 +106,10 @@ def common_options(app_name: str) -> Callable:
                     filename = f"{app_name}-{chain_id}-{contracts_version}.db"
                     params["state_db"] = os.path.join("deployment", "state", filename)
 
-                    # remove `chain_id` paramater, when available
-                    if params.get("chain_id") is not None:
-                        params.pop("chain_id")
+                # Need to delete the `chain_id` key
+                if params.get("chain_id") is not None:
+                    del params["chain_id"]
+
                 return func(**params)
             finally:
                 structlog.reset_defaults()
--- a/src/raiden_libs/cli.py
+++ b/src/raiden_libs/cli.py
@@ -100,10 +100,15 @@ def common_options(app_name: str) -> Callable:
             try:
                 setup_logging(params.pop("log_level"))
                 if not params["state_db"]:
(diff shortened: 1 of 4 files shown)


src/raiden_libs/cli.py
```diff
@@ -100,10 +100,15 @@ def call_with_common_options_initialized(**params: Any) -> Callable:
             try:
                 setup_logging(params.pop("log_level"))
                 if not params["state_db"]:
-                    chain_id = str(params["web3"].net.version)
+                    # only RC has `chain_id`, MS and PFS have `web3` object
+                    chain_id = str(params.get("chain_id") or params["web3"].net.version)
```
karlb:
```suggestion
                    chain_id = str(params.pop("chain_id", params["web3"].net.version))
```

Then you should not need the extra pop below.

src/request_collector/cli.py
```diff
@@ -2,22 +2,49 @@
...
+from raiden.utils.cli import NetworkChoiceType
 from raiden_libs.cli import common_options
 
 log = structlog.get_logger(__name__)
 
 
 @click.command()
+@click.option(
+    "--chain-id",
+    type=NetworkChoiceType(["mainnet", "ropsten", "rinkeby", "goerli", "kovan", "<NETWORK_ID>"]),
```
karlb: Nice! Is there a way to reuse the helpstring, too?

src/raiden_libs/cli.py
palango: Nice, I expected this but didn't find it in any of the first google results.

src/request_collector/cli.py
palango: Don't think so, but we can make it a constant and use that in the future.

src/raiden_libs/cli.py
palango: So actually this doesn't work, as it eagerly evaluated and would complain about `web`  not being a valid key.

karlb: Well, then you could still do
```python
chain_id = str(params.pop("chain_id", None) or params["web3"].net.version)
```

palango: Yes, but as we need to delete it as well when the `state-db` is given, I think it's cleaner that way.

PR #403, opened 2019-06-17
Fix depoloyment

palango:
Before the services ran as root in the docker file. After the change to
a user, it cannot read the keyfile any more. Therefore the keyfile is
now put under `/data`.

### DIFF
--- a/deployment/docker-compose.yml
+++ b/deployment/docker-compose.yml
@@ -5,8 +5,8 @@ x-defaults: &defaults
   image: raiden-services
   env_file: services.env
   volumes:
-    - /data/state:/state
-    - /data/keystore:/keystore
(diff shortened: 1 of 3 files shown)


deployment/docker-compose.yml
```diff
@@ -6,7 +6,7 @@ x-defaults: &defaults
   env_file: services.env
   volumes:
     - /data/state:/state
-    - ~/keystore:/keystore
+    - /data/keystore:/keystore
```
ulope: One suggestion to make this more portable would be to only put a default here that works for local development but is overridable via a variable in `.env` ([example](https://github.com/raiden-network/raiden-transport/blob/b825292844c67327873aaa8231228774d8866015/docker-compose.yml#L16)).

PR #414, opened 2019-06-27
WIP: Information regarding raiden usage and contracts

agatsoh:
Gives some more information regarding raiden usage and deployed service contracts and associated tokens

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -45,6 +45,74 @@ the services follow these instructions:
 This installs the executables ``pathfinding-service``, ``monitoring-service``
 and ``request-collector`` into the virtual environment.
 
+Using MS and PFS in raiden
+--------------------------
+We need to install raiden from the latest development branch.
+
+.. code:: bash
+
+    git clone https://github.com/raiden-network/raiden.git
+    cd raiden
+    git checkout development
+    make install-dev
+
+Raiden client needs to be started with the raiden-services flags like
+
+* --enable-monitoring
+* --routing-mode [pfs|local|private]
+
+Doing ``raiden --help`` and going to Raiden Services options would give more
+information.
+
+Payments related to Services
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+The monitoring service registration, client onboarding(raiden node asking for
+a raiden service), monitoring service payment requires acquiring tokens in the
+token contracts deployed in various testnets and the mainnet. All deployment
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -45,6 +45,74 @@ the services follow these instructions:
 This installs the executables ``pathfinding-service``, ``monitoring-service``
 and ``request-collector`` into the virtual environment.
 
+Using MS and PFS in raiden
```
palango: While this is nice, doesn't this belong into the client docs?

docs/index.rst
```diff
@@ -45,6 +45,74 @@ the services follow these instructions:
...
+
+Raiden client needs to be started with the raiden-services flags like
+
+* --enable-monitoring
+* --routing-mode [pfs|local|private]
+
+Doing ``raiden --help`` and going to Raiden Services options would give more
+information.
+
+Payments related to Services
```
palango: What target group is this written for? I'd say users don't care about the contracts...

docs/index.rst
```diff
@@ -45,6 +45,74 @@ the services follow these instructions:
...
+.. _contracts: https://github.com/raiden-network/raiden-contracts
+  /blob/master/raiden_contracts/data_0.17.0/deployment_services_goerli.json
+
+
+The various *raiden-service* contracts and associated tokens can be deciphered
+from the json file containing the deployed data. Thus navigating to the
+``data_0.17.0`` folder and opening the ``deployment_services_goerli.json``
+shows us the following information.
+
++------------------+----------------------------------------------------------+
```
palango: Not sure if this really helpful, but if we leave it I'd recommend to make all addresses links to etherscan.

docs/index.rst
agatsoh: Yes this does belong to the client about to remove it.

docs/index.rst
agatsoh: Yes users exclusively of the MS and PFS services.

agatsoh:
Main use of writing this is to present to the users of MS and PFS as much information as we can provide so the users are really clear. Still a work in progress.

docs/index.rst
agatsoh: yes I will include the links to etherscan thanks for reminding

palango:
@agatsoh Are you still working on this?

palango:
@agatsoh Please reopen once you work on this again.

PR #415, opened 2019-06-28
Use gas costs in gas.json

pirapira:
instead of GAS_REQUIRED_FOR_* from raiden_contracts.constants
because that number is deprecated.

### DIFF
--- a/src/monitoring_service/cli.py
+++ b/src/monitoring_service/cli.py
@@ -19,7 +19,12 @@ log = structlog.get_logger(__name__)
 
 
 @blockchain_options(
-    contracts=[CONTRACT_TOKEN_NETWORK_REGISTRY, CONTRACT_USER_DEPOSIT, CONTRACT_MONITORING_SERVICE]
+    contracts=[
+        CONTRACT_TOKEN_NETWORK_REGISTRY,
+        CONTRACT_USER_DEPOSIT,
+        CONTRACT_MONITORING_SERVICE,
+    ],
+    with_gas_numbers=True,
 )
 @click.command()
 @click.option(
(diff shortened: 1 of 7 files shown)


src/monitoring_service/cli.py
```diff
@@ -32,14 +37,15 @@
     "--confirmations",
     default=DEFAULT_NUMBER_OF_BLOCK_CONFIRMATIONS,
     type=click.IntRange(min=0),
-    help="Number of block confirmations to wait for",
+    help="Number of block confirmations to wait for",  # pylint: disable=too-many-arguments
```
karlb: Where does that pylint exception come from? The code is the same as before and four args don't seem too many.

pirapira: I added an argument seven lines below, but `pylint` complained about this line.

pirapira: `raiden` client uses two `contracts_version`s: `PRODUCTION_CONTRACT_VERSION` and `DEVELOPMENT_CONTRACT_VERSION`  https://github.com/raiden-network/raiden/blob/93081972c11dc89c32ea0f2bab4f104baccf946e/raiden/settings.py#L46 .

Currently these two coincide, but sometimes they don't. Depending on the development/production flag given to `raiden` command, our codebase will be talking with different versions of contracts.

karlb: I don't think we need such a split for the services, but maybe I'm missing something. @palango, what do you think?

palango: Good question. So fare we're always on `develop` but that will change once we release. I guess we need to see how the client will handle this going forward

pirapira: What for now? I'm happy with either way.

palango: Let's just keep it as is, and we fix it once it's clear how the client will handle it.

karlb: Even if the client has some split between production and development, I don't see a need to support different contract versions inside a single release. Thus I suggest replacing this PR by https://github.com/raiden-network/raiden-services/pull/417.

karlb:
Closed in favor of https://github.com/raiden-network/raiden-services/pull/417.

PR #416, opened 2019-07-01
Cap update for unconfirmed channels

### LINKED ISSUE #250: Handle updates for unconfirmed channels
Currently a channel is only registered in the PFS once the containing block is confirmed (right now 8 blocks ~ 2mins).

It's likely that during this time a `PFSCapacityUpdate` message is sent. We need to find a way to save this information and apply it once the channel in confirmed.

karlb:
Closes #250.

### DIFF
--- a/src/pathfinding_service/service.py
+++ b/src/pathfinding_service/service.py
@@ -291,7 +291,7 @@ class PathfindingService(gevent.Greenlet):
 
         return token_network.handle_channel_fee_update(message)
 
-    def _validate_pfs_update(self, message: PFSCapacityUpdate) -> TokenNetwork:
+    def _validate_pfs_capacity_update(self, message: PFSCapacityUpdate) -> TokenNetwork:
         token_network_address = TokenNetworkAddress(
             message.canonical_identifier.token_network_address
         )
@@ -336,7 +336,7 @@ class PathfindingService(gevent.Greenlet):
         return token_network
 
     def on_capacity_update(self, message: PFSCapacityUpdate) -> List[ChannelView]:
-        token_network = self._validate_pfs_update(message)
+        token_network = self._validate_pfs_capacity_update(message)
         log.info("Received Capacity Update", message=message)
         self.database.upsert_capacity_update(message)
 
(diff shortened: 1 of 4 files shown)


src/pathfinding_service/service.py
```diff
@@ -39,6 +40,16 @@
 log = structlog.get_logger(__name__)
 
 
+class DeferMessage(Exception):
+    """Stop processing the message and handle it when the channel has been opened"""
+
+    deferred_message: DeferableMessage
+
+    def __init__(self, deferred_message: DeferableMessage):
```
palango: This should have a `-> None` annotation, strange that mypy doesn't complain.

src/pathfinding_service/service.py
```diff
@@ -242,7 +253,7 @@ def handle_channel_closed(self, event: ReceiveChannelClosedEvent) -> None:
     def handle_message(self, message: Message) -> None:
         try:
             if isinstance(message, PFSCapacityUpdate):
-                changed_cvs = self.on_pfs_update(message)
+                changed_cvs = self.on_capacity_update(message)
```
palango: Nice, I forgot to update that one!

src/pathfinding_service/service.py
```diff
@@ -293,13 +305,6 @@ def _validate_pfs_update(self, message: PFSCapacityUpdate) -> TokenNetwork:
         if token_network is None:
```
palango: I can't comment in the signature, but could you also update the naming there? `_validate_pfs_update` -> `_validate_pfs_capacity_update`?

src/pathfinding_service/service.py
karlb: Mypy does not require a `-> None` annotation for `__init__` unless it has not arguments. In that case an annotation is required to tell mypy whether this method is typed or not. If you like I can add it, but I don't see the benefit, personally.

src/pathfinding_service/service.py
karlb: Good point! Consistent renaming is harder than it sounds.

src/pathfinding_service/service.py
palango: Ah right. Fine for me! 👍

PR #417, opened 2019-07-02
Use gas costs in gas.json (simplified)

karlb:
This tries to be a simplified version of https://github.com/raiden-network/raiden-services/pull/415, by assuming that each service release is fixed to a specific contracts version.

### DIFF
--- a/src/monitoring_service/service.py
+++ b/src/monitoring_service/service.py
@@ -20,8 +20,8 @@ from raiden_contracts.constants import (
     CONTRACT_MONITORING_SERVICE,
     CONTRACT_TOKEN_NETWORK_REGISTRY,
     CONTRACT_USER_DEPOSIT,
-    GAS_REQUIRED_FOR_MS_MONITOR,
 )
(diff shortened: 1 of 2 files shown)


src/monitoring_service/service.py
```diff
@@ -33,7 +33,7 @@
 def check_gas_reserve(web3: Web3, private_key: str) -> None:
     """ Check periodically for gas reserve in the account """
     gas_price = web3.eth.gasPrice
-    gas_limit = GAS_REQUIRED_FOR_MS_MONITOR
+    gas_limit = gas_measurements()["MonitoringService.monitor"]
```
palango: Should we make the `gas_measurements()` a global like the contracts manager? Otherwise we might forget to update this call when adding the version?

karlb: If we don't specify the version it uses the latest code, which sounds just like what we want to do.

palango: Sure, we just need to remember to add the version in thswi place as well.

PR #435, opened 2019-07-09
Release 0.1

palango:
Also updated the release procedure here: https://github.com/raiden-network/team/wiki/%5BWIP%5D-Making-releases#services

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,3 +29,20 @@ install:
 
 install-dev:
 	pip install -r requirements-dev.txt
+
+dist:
+	python3 setup.py sdist bdist_wheel
+
+clean: clean-build clean-pyc
+
+clean-build:
+	rm -fr build/
+	rm -fr dist/
+	rm -fr .eggs/
+	find . -name '*.egg-info' -delete
+	find . -name '*.egg' -delete
+
+clean-pyc:
+	find . -name '*.pyc' -delete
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -29,3 +29,20 @@ install:
 
 install-dev:
 	pip install -r requirements-dev.txt
+
+package:
```
karlb: Is `package` a usual name for this target? Isn't something like `dist` very common?

Makefile
```diff
@@ -29,3 +29,20 @@ install:
...
+
+clean-build:
+	rm -fr build/
+	rm -fr dist/
+	rm -fr .eggs/
+	find . -name '*.egg-info' -exec rm -fr {} +
+	find . -name '*.egg' -exec rm -f {} +
+
+clean-pyc:
+	find . -name '*.pyc' -exec rm -f {} +
```
karlb:
```suggestion
	find . -name '*.pyc' -delete
```
If that works on OSX, too. That might be a GNU find extension.

Makefile
palango: Sounds good as well

Makefile
palango: Seems to work fine

palango:
> Great! Now, where is our release party?

Most important point! We should at least get some champagne on Thursday!

PR #436, opened 2019-07-09
Catching exceptions on_handle_message

### LINKED ISSUE #420: Request collector crashes on unsigned RequestMonitoring Messages
A server hosting the request collector can be crashed with different malicious messages, e.g.

- `request_monitoring.non_closing_signature = None` causes an assertion error,

- `request_monitoring.balance_proof.token_network_address = b'\0' * 2147483647` causes a memory error

once `handle_message` is called.

Dominik1999:
closes #420

### DIFF
--- a/src/raiden_libs/matrix.py
+++ b/src/raiden_libs/matrix.py
@@ -1,4 +1,5 @@
 import json
+import sys
 from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Union
 
 import gevent
@@ -59,6 +60,10 @@ def message_from_dict(data: Dict[str, Any]) -> Message:
 
 def deserialize_messages(data: str, peer_address: Address) -> List[SignedMessage]:
     messages: List[SignedMessage] = list()
+
+    if sys.getsizeof(data) >= 1000000:
+        log.warning("Received message that is too big")
+        return messages
 
     for line in data.splitlines():
         line = line.strip()
--- a/src/request_collector/server.py
+++ b/src/request_collector/server.py
@@ -49,10 +49,14 @@ class RequestCollector(gevent.Greenlet):
         self.matrix_listener.join()
 
     def handle_message(self, message: Message) -> None:
-        if isinstance(message, RequestMonitoring):
-            self.on_monitor_request(message)
-        else:
-            log.debug("Ignoring message", message=message)
+        try:
+            if isinstance(message, RequestMonitoring):
+                self.on_monitor_request(message)
+            else:
+                log.debug("Ignoring message", message=message)
+        # add more advanced exception catching
+        except AssertionError as ex:
+            log.error("Error while handling message", message=message, _exc=ex)
 
     def on_monitor_request(self, request_monitoring: RequestMonitoring) -> None:
         assert isinstance(request_monitoring, RequestMonitoring)
--- a/tests/monitoring/request_collector/test_server.py
+++ b/tests/monitoring/request_collector/test_server.py
@@ -1,4 +1,6 @@
 # pylint: disable=redefined-outer-name
+from unittest.mock import Mock, patch
+
 import pytest
 from eth_utils import encode_hex, to_checksum_address
 
@@ -93,3 +95,11 @@ def test_ignore_old_nonce(ms_database, build_request_monitoring, request_collect
 
     # update without higher nonce must succeed
     assert stored_mr_after_proccessing(amount=2, nonce=2).balance_hash != initial_hash
+
+
+def test_request_collector_doesnt_crash_with_invalid_messages(request_collector):
+    # We want to test that the request collector does not crash,
+    # in case an assertion on the MonitorRequest fails
+    # In theory, the collector crashed before, when on_monitor_request raised an AssertionError
+    with patch.object(request_collector, "on_monitor_request", side_effect=AssertionError):
+        request_collector.handle_message(Mock())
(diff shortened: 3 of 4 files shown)


Dominik1999:
Augusto had a good point, not to test impossible stuff. When the RequestMonitor Message cannot go without a reward_proof_signature, then we should not test it.

palango:
> When the RequestMonitor Message cannot go without a reward_proof_signature, then we should not test it.

How do you mean? Someone could still create a `RequestMonitor` with the signature missing and crash the service if we don't catch that properly.

Dominik1999:
> > When the RequestMonitor Message cannot go without a reward_proof_signature, then we should not test it.
> 
> How do you mean? Someone could still create a `RequestMonitor` with the signature missing and crash the service if we don't catch that properly.

Actually, when you use a Raiden Client as we support that, then you can not create that message. It isn't allowed when you try to instantiate such a message. Someone could still do so, I guess, by hacking it together?

palango:
> Someone could still do so, I guess, by hacking it together?

That's the case we try to handle here as I understand.

Dominik1999:
> > Someone could still do so, I guess, by hacking it together?
> 
> That's the case we try to handle here as I understand.

ok cool, because that is also what this PR is tackling. I just wanted to give the above feedback such that you know

src/raiden_libs/matrix.py
```diff
@@ -60,6 +61,10 @@ def message_from_dict(data: Dict[str, Any]) -> Message:
 def deserialize_messages(data: str, peer_address: Address) -> List[SignedMessage]:
     messages: List[SignedMessage] = list()
 
+    if sys.getsizeof(data) >= 1000000:
```
karlb: This does not check the size of a single message, but the total size of a set of messages. This makes it hard to tell if 1MB is a safe value to check against unless you depend on a specific message batching behavior in the transport..

src/request_collector/server.py
```diff
@@ -49,10 +49,14 @@ def stop(self) -> None:
...
-            self.on_monitor_request(message)
-        else:
-            log.debug("Ignoring message", message=message)
+        try:
+            if isinstance(message, RequestMonitoring):
+                self.on_monitor_request(message)
+            else:
+                log.debug("Ignoring message", message=message)
+        # add more advanced exception catching
+        except AssertionError as ex:
```
karlb: This is one of the rare cases where I would go as far as doing `except Exception`. I don't see a way to make sure that maliciously constructed messages can't raise some exception we didn't think of and catching all those is a simple way to keep the RC running to accept valid messages.

However, the `reward_proof_signature is None` problems won't cause an exception here, but in https://github.com/raiden-network/raiden-services/blob/master/src/raiden_libs/matrix.py#L71 where it is properly handled. Maybe I'm being a bit too defensive, here. What do you thing @palango?

tests/monitoring/request_collector/test_server.py
```diff
@@ -93,3 +95,11 @@ def stored_mr_after_proccessing(amount, nonce):
 
     # update without higher nonce must succeed
     assert stored_mr_after_proccessing(amount=2, nonce=2).balance_hash != initial_hash
+
+
+def test_request_collector_doesnt_crash_with_invalid_messages(request_collector):
+    # We want to test that the request collector does not crash,
+    # in case an assertion on the MonitorRequest fails
+    # In theory, the collector crashed before, when on_monitor_request raised an AssertionError
```
karlb: >  In theory, the collector crashed before, when on_monitor_request raised an AssertionError
	
Before what? The comment will stay in the code, so if you're referring to this PR, this won't be understood after this PR is merged.

src/request_collector/server.py
palango: I agree, let's catch everything that isn't handled before as a last resort kind of thing.

For the `reward_proof_signature is None`: I'd like to have a test at least to be sure in the future that it is handled.

src/raiden_libs/matrix.py
Dominik1999: But this might be the attack vector, right? Or should we analyse on a message level?

tests/monitoring/request_collector/test_server.py
Dominik1999: ah that is a valid point hehe changed it

src/raiden_libs/matrix.py
Dominik1999: changed it to an per line analysis

src/request_collector/server.py
```diff
@@ -49,10 +49,14 @@ def stop(self) -> None:
...
-        if isinstance(message, RequestMonitoring):
-            self.on_monitor_request(message)
-        else:
-            log.debug("Ignoring message", message=message)
+        try:
+            if isinstance(message, RequestMonitoring):
+                self.on_monitor_request(message)
+            else:
+                log.debug("Ignoring message", message=message)
+        # add more advanced exception catching
```
karlb: What "more advanced exception catching" do you expect, here?

Dominik1999: Maybe something more differentiated with better logs to distinguish. does that make sense?

karlb: I see this except as a safety measure in case we fail to properly handle some case at a different place. I don't see what I would add here, since other errors are usually better handled closer to where they are thrown.

Dominik1999: ok then I remove the comment

Dominik1999:
can we merge that?

karlb:
> can we merge that?

Yes, as soon as it is rebased to the current master.

Issue #418, opened 2019-07-02
Monitoring Service database vulnerable to timing based Monitoring Request injection


palango:
If an attacker intercepts a monitoring request before it reaches the request collector, it can alter the attached data and replay the signature to effectively prevent channel monitoring for the victim node, potentially causing the node loosing funds.

Dominik1999:
but can we do something about that?

err508:
Yes we can! 
The vulnerability essentially comes from the two different signatures of a monitoring request, where only the `non_closing_signature` is relevant for the database and the reward_proof_signature only wraps: 
```
                self.channel_identifier,
                self.reward_amount,
                self.token_network_address,
                self.chain_id,
                self.nonce,
```
A concise solution would be to wrap all required data e.g. the whole balance proof + reward proof data, than the `non_closing_signer` would become obsolete, as any altered MonitorRequest would be stored, but if the original one arrives it would be stored as well. If the channel is than closed, only the original one would be used as the recovered addresses of the tempered version would not match the channel participants. 

A fix would be to require `non_closing_signer == reward_proof_signer` in the request collector before storing anything to the db. This fixes the particular issue and is helpful anyway, but there might be more subtle ways of exploiting the independence of the two signatures.

karlb:
The `non_closing_signer == reward_proof_signer` will solve this specific attack, but I also think that the independence of those two parts of the monitoring request might open up a whole class of attacks. It's also a lot harder to reason about the security of that message when that is the case.
I just suggested changing the reward proof to consist of `(non_closing_signature, reward_amount, monitoring_service_contract_address)` to @pirapira and he agreed that this is a good approach. I'll prepare PRs for the contract and the client.

PR #442, opened 2019-07-11
Prevent acceptance of bad reward proof

karlb:
By checking that both MR signatures are done by the same key, attackers are prevented from adding a bad reward proof to an otherwise valid MR. This prevents https://github.com/raiden-network/raiden-services/issues/418 immediately.

After https://github.com/raiden-network/raiden-contracts/issues/1117 is done, this will be a much more useful check because it will not only fail when one of the proofs is tampered with, but also when they were not intended to be used in the same MR.

Closes https://github.com/raiden-network/raiden-services/issues/418.

### DIFF
--- a/src/request_collector/server.py
+++ b/src/request_collector/server.py
@@ -84,6 +84,8 @@ class RequestCollector(gevent.Greenlet):
         if monitor_request.chain_id != self.chain_id:
             log.debug("Bad chain_id", monitor_request=monitor_request, expected=self.chain_id)
             return
+        if monitor_request.non_closing_signer != monitor_request.reward_proof_signer:
+            log.debug("The two MR signatures don't match", monitor_request=monitor_request)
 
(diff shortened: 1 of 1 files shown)


src/request_collector/server.py
```diff
@@ -84,6 +84,8 @@ def on_monitor_request(self, request_monitoring: RequestMonitoring) -> None:
         if monitor_request.chain_id != self.chain_id:
             log.debug("Bad chain_id", monitor_request=monitor_request, expected=self.chain_id)
             return
+        if monitor_request.non_closing_signer != monitor_request.reward_proof_signer:
+            log.debug("The two MR signatures don't match", monitor_request=monitor_request)
```
err508: IMO there's a return missing here ;-)

karlb: You are so right!

PR #443, opened 2019-07-11
Return if signature check fails

err508:


palango:
@err508 Can you rebase to trigger the build again?

PR #444, opened 2019-07-11
Add more logging about why paths are invalid

palango:
Adds more debug information


palango:
Yeah, I'm also scared of that, but for now I prefer to have information available.

PR #445, opened 2019-07-11
Recursivly prettify log messages

palango:
Now also nested structures look nice

### DIFF
--- a/src/raiden_libs/logging.py
+++ b/src/raiden_libs/logging.py
@@ -36,23 +36,30 @@ def setup_logging(log_level: str) -> None:
     )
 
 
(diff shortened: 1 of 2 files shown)


src/raiden_libs/logging.py
```diff
@@ -36,23 +36,30 @@ def setup_logging(log_level: str) -> None:
...
+        if len(value) == 20:
+            return to_checksum_address(value)
+
+        return to_hex(value)
+    return value
+
+
+def apply_recursive(value: Any) -> Any:
+    if isinstance(value, list):
+        return [apply_recursive(x) for x in value]
```
karlb: We should do the same for tuples, they are also quite common.

palango: Good, point. Added it.

PR #448, opened 2019-07-12
Add register-service script for testnet ServiceRegistry

### LINKED ISSUE #447: Register service at ServiceRegistry
## Description

register my services in the ServiceRegistry

**So that I can**

be fund by clients.

## Acceptance Criteria

- [ ] Have a script that registers the services in the ServiceRegistry.

palango:
Fixes #447

I need to write tests for this too....

### DIFF
--- a/src/raiden_libs/cli.py
+++ b/src/raiden_libs/cli.py
@@ -21,6 +21,7 @@ from raiden.utils.typing import Address, BlockNumber, ChainID
 from raiden_contracts.constants import (
     CONTRACT_MONITORING_SERVICE,
     CONTRACT_ONE_TO_N,
+    CONTRACT_SERVICE_REGISTRY,
     CONTRACT_TOKEN_NETWORK_REGISTRY,
     CONTRACT_USER_DEPOSIT,
     CONTRACTS_VERSION,
@@ -132,10 +133,11 @@ def blockchain_options(contracts: List[str]) -> Callable:
     ]
 
     arg_for_contract = {
-        CONTRACT_TOKEN_NETWORK_REGISTRY: "token-network-registry",
+        CONTRACT_TOKEN_NETWORK_REGISTRY: "token-network-registry-contract",
         CONTRACT_USER_DEPOSIT: "user-deposit-contract",
         CONTRACT_MONITORING_SERVICE: "monitor-contract",
         CONTRACT_ONE_TO_N: "one-to-n-contract",
+        CONTRACT_SERVICE_REGISTRY: "seervice-registry-contract",
     }
 
     param_for_contract: Dict[str, str] = {}
--- a/src/raiden_libs/cli.py
+++ b/src/raiden_libs/cli.py
@@ -133,11 +133,11 @@ def blockchain_options(contracts: List[str]) -> Callable:
     ]
 
     arg_for_contract = {
-        CONTRACT_TOKEN_NETWORK_REGISTRY: "token-network-registry-contract",
+        CONTRACT_TOKEN_NETWORK_REGISTRY: "token-network-registry",
         CONTRACT_USER_DEPOSIT: "user-deposit-contract",
         CONTRACT_MONITORING_SERVICE: "monitor-contract",
         CONTRACT_ONE_TO_N: "one-to-n-contract",
-        CONTRACT_SERVICE_REGISTRY: "seervice-registry-contract",
+        CONTRACT_SERVICE_REGISTRY: "service-registry-contract",
     }
 
     param_for_contract: Dict[str, str] = {}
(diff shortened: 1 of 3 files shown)


src/raiden_libs/cli.py
```diff
@@ -132,10 +133,11 @@ def blockchain_options(contracts: List[str]) -> Callable:
     ]
 
     arg_for_contract = {
-        CONTRACT_TOKEN_NETWORK_REGISTRY: "token-network-registry",
+        CONTRACT_TOKEN_NETWORK_REGISTRY: "token-network-registry-contract",
```
karlb: This needs to be updated in the docs.

src/raiden_libs/register_service.py
```diff
@@ -0,0 +1,142 @@
...
+        service_address=service_address,
+        current_deposit=current_deposit,
+        current_url=current_url,
+    )
+
+    # Register if not yet done
+    if current_deposit <= 0:
+        log.info("Address not registered in ServiceRegistry")
+
+        custom_token_address = contracts[CONTRACT_USER_DEPOSIT].functions.token().call()
```
karlb: Custom token? Isn't this assumed to be RDN? For the test contract deploys it isn't, but `custom_token` still sounds very vague.

src/raiden_libs/register_service.py
```diff
@@ -0,0 +1,142 @@
...
+            abi=CONTRACT_MANAGER.get_contract_abi(CONTRACT_CUSTOM_TOKEN),
+        )
+
+        # Check current token balance
+        account_balance = custom_token_contract.functions.balanceOf(service_address).call()
+        log.info("Current account balance", balance=account_balance, desired_deposit=deposit)
+
+        # mint token if necessary
+        if account_balance < deposit:
+            run_task(
```
karlb: `run_task` sounds like concurrent stuff is happening. Maybe just call it something like `checked_transact`?

src/raiden_libs/register_service.py
```diff
@@ -0,0 +1,142 @@
...
+@common_options("register_service")
+def main(
+    private_key: str,
+    state_db: str,
+    web3: Web3,
+    contracts: Dict[str, Contract],
+    start_block: BlockNumber,
+    deposit: TokenAmount,
+    service_url: str,
+) -> None:
```
karlb: If we add a docstring, it should show up in the cli. We could inform the user that used key define be both the payer as well as the service address.

src/raiden_libs/register_service.py
```diff
@@ -0,0 +1,142 @@
...
+                function_call=custom_token_contract.functions.mint(deposit),
+                task_name="Minting new Test RDN tokens",
+            )
+
+            account_balance = custom_token_contract.functions.balanceOf(service_address).call()
+            log.info("Updated account balance", balance=account_balance, desired_deposit=deposit)
+
+        current_allowance = custom_token_contract.functions.allowance(
+            service_address, service_registry_contract.address
+        ).call()
```
karlb: Do we need the allowance check? Usually there should be now allowance left, and if there is, setting it again won't cause more problems than the wasted gas.

src/raiden_libs/cli.py
palango: Do you think the change is worthwhile or should I undo it? It just didn't seem consistent

karlb: When I added it, I intentionally kept the name to keep it the same as for the client. But now neither have to be entered explicitly, usually. I don't mind either way.

src/raiden_libs/register_service.py
palango: You mean to get rid of the check and call `allow` unconditionally?

karlb: `approve` unconditionally, yes. From my understanding, that's what will happen in > 99% of the cases, anyway.

src/raiden_libs/cli.py
palango: Let's do it in a different PR

palango:
@karlb Please have another look.

src/raiden_libs/register_service.py
```diff
@@ -0,0 +1,145 @@
...
+    web3: Web3,
+    contracts: Dict[str, Contract],
+    start_block: BlockNumber,  # pylint: disable=unused-argument
+    deposit: TokenAmount,
+    service_url: str,
+) -> None:
+    """
+    Registers the address of a service deployment with the `ServiceRegistry`.
+
+    It also set or updates the URL of the services deployment.
```
karlb:
```suggestion
    It also sets or updates the URL of the services deployment.
```

Maybe add one more sentence to say that the `private_key`s address will be registered. Or is that obvious?

Issue #419, opened 2019-07-02
Request collector stores arbitrary Monitoring Requests


palango:
An attacker can send an arbitrary amount of MonitoringRequests to the request collector, which will - given matching chain id - store it to the shared db. This allows cheap denial of service attacks against the shared database, the request collector and the monitoring service.

karlb:
We must accept MonitoringRequests before those channels are confirmed. This makes it hard to just discard MRs in case of a DOS attack without complicating the request collector a lot. I suggest a combination of
* rate limiting (will be useful for many other attacks, too)
* regularly checking the DB for MRs that don't have a matching channel and are older than a certain time.

palango:
I agree with these two mitigations, @err508 what do you think?

Regarding rate limiting. Is it enough to rate limit in the matrix client when receiving messages, or do we have to put that inside matrix somehow? @ulope @err508 Maybe you can give input on this!

ulope:
Well that's a bit of a circular problem.

Matrix has server side rate limiting support.
But since we don't want transfers to get throttled we have tuned those to rather high values.
Unfortunately this is a global per-server setting.

But how would an attacker generate arbitrary amounts of requests? 
AFAIU only channels that are open on-chain should be allowed to have monitoring requests posted for. 

Based on that it should be possible to implement some per-channel-peer rate limiting yourselves, shouldn't it?

err508:
I'm not really familiar with the client code base, but before a transfer happens there seems to be done [this](https://github.com/raiden-network/raiden/blob/60afd1afecf4cb4d6d98b5d99a7437b99e84f1b2/raiden/network/proxies/payment_channel.py#L161) check in the channel which relies on [this](
https://github.com/raiden-network/raiden/blob/60afd1afecf4cb4d6d98b5d99a7437b99e84f1b2/raiden/network/proxies/token_network.py#L626) check in the token_network view. 
@karlb could you elaborate on cases where you think the request_collector would have to accept monitoring requests from properly configured clients, where the channel_opened event might not be visible yet? 
If it is not to many cases, where valid monitoring requests arrive before the ChannelOpened event is visible, I'd just perform the same check before storing stuff to the MonitoringService db.

err508:
Also regarding rate limiting or other checks performed in matrix: I think it's good practice to think of a module as an autonomous security perimeter. 
MatrixServers might be deployed by some party and I think you don't want to be in the situation where you try to find their phone number in the middle of the night because some other module gets spammed.

[...]

PR #449, opened 2019-07-12
Purge old monitor requests from database

karlb:
Relates to https://github.com/raiden-network/raiden-services/issues/419.


karlb:
The failure in CI could be due to different sqlite versions. I'm looking into it.

src/monitoring_service/schema.sql
```diff
@@ -48,10 +48,20 @@ CREATE TABLE monitor_request (
     reward_proof_signature  CHAR(132)   NOT NULL,
 
     non_closing_signer      CHAR(42)    NOT NULL,
+
+    -- These two columns are just for handling MRs before we have confirmed
+    -- that a matching channel exists.
+    -- * If `waiting_for_channel` is false, we've already checked that such a
+    --   channel exists and everything is ok.
+    -- * If `saved_at` is sufficiently recent, a missing channel is acceptable.
+    -- * If `saved_at` is old, we will delete the MR if not matching channel is
```
palango: s/not/no

src/monitoring_service/schema.sql
```diff
@@ -48,10 +48,20 @@ CREATE TABLE monitor_request (
...
+    --   found.
+    saved_at                TIMESTAMP   NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
palango: Maybe add a comment that `CURRENT_TIMESTAMP` is GMT by default.

src/monitoring_service/service.py
```diff
@@ -186,3 +188,32 @@ def _process_new_blocks(self, last_block: BlockNumber) -> None:
...
+                WHERE waiting_for_channel
+                  AND EXISTS (
+                    SELECT 1
+                    FROM channel
+                    WHERE channel.identifier = monitor_request.channel_identifier
+                      AND channel.token_network_address = monitor_request.token_network_address
+                  )
+            """
+            )
+            before_this_is_old = datetime.utcnow() - timedelta(minutes=15)
```
palango: In other places I needed to use `datetime.now(timezone.utc)` to get proper timezone enabled datetimes, should we do that here as well?

palango: I'd make this a constant, so it is more discoverable.

tests/monitoring/fixtures/factories.py
```diff
@@ -0,0 +1,44 @@
...
+            chain_id=chain_id,
+            nonce=nonce,
+            additional_hash="",
+            balance_hash=encode_hex(bytes([amount])),
+            priv_key=get_random_privkey(),
+        )
+        request_monitoring = balance_proof.get_request_monitoring(
+            privkey=non_closing_privkey,
+            reward_amount=TokenAmount(55),
+            monitoring_service_contract_address=Address(bytes([11] * 20)),
```
palango: Maybe use `TEST_MSC_ADDRESS` here

src/monitoring_service/service.py
karlb: Both are without timezone, so we should be fine. All the trouble we had was just related to marshmallow adding a time zone, which caused problems for equality checks and signatures, right? I don't remember the time actually being off in any case.

palango: That's true, my question was more if we should care to use this consistently or not.

PR #450, opened 2019-07-15
Fix recently introduced reward signature problems

karlb:
Caused by e9110e2cf5ccb91b2e3e84e99164104c03478fa5 and
693c9701ed41b2a471df53af5dea311aee6335cd.

### DIFF
--- a/src/monitoring_service/database.py
+++ b/src/monitoring_service/database.py
@@ -64,9 +64,11 @@ class SharedDatabase(BaseDatabase):
     ) -> Optional[MonitorRequest]:
         row = self.conn.execute(
             """
-                SELECT *,
-                    (SELECT chain_id FROM blockchain) As chain_id
-                FROM monitor_request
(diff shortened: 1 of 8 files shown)


src/monitoring_service/database.py
```diff
@@ -64,9 +64,11 @@ def get_monitor_request(
     ) -> Optional[MonitorRequest]:
         row = self.conn.execute(
             """
-                SELECT *,
-                    (SELECT chain_id FROM blockchain) As chain_id
-                FROM monitor_request
+                SELECT monitor_request.*,
+                    chain_id,
```
palango: Should we reference and/or check that this is the same as blockchain.chain_id somehow. It should be checked during message receival, but...

karlb: It *is* `blockchain.chain_id`, we don't store that separately in `monitor_request`. Maybe I should prefix the column names with the table name to make that clearer.

PR #451, opened 2019-07-15
Fix logging the event/message name

palango:
Minor fix for logging of events and messages.

### DIFF
--- a/src/raiden_libs/logging.py
+++ b/src/raiden_libs/logging.py
@@ -57,8 +57,9 @@ def apply_recursive(value: Any) -> Any:
 def format_to_hex(_logger: Any, _log_method: Any, event_dict: Dict) -> Dict[str, Any]:
     for key, val in event_dict.items():
         if isinstance(val, (Event, Message)):
+            name = val.__class__.__name__
             val = asdict(val)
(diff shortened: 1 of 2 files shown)


src/raiden_libs/logging.py
```diff
@@ -57,8 +57,9 @@ def apply_recursive(value: Any) -> Any:
 def format_to_hex(_logger: Any, _log_method: Any, event_dict: Dict) -> Dict[str, Any]:
     for key, val in event_dict.items():
         if isinstance(val, (Event, Message)):
+            name = val.__class__.__name__
             val = asdict(val)
-            val["event_name"] = val.__class__.__name__
+            val["type_name"] = name
```
karlb: I liked the compact version
```python
            val["type_name"] = val.__class__.__name__
```
better. It already has a name (`type_name`), so the intermediate name does not add any value.

But this is really a pedantic nitpick.

palango: The problem was that `asdict(val)` changed to the type to `dict`. So we got no event names, but alsways `"event_name": "dict"`

karlb: OK, now I get it.

Issue #452, opened 2019-07-15
Add test for register-service script


palango:
Add tests for the new `register-service`  script.

karlb:
Done in https://github.com/raiden-network/raiden-services/pull/890. It does not cover error cases, but it exercises the main functionality of registering, extending and withdrawing.

PR #455, opened 2019-07-15
Update `register_service` script and add some tests

### LINKED ISSUE #491: Update register-service script for new contracts

palango:
There should be an CLI test as well, but this requires so much mocking that I didn't dare to do it now.

Part of #452 
Fixes #491

### DIFF
--- a/src/raiden_libs/register_service.py
+++ b/src/raiden_libs/register_service.py
@@ -3,11 +3,12 @@ from typing import Dict
 
 import click
 import structlog
+from eth_utils import to_checksum_address, to_hex
 from web3 import Web3
 from web3.contract import Contract, ContractFunction
 from web3.middleware import construct_sign_and_send_raw_middleware
 
-from raiden.utils.typing import Address, BlockNumber, TokenAmount
+from raiden.utils.typing import Address, BlockNumber
 from raiden_contracts.constants import (
     CONTRACT_CUSTOM_TOKEN,
     CONTRACT_SERVICE_REGISTRY,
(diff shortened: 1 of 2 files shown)


src/raiden_libs/register_service.py
```diff
@@ -20,6 +20,26 @@
...
+    web3: Web3, service_address: Address, function_call: ContractFunction, task_name: str
+) -> None:
+    log.info(f"Starting: {task_name}")
+    transaction_hash = function_call.transact({"from": service_address})
+    transaction_receipt = web3.eth.waitForTransactionReceipt(transaction_hash)
+    was_successful = transaction_receipt["status"] == 1
+
+    if not was_successful:
+        log.error(
+            f"Failed: {task_name}\nPlease check that you account is funded.",
```
Dominik1999: you**r** account is funded

But maybe even -> "Please check that you have enough funds" ?

src/raiden_libs/register_service.py
```diff
@@ -65,48 +82,86 @@ def main(
...
-            log.info("Updated account balance", balance=account_balance, desired_deposit=deposit)
 
-        # Approve token transfer
+def deposit_to_registry(
+    web3: Web3,
+    service_registry_contract: Contract,
+    user_deposit_contract: Contract,
+    service_address: Address,
+) -> None:
+    log.info("Address not registered in ServiceRegistry")
```
Dominik1999: how do you know that? Do you check for duplicates somewhere?

src/raiden_libs/register_service.py
```diff
@@ -65,48 +82,86 @@ def main(
 
     # Register if not yet done
     if current_deposit <= 0:
```
Dominik1999: is that the stake of the service? then maybe call it stake?

palango: Good point, I wonder how that actually worked. There is no field `deposits` any more?!

src/raiden_libs/register_service.py
palango: `deposit_to_registry` is only called if the check you commented on above is false.

PR #456, opened 2019-07-15
Rate limit each sender based on message size

karlb:
I wanted to use a pre-built rate limiter, but none really fit well for
this use case.

Should be a sufficient mitigation to close
https://github.com/raiden-network/raiden-services/issues/419.


src/monitoring_service/constants.py
```diff
@@ -6,3 +6,6 @@
 DEFAULT_GAS_CHECK_BLOCKS: int = 100
 DEFAULT_PAYMENT_RISK_FAKTOR: int = 2
 KEEP_MRS_WITHOUT_CHANNEL: timedelta = timedelta(minutes=15)
+
+MATRIX_RATE_LIMIT_ALLOWED_BYTES = 5_000_000
```
palango: Can you elaborate a bit how you came up with these values? It means we can receive 1M bytes per minute, how many messages are that roughly?

tests/libs/test_matrix.py
```diff
@@ -134,7 +141,25 @@ def test_deserialize_messages_that_is_too_big(request_monitoring_message, capsys
...
 
-    assert "Received message that is too big" in captured.out
+    assert "Sender is rate limited" in captured.out
+
+
+def test_rate_limiter():
+    limiter = RateLimiter(allowed_bytes=100, reset_interval=timedelta(seconds=1))
+    sender = Address(b"1" * 20)
+    for _ in range(50):
+        assert limiter.check_and_count(sender, 2)
```
palango: adding kwargs here would make it easier to understand that 2 is the number of bytes, not the value IMO

src/monitoring_service/constants.py
karlb: A message like
```json
{
    "type": "LockedTransfer",
    "chain_id": 337,
    "message_identifier": 123456,
    "payment_identifier": 1,
    "nonce": 1,
    "token_network_address": "0xe82ae5475589b828D3644e1B56546F93cD27d1a4",
    "token": "0xc778417E063141139Fce010982780140Aa0cD5Ab",
    "channel_identifier": 1338,
    "transferred_amount": 0,
    "locked_amount": 10,
    "recipient": "0x2A915FDA69746F515b46C520eD511401d5CCD5e2",
    "locksroot": "0x7b3cb8717939d1fc4054b9ef46978ba3780556aa7b1482c65585f65a3a97f644",
    "lock": {
        "type": "Lock",
        "amount": 10,
        "expiration": 8155044,
        "secrethash": "0x59cad5948673622c1d64e2322488bf01619f7ff45789741b15a9f782ce9290a8"
    },
    "target": "0x811957b07304d335B271feeBF46754696694b09e",
    "initiator": "0x540B51eDc5900B8012091cc7c83caf2cb243aa86",
    "fee": 0,
    "signature": "0x4c9baebbaefa9a5ae6589c7764204303edc27c6186bc9ad94267fd1e9be2d3f74ddbbdb6fd0c332d6bb9943a54aa756ab3f787852a7b1ae977e5316b1b0377891c"
}
```

Is roughly 1kb. Having 1000/min = 17/sec will be hard to achieve outside of benchmarks for now. To have some safety margin for bursts of messages, this is only enforced as an average over 5 minutes. If have no problems increasing the allowance if you think this is too strict.

palango: Thanks for the explanation. It seems fine, maybe even to high for the beginning.

Issue #80, opened 2019-02-05
Provide release version PFS deployments


palango:
## Problem Definition

For testing purposes we want to provide services of the latest released version as well as current master.

nlsdfnbch:
In order to supply a `latest` and `master` version, you will have to supply a Dockerfile, which will be built via Dockerhub.

I would suggest using build/run args to keep the image customizable, otherwise you'd have to build several different images, one for each chain/config combination.

Once you have done so, I can set up hooks on docker hub to build these files.

`latest` is typically built on the latest commit on `dev`, whereas `master` is typically the latest pushed `tag`. If that does not suit you, let me know how you'd like it to be configured.

palango:
There is a dockerfile in the repo: https://github.com/raiden-network/raiden-services/blob/master/Dockerfile

If necessary please update that as well. I guess for now we'd only need to add a build arg for the version of the services to install. Everything else could stay a run arg in my opinion.

nlsdfnbch:
That will require a bit more than a build arg, since that does not even define an Entry point/cmd step.

Please specify the args required to run the service, and the args you'd like to bake-in/have customizable.

Also, will different images be required for each service? If not, I will have to accept a build_arg to set the environment vars (I believe you do set that via env vars, iirc?)

palango:
The docker-compose file is here: https://github.com/raiden-network/raiden-services/blob/master/deployment/docker-compose.yml

We don't have an entry point because there are multiple services in the same image.

To keep in mind: the installation guide/package will add further requirements (like running the registration script). See https://github.com/raiden-network/raiden-services/pull/438 for that

karlb:
This corresponds to https://github.com/raiden-network/team/issues/591.

PR #457, opened 2019-07-16
[FEAT-#80] Update docker setup to use docker registry

nlsdfnbch:
The docker-compose files provided now use the hub.docker.com
registry to pull built images.

This replaces the builder's build functionality with an
update functionality - images are no longer built
directly from the repo - instead, we let hub.docker.com
handle the build process. Images are updated using
`docker-compose pull && docker-compose restart` to
download the newest images for the configured tag.

Additionally, a docker-compose.latest.yml has been added
as a drop-in replacement for docker-compose.override.yml,
which is a more explicit name for the file imo.

Doc strings have been added to the compose yamls.

Exception handling was slightly refacored in the builder app.

### DIFF
--- a/deployment/builder/app/main.py
+++ b/deployment/builder/app/main.py
@@ -36,8 +36,12 @@ REPOS = {
 app = Flask(__name__)
 
 
-class BuildError(Exception):
+class ImageUpdateError(Exception):
     """Couldn't build/update our images."""
+
+
+def print_to_stderr(s):
+    print(s, file=sys.stderr)
 
 
 @app.route("/", methods=["get", "post"])
@@ -49,10 +53,10 @@ def main():
     if branch_config and branch in branch_config:
         try:
             res = update(branch, branch_config["names"], **branch_config[branch])
-        except BuildError:
-            print("Error building", file=sys.stderr)
+        except ImageUpdateError:
+            print_to_stderr("Error updating local images via docker registry!")
         except Exception as e:
-            print(f"Fatal Error while updating images: Unhandled exception encountered: {e}")
+            print_to_stderr(f"Fatal Error while updating images: Unhandled exception encountered: {e}")
             raise
         else:
             pprint(
@@ -69,28 +73,24 @@ def main():
             return "OK"
 
 
-def _print(s):
-    print(s, file=sys.stderr)
-
-
 def update(branch, container_names, source, deployment, **kw):
-    _print(f"Changing working directory to {deployment}")
+    print_to_stderr(f"Changing working directory to {deployment}")
     try:
         os.chdir(deployment)
     except FileNotFoundError as e:
-        _print(f"Could not change to directory {deployment} - Not found!")
-        raise BuildError from e
+        print_to_stderr(f"Could not change to directory {deployment} - Not found!")
+        raise ImageUpdateError from e
 
-    _print("Pulling containers containers: docker pull")
+    print_to_stderr("Pulling containers containers: docker pull")
     try:
         subprocess.run(["docker-compose", "pull"], check=True)
     except subprocess.SubprocessError as e:
-        _print(f"Pulling new images from docker registry failed: {e}")
-        raise BuildError from e
+        print_to_stderr(f"Pulling new images from docker registry failed: {e}")
+        raise ImageUpdateError from e
 
-    _print(f"Restarting containers: docker restart: {container_names}")
+    print_to_stderr(f"Restarting containers: docker restart: {container_names}")
     try:
         subprocess.run(["docker-compose", "restart"] + container_names, check=True)
     except subprocess.SubprocessError as e:
-        _print(f"Restarting containers {container_names} failed: {e}")
-        raise BuildError from e
+        print_to_stderr(f"Restarting containers {container_names} failed: {e}")
+        raise ImageUpdateError from e
--- a/deployment/builder/app/main.py
+++ b/deployment/builder/app/main.py
@@ -36,6 +36,10 @@ REPOS = {
 app = Flask(__name__)
 
 
+class BuildError(Exception):
+    """Couldn't build/update our images."""
+
+
 @app.route("/", methods=["get", "post"])
 def main():
     data = request.json or {}
@@ -43,8 +47,14 @@ def main():
     branch = data.get("ref", "").replace("refs/heads/", "")
     branch_config = REPOS.get(repo)
     if branch_config and branch in branch_config:
-        res = build(branch, branch_config["names"], **branch_config[branch])
-        if res:
+        try:
+            res = update(branch, branch_config["names"], **branch_config[branch])
+        except BuildError:
+            print("Error building", file=sys.stderr)
+        except Exception as e:
+            print(f"Fatal Error while updating images: Unhandled exception encountered: {e}")
+            raise
+        else:
             pprint(
                 {
                     "repo": repo,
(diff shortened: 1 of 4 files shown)


nlsdfnbch:
These changes take care of providing `latest` and `stable` images from the `master` branch. They're built directly on hub.docker.com and are available under the `raidennetwork/raiden-services` namespace.

deployment/builder/app/main.py
```diff
@@ -36,15 +36,25 @@
...
     repo = data.get("repository", {}).get("full_name", "")
     branch = data.get("ref", "").replace("refs/heads/", "")
     branch_config = REPOS.get(repo)
     if branch_config and branch in branch_config:
-        res = build(branch, branch_config["names"], **branch_config[branch])
-        if res:
+        try:
+            res = update(branch, branch_config["names"], **branch_config[branch])
+        except BuildError:
+            print("Error building", file=sys.stderr)
```
karlb: Is `BuildError` and `"Error building"` the right description? Since it's not building locally anymore, `UpdateError` might be more descriptive.

deployment/builder/app/main.py
```diff
@@ -36,15 +36,25 @@
...
+        except Exception as e:
+            print(f"Fatal Error while updating images: Unhandled exception encountered: {e}")
```
karlb: Shouldn't this go to stderr?

deployment/builder/app/main.py
karlb: There's `_print` for printing to stderr. That's not the best function name, but we should use the same way to print to stderr in the whole file.

deployment/builder/app/main.py
```diff
@@ -36,15 +36,25 @@
 app = Flask(__name__)
 
 
+class BuildError(Exception):
```
palango: As the function is now called `update`, should this be an `UpdateError`?

deployment/docker-compose.latest.yml
```diff
@@ -0,0 +1,89 @@
+# Development Setup.
+# Uses the latest commit of the `master` branch to run each service.
+# Start individual services using the following command::
```
karlb: > Additionally, a docker-compose.latest.yml has been added
as a drop-in replacement for docker-compose.override.yml,
which is a more explicit name for the file imo.

The `docker-compose.override.yml` allowed to run the local code including all changes that are not committed and pushed, yet. AFAICS, this does not work with `docker-compose.latest.yml`. That is only relevant for people developing the services, not for anyone who just wants to run the latest version from master.

It's just a nice to have, so I won't insist on having that functionality and it also does not work out of the box, as the keystore is not in the repo. But reusing the docker setup locally looked like a nice way to allow people to run services without digging out all the required CLI params.

If this is indeed meant to replace the `docker-compose.override.yml`, why didn't you remove it in the PR?

nlsdfnbch: >If this is indeed meant to replace the docker-compose.override.yml, why didn't you remove it in the PR?

Because I'm not sure if it isn't used for something else (which you confirmed just now). I assumed it was used to deploy the latest dev version to a server, not run them locally. 

If this is something you guys need/use, I'll gladly just leave it in there. However, if it does not "run out of the box", we should probably adapt it so it does, if that's possible.

nlsdfnbch:
> Thanks @nlsdfnbch ! This is a nice improvement.
> 
> Two questions:
> 
>     * What happens to `docker-compose.override.yml`, shouldn't that be deleted?

See my reply to karl's comment

> 
>     * Now there is the options to run `latest`/`stable`, but they don't run both simultaneously. Can you fix that as well?

what do you mean they don't run simultaneously? :thinking:

palango:
> what do you mean they don't run simultaneously? 🤔

There need to be versions of the stable and latest versions running on services-dev.raiden.network.

So when we currently have `pfs-georli.services-dev.raiden.network` (I assume this is latest) we also need to add `pfs-georli-release.services-dev.raiden.network` (which is running stable)

deployment/builder/app/main.py
nlsdfnbch: done, also renamed `_print` to `print_to_stderr`, because duh.

deployment/builder/app/main.py
nlsdfnbch: it should!

nlsdfnbch:
I've updated the `docker-compose.stable.yml` to use a `-stable` suffix in the urls. This *should* allow latest and stable versions to run simultaneiously.

deployment/docker-compose.latest.yml
karlb: > However, if it does not "run out of the box", we should probably adapt it so it does, if that's possible.

I didn't want to put a walled with tokens inside the public repo, even if it's just test eth. That's the only thing missing for it to run. I don't see a good way to handle that. Do you see one?

palango:
Mh, How would traefik now where to route stuff? I thought we would have the domains `services-dev.r.n` and `services-stable.r.n`, each with it's own traefik instance

nlsdfnbch:
> Mh, How would traefik now where to route stuff? I thought we would have the domains `services-dev.r.n` and `services-stable.r.n`, each with it's own traefik instance

if I understood @ulope correctly, there's a wildcard inplace, so the above setup should work. 

But of course, having a different domain for the stable services would be neat too. I think that needs more configuring though.

deployment/docker-compose.latest.yml
```diff
@@ -0,0 +1,89 @@
...
+#   docker-compose pull && docker-compose restart
+#
+# ..Note::
+#
+#   The above commands assume you're already in the directory where
+#   where this docker-compose file is located.
+version: '3.4'
+
+x-defaults: &defaults
+  restart: "no"
```
palango: Why is this `"no"`

deployment/docker-compose.latest.yml
```diff
@@ -0,0 +1,89 @@
...
+  image: raidennetwork/raiden-services:latest
+  env_file: services.env
+    - ./state:/state
+    - ~/.ethereum/keystore:/keystore
+
+services:
+  pfs-ropsten:
```
palango: will these work? there's no config attached, will this take the config from `docker-compose.yml` by default? If so, we need to change the DB names and keystore files as well

deployment/docker-compose.yml
```diff
@@ -1,8 +1,26 @@
...
+# ..Note::
+#
+#   The above commands assume you're already in the directory where
+#   where this docker-compose file is located.
 version: '3.4'
 
 x-defaults: &defaults
   restart: always
-  image: raiden-services
+  image: raidennetwork/raiden-services:stable
```
palango: how do we get these to build?

nlsdfnbch: The build config for these can be found here:
https://cloud.docker.com/u/raidennetwork/repository/docker/raidennetwork/raiden-services/builds/edit

So you'll need a tag to be pushed to master in order to have that build.

deployment/docker-compose.latest.yml
nlsdfnbch: There was confusion on my side - this file should the `local` dev file. I'll submit a fix.

deployment/docker-compose.latest.yml
nlsdfnbch: see above

deployment/docker-compose.yml
palango: there is a tag, `v0.1.0` exists and should be picked up afaict

Issue #458, opened 2019-07-16
Crash in PFS related to rate limiter


palango:
## Problem Definition

@karlb Just saw this in our deployment

```
pfs-goerli_1            | 2019-07-16 09:38:12.429405 [info     ] Stopping Pathfinding Service... [__main__]
pfs-goerli_1            | Traceback (most recent call last):
pfs-goerli_1            |   File "/usr/local/lib/python3.7/runpy.py", line 193, in _run_module_as_main
pfs-goerli_1            |     "__main__", mod_spec)
pfs-goerli_1            |   File "/usr/local/lib/python3.7/runpy.py", line 85, in _run_code
pfs-goerli_1            |     exec(code, run_globals)
pfs-goerli_1            |   File "/services/src/pathfinding_service/cli.py", line 101, in <module>
pfs-goerli_1            |     main(auto_envvar_prefix="PFS")  # pragma: no cover
pfs-goerli_1            |   File "/usr/local/lib/python3.7/site-packages/click/core.py", line 764, in __call__
pfs-goerli_1            |     return self.main(*args, **kwargs)
pfs-goerli_1            |   File "/usr/local/lib/python3.7/site-packages/click/core.py", line 717, in main
pfs-goerli_1            |     rv = self.invoke(ctx)
pfs-goerli_1            |   File "/usr/local/lib/python3.7/site-packages/click/core.py", line 956, in invoke
pfs-goerli_1            |     return ctx.invoke(self.callback, **ctx.params)
pfs-goerli_1            |   File "/usr/local/lib/python3.7/site-packages/click/core.py", line 555, in invoke
pfs-goerli_1            |     return callback(*args, **kwargs)
pfs-goerli_1            |   File "/services/src/raiden_libs/cli.py", line 173, in call_with_blockchain_info
pfs-goerli_1            |     return callback(**params)
pfs-goerli_1            |   File "/services/src/raiden_libs/cli.py", line 118, in call_with_common_options_initialized
pfs-goerli_1            |     return func(**params)
pfs-goerli_1            |   File "/services/src/pathfinding_service/cli.py", line 75, in main
pfs-goerli_1            |     db_filename=state_db,
pfs-goerli_1            |   File "/services/src/pathfinding_service/service.py", line 99, in __init__
pfs-goerli_1            |     address_reachability_changed_callback=self.handle_reachability_change,
pfs-goerli_1            |   File "/services/src/raiden_libs/matrix.py", line 185, in __init__
[...]

PR #459, opened 2019-07-16
Fix 'MatrixListener' has no attr. 'rate_limiter'

karlb:
This uses far more mocking than I like, but I didn't see a simple way to
test this without having it do network stuff apart from using mocks.

Fixes https://github.com/raiden-network/raiden-services/issues/458.


tests/libs/test_matrix.py
```diff
@@ -163,3 +164,25 @@ def test_rate_limiter():
...
+        join_global_room=Mock(),
+    ):
+        listener = MatrixListener(
+            private_key=get_private_key(c1),
+            chain_id=ChainID(1),
+            service_room_suffix="_service",
+            message_received_callback=lambda _: None,
+            address_reachability_changed_callback=lambda _addr, _reachability: None,
+        )
+        listener._start_client()  # pylint: disable=protected-access
```
palango: Can you add a simple assert, like that the `startup_finished` event is set?

karlb: Done!

PR #466, opened 2019-07-19
Update docker setup

nlsdfnbch:
Follow-up to #457


palango:
How will this work in general? Can I run two traefik instances at the same server and the same (sub)domain? If so, are all requests routes to both of the instances?

palango:
This is outdated by now, let's revive this later.

PR #469, opened 2019-07-19
Adds logging in the triggering monitor situation

Dominik1999:
Adds logging when the MS scheduled for triggering monitor but didn't at the end

### DIFF
--- a/src/monitoring_service/handlers.py
+++ b/src/monitoring_service/handlers.py
@@ -381,6 +381,11 @@ def action_monitoring_triggered_event_handler(event: Event, context: Context) ->
         non_closing_signer=event.non_closing_participant,
     )
     if monitor_request is None:
+        log.error(
+            "MonitorRequest cannot be found",
+            token_network_address=event.token_network_address,
+            channel_id=event.channel_identifier,
+        )
         return
 
     channel = context.db.get_channel(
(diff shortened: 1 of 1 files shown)


src/monitoring_service/handlers.py
```diff
@@ -381,16 +381,19 @@ def action_monitoring_triggered_event_handler(event: Event, context: Context) ->
         non_closing_signer=event.non_closing_participant,
     )
     if monitor_request is None:
+        log.error("MonitorRequest was lost or deleted before triggering", monitor_request=monitor_request)
```
palango: I'd write something like: "MonitorRequest cannot be found"

Also, it doesn't make sense to log `monitor_request` as it is `None`, but I'd log `token_network_address` and `channel_id`.

src/monitoring_service/handlers.py
```diff
@@ -381,16 +381,19 @@ def action_monitoring_triggered_event_handler(event: Event, context: Context) ->
...
         return
 
     channel = context.db.get_channel(
         token_network_address=monitor_request.token_network_address,
         channel_id=monitor_request.channel_identifier,
     )
     if channel is None:
+        log.error("Channel info was lost or deleted before triggering", monitor_request=monitor_request, channel=channel)
```
palango: Same as above, just put that the channel info isn't there, and logging channel doesn't make sense

src/monitoring_service/handlers.py
```diff
@@ -381,16 +381,19 @@ def action_monitoring_triggered_event_handler(event: Event, context: Context) ->
...
         return
 
     if not _is_mr_valid(monitor_request, channel):
+        log.error("MonitorRequest lost its validity, this is an error!", monitor_request=monitor_request, channel=channel)
```
palango: No need for "this is an error"

src/monitoring_service/handlers.py
Dominik1999: haha true

Issue #406, opened 2019-06-24
Automatic error monitoring


palango:
For deployments we should add an option to automatically report problems (aka uncatched exceptions).

A possible option is https://sentry.io

PR #470, opened 2019-07-19
Make services.env a template

palango:
Deployment side for https://github.com/raiden-network/raiden-services/issues/406


deployment/services.env.template
```diff
@@ -0,0 +1,15 @@
+PFS_HOST=0.0.0.0
+PFS_KEYSTORE_FILE=/keystore/<FILENAME>
```
karlb: Is this meant to be filled in manually? If it is, adding a comment like
```
# Make a copy called `services.env` and replace all <PLACEHOLDERS>.
```
would make that more clear. Otherwise I would wonder if some tooling is supposed to do that.

palango: Good point

PR #478, opened 2019-07-23
Update builder

palango:
Needs some testing. I'll start this after the scenario run finished.

Edit: good to go now.

---

This adds a new endpoint to the builder which listens to dockerhub webhooks. When the webhhok triggers, the git checkout is updated, the latest containers pulled and all services updated.

### DIFF
--- a/deployment/builder/app/main.py
+++ b/deployment/builder/app/main.py
@@ -1,37 +1,40 @@
+import json
 import os
 import subprocess
 import sys
 from pprint import pprint
+from typing import List
 
+import requests
 from flask import Flask, request
 
+ALL_SERVICE_NAMES = [
+    "pfs-ropsten",
+    "pfs-rinkeby",
+    "pfs-kovan",
+    "pfs-goerli",
+    "pfs-ropsten-with-fee",
+    "pfs-rinkeby-with-fee",
+    "pfs-kovan-with-fee",
+    "pfs-goerli-with-fee",
+    "ms-ropsten",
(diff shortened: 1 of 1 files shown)


deployment/builder/app/main.py
```diff
@@ -49,33 +52,50 @@ def print_to_stderr(s):
...
-                },
-                stream=sys.stderr,
+            print_to_stderr(
+                f"Fatal Error while updating images: Unhandled exception encountered: {e}"
             )
+            raise
         finally:
-            return "OK"
+            if callback_url:
+                requests.post(url=callback_url, json={"state": "error"})
```
karlb: This will submit an error also in successful cases. That's probably not right.

deployment/builder/app/main.py
```diff
@@ -78,20 +78,17 @@ def dockerhub():
...
-            return "OK"
-
         except ImageUpdateError:
             print_to_stderr("Error updating local images via docker registry!")
+            if callback_url:
+                requests.post(url=callback_url, json={"state": "error"})
         except Exception as e:
             print_to_stderr(
                 f"Fatal Error while updating images: Unhandled exception encountered: {e}"
             )
```
karlb: Is there any advantage of printing the error compared to just reraising it? I would both expect to leave a traceback in the log file.

palango: No, I just don't know how flask handles that and it didn't feel important enough to investigate

PR #483, opened 2019-07-29
Fix gas cost calculation in `claim_fees`

karlb:


palango:
Ah, I was talking about a (maybe 10%) margin on `claim_cost_gas`.

You're point is valid as well, but I think that's not important to add right now.

> For this case, implementing the bulk claim will make things a lot more reasonable. Did we decide whether that will go into the next mainnet release?

Not sure. Is it a simple extension of the contracts? If so I guess we can add it if we have some spare time?!

PR #484, opened 2019-07-30
Set pfs-info infos via cli arguments

### LINKED ISSUE #479: Add CLI options for PFS info settings
Currently the information shown in the PFS info endpoint is not configurable. We need to add CLI args for them.

Dominik1999:
Closes #479

### DIFF
--- a/src/pathfinding_service/cli.py
+++ b/src/pathfinding_service/cli.py
@@ -13,7 +13,11 @@ from web3 import Web3
 from web3.contract import Contract
 
 from pathfinding_service.api import ServiceApi
-from pathfinding_service.config import DEFAULT_API_HOST, DEFAULT_POLL_INTERVALL
+from pathfinding_service.config import (
+    DEFAULT_API_HOST,
+    DEFAULT_INFO_MESSAGE,
+    DEFAULT_POLL_INTERVALL,
+)
 from pathfinding_service.service import PathfindingService
 from raiden.settings import DEFAULT_NUMBER_OF_BLOCK_CONFIRMATIONS
 from raiden.utils.typing import BlockNumber, TokenAmount
@@ -47,6 +51,13 @@ log = structlog.get_logger(__name__)
     help="Number of block confirmations to wait for",
 )
 @click.option("--enable-debug", default=False, is_flag=True, hidden=True)
+@click.option("--operator", default="John Doe", type=str, help="Name of the service operator")
+@click.option(
+    "--info-message",
+    default=DEFAULT_INFO_MESSAGE,
+    type=str,
+    help="Place for a personal message to the customers",
+)
 @common_options("raiden-pathfinding-service")
 def main(  # pylint: disable-msg=too-many-arguments
     private_key: str,
@@ -57,6 +68,8 @@ def main(  # pylint: disable-msg=too-many-arguments
     confirmations: int,
     host: str,
     service_fee: TokenAmount,
+    operator: str,
+    info_message: str,
     enable_debug: bool,
 ) -> int:
     """ The Pathfinding service for the Raiden Network. """
@@ -80,6 +93,8 @@ def main(  # pylint: disable-msg=too-many-arguments
             service_fee=service_fee,
             debug_mode=enable_debug,
             one_to_n_address=contracts[CONTRACT_ONE_TO_N].address,
+            operator=operator,
+            info_message=info_message,
         )
         api.run(host=host)
 
--- a/tests/pathfinding/fixtures/api.py
+++ b/tests/pathfinding/fixtures/api.py
@@ -32,7 +32,12 @@ def api_sut(
     populate_token_network_case_1,  # pylint: disable=unused-argument
 ) -> Iterator[ServiceApi]:
     pathfinding_service_mock.address_to_reachability = address_to_reachability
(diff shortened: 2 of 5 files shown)


Dominik1999:
Might be missing (I don't know if needed):
- test cli args itself (they are not required and take any string)
- too many arguments / instances in `ServiceApi` acc. to mypy. Error is now ignored, but we could put the three new args into a dict and change that ...?!

src/pathfinding_service/cli.py
```diff
@@ -47,6 +47,19 @@
     help="Number of block confirmations to wait for",
 )
 @click.option("--enable-debug", default=False, is_flag=True, hidden=True)
+@click.option("--pfs-operator", default="John Doe", type=str, help="Name of the service operator")
+@click.option(
+    "--pfs-settings",
```
palango: I wonder if we should remove this for now, as it doesn't have a purpose right now.

src/pathfinding_service/cli.py
```diff
@@ -47,6 +47,19 @@
...
+    default="Raiden Default Settings - bidirectional Dijkstra",
+    type=str,
+    help="Information on technical implementation",
+)
+@click.option(
+    "--pfs-message",
+    default="This is your favorite pathfinding service",
+    type=str,
+    help="Number of block confirmations to wait for",
```
palango: This needs to be updated

src/pathfinding_service/cli.py
```diff
@@ -47,6 +47,19 @@
...
+@click.option("--pfs-operator", default="John Doe", type=str, help="Name of the service operator")
+@click.option(
+    "--pfs-settings",
+    default="Raiden Default Settings - bidirectional Dijkstra",
+    type=str,
+    help="Information on technical implementation",
+)
+@click.option(
+    "--pfs-message",
+    default="This is your favorite pathfinding service",
```
palango: Can we make this more meaningful? Like `PFS for token network registry 0xdeadbeef`, or is this to technical?

tests/pathfinding/fixtures/api.py
```diff
@@ -32,7 +32,13 @@ def api_sut(
     populate_token_network_case_1,  # pylint: disable=unused-argument
 ) -> Iterator[ServiceApi]:
     pathfinding_service_mock.address_to_reachability = address_to_reachability
-    api = ServiceApi(pathfinding_service_mock, one_to_n_address=Address(bytes([1] * 20)))
+    api = ServiceApi(
+        pathfinding_service_mock,
```
palango: Can you add the keyword in this line

tests/pathfinding/fixtures/api.py
```diff
@@ -47,7 +53,12 @@ def api_sut_with_debug(
 ) -> Iterator[ServiceApi]:
     pathfinding_service_mock.address_to_reachability = address_to_reachability
     api = ServiceApi(
-        pathfinding_service_mock, one_to_n_address=Address(bytes([1] * 20)), debug_mode=True
+        pathfinding_service_mock,
```
palango: same here

src/pathfinding_service/cli.py
Dominik1999: yep that is a good idea

src/pathfinding_service/cli.py
Dominik1999: done

tests/pathfinding/fixtures/api.py
Dominik1999: done

palango:
Looks good, but let's wait till https://github.com/raiden-network/raiden/pull/4486 is green as well

tests/pathfinding/test_api.py
```diff
@@ -264,20 +264,24 @@ def test_get_paths(api_url: str, addresses: List[Address], token_network_model:
...
             "chain_id": pathfinding_service_mock.chain_id,
-            "registry_address": to_checksum_address(pathfinding_service_mock.registry_address),
+            "registry_address": token_network_registry,
         },
-        "settings": "PLACEHOLDER FOR PATHFINDER SETTINGS",
         "version": pkg_resources.require("raiden-services")[0].version,
-        "operator": "PLACEHOLDER FOR PATHFINDER OPERATOR",
-        "message": "PLACEHOLDER FOR ADDITIONAL MESSAGE BY THE PFS",
+        "operator": "John Doe",
+        "message": "This is your favorite pfs for token network registry "
```
palango: Oh, one problem here:

When I put a custom message this will still add the registry address. That shouldn't happen.

Dominik1999: oh yes, you are right, will change

src/pathfinding_service/cli.py
```diff
@@ -47,6 +47,13 @@
     help="Number of block confirmations to wait for",
 )
 @click.option("--enable-debug", default=False, is_flag=True, hidden=True)
+@click.option("--operator", default="John Doe", type=str, help="Name of the service operator")
+@click.option(
+    "--info-message",
+    default="This is your favorite pfs for token network registry ",
```
palango: I'd put this into a constant and use that in the `info` endpoint for the comparison as well

Dominik1999: ok

PR #485, opened 2019-07-30
WIP: Update to latest raiden and raiden-contracts

karlb:
The raiden release is not ready yet, so I'm providing this for early
review to speed up the release process.

[skip ci]


karlb:
> Looks good that far. Should we update the raiden commit and just update the version later?

I'll do it. I expected the release to be done any minute now, but apparently this has not happened.

Issue #494, opened 2019-07-31
Remove handling of ChannelDeposit event


palango:
I see no point in handling the `ChannelDeposit` any more. The `CapacityUpdate`s provide all necessary information.

Removing the event handling would allow us to simplify the code and remove corner cases and possible timing issues.

palango:
This actually causes one bug in the scenario player:
```
2019-07-31 14:29:34.580800 [info     ] Received ChannelNewDeposit event [pathfinding_service.service] event_={'token_network_address': '0x07ba6a78dF3335D9f923051830e620aEA95691aA', 'channel_identifier': 4, 'participant_address': '0x13F5a5b0dC62343339900c89c51CE20fD9cA019c', 'total_deposit': 5, 'block_number': 1037667, 'type_name': 'ReceiveChannelNewDepositEvent'}
2019-07-31 14:29:34.580973 [debug    ] Calling `update_deposit`       [pathfinding_service.model.channel_view] capacity=5 deposit=0 new_capacity=10 total_deposit=5
```

Here the event arrives after the `CapacityUpdate`, but the capacity is still increased.

PR #496, opened 2019-08-01
Remove handling of ChannelNewDeposit event

palango:
Handling the event turned out to be a source of issues. Besides, it
doesn't provide any advantages over just using the capacity updates from
the nodes.

This commit removes all internal handling of the deposit event.

Fixes https://github.com/raiden-network/raiden-services/issues/494

### DIFF
--- a/tests/pathfinding/test_blockchain_integration.py
+++ b/tests/pathfinding/test_blockchain_integration.py
@@ -106,7 +106,7 @@ def test_pfs_with_mocked_client(  # pylint: disable=too-many-arguments
     )
 
     # check that deposits, settle_timeout and transfers got registered
-    for index, _ in enumerate(channel_descriptions_case_1):
+    for index in range(len(channel_descriptions_case_1)):
         channel_identifier = channel_identifiers[index]
         p1_address, p2_address = token_network_model.channel_id_to_addresses[channel_identifier]
         view1: ChannelView = graph[p1_address][p2_address]["view"]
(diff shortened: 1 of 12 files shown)


tests/pathfinding/test_blockchain_integration.py
```diff
@@ -108,30 +106,11 @@ def test_pfs_with_mocked_client(  # pylint: disable=too-many-arguments
...
-            _p2_index,
-            p2_deposit,
-            _p2_capacity,
-            _p2_fee,
-            _p2_reveal_timeout,
-            _p2_reachability,
-            _settle_timeout,
-        ),
-    ) in enumerate(channel_descriptions_case_1):
+    for index, _ in enumerate(channel_descriptions_case_1):
```
karlb: Nice to see this more concise. Since we don't need the list elements, we could even go with
```suggestion
    for index in range(len(channel_descriptions_case_1)):
```

tests/pathfinding/test_capacity_updates.py
```diff
@@ -112,7 +94,7 @@ def test_pfs_rejects_capacity_update_with_wrong_chain_id(
 ):
     setup_channel(pathfinding_service_web3_mock)
 
-    message = get_updatepfs_message(
+    message = get_capacity_update_message(
```
karlb: Good that you changed that, too! I already though we got all the updatepfs occurrences last time...

tests/pathfinding/test_graphs.py
```diff
@@ -99,7 +99,7 @@ def test_routing_simple(
     view01: ChannelView = token_network_model.G[addresses[0]][addresses[1]]["view"]
     view10: ChannelView = token_network_model.G[addresses[1]][addresses[0]]["view"]
 
-    assert view01.deposit == 100
+    # assert view01.deposit == 100
```
karlb: Why not remove that line?

palango: Classic, I was sure I would remember to remove this once the tests pass :-/

PR #497, opened 2019-08-01
Update dev deps and add new mypy check

palango:
~~Merge after #496~~

Updates the dev dependencies and enables a new mypy check

### DIFF
--- a/src/monitoring_service/handlers.py
+++ b/src/monitoring_service/handlers.py
@@ -490,8 +490,7 @@ def action_claim_reward_triggered_event_handler(event: Event, context: Context) 
 
     # check that the latest update was ours and that we didn't send a transaction yet
     can_claim = (
-        channel is not None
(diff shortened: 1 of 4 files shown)


src/monitoring_service/handlers.py
```diff
@@ -490,8 +490,7 @@ def action_claim_reward_triggered_event_handler(event: Event, context: Context)
 
     # check that the latest update was ours and that we didn't send a transaction yet
     can_claim = (
-        channel is not None
-        and channel.claim_tx_hash is None
+        channel.claim_tx_hash is None
```
Dominik1999: is this related to the title? Here I understand that we add a new condition to the `can_claim` statement

palango: That's a "problem" found by a new option in the updated mypy, I adapted the title.

PR #499, opened 2019-08-05
Return early in case of unknown message type

### LINKED ISSUE #500: Unassigned variable when receiving unhandled message
Issue for #499, so we don't forget this in the changelog.

palango:
Otherwise `changed_cvs` will not be initialized.

Closes #500


palango:
Yes, I'll add a test

PR #503, opened 2019-08-07
Confirmed receipts

palango:
This waits for confirmation blocks before checking the  status of pending transactions.

Follow-up for #502

### DIFF
--- a/src/monitoring_service/service.py
+++ b/src/monitoring_service/service.py
@@ -167,14 +167,30 @@ class MonitoringService:  # pylint: disable=too-few-public-methods
             handle_event(event, self.context)
             self.context.db.remove_scheduled_event(scheduled_event)
 
-        # check pending transactions
-        # this is done here so we don't have to block waiting for receipts in the state machine
+        # Check pending transactions
+        # This is done here so we don't have to block waiting for receipts in the state machine
+        self._check_pending_transactions()
+
+    def _check_pending_transactions(self) -> None:
+        """ Checks if pending transaction have been mined and confirmed.
+
+        In theory it's not necessary to check all pending transactions, but only the one with the
+        smallest nonce, and continue from there when this one is mined and confirmed. However,
+        as it is not expected that this list becomes to big this isn't optimized currently.
+        """
+
         for tx_hash in self.context.db.get_waiting_transactions():
             receipt = self.web3.eth.getTransactionReceipt(tx_hash)
-
             if receipt is not None:
-                self.context.db.remove_waiting_transaction(tx_hash)
+                tx_block = receipt.get("blockNumber")
+                if tx_block is None:
+                    return
 
+                confirmation_block = tx_block + self.context.required_confirmations
+                if self.context.last_known_block < confirmation_block:
+                    return
+
+                self.context.db.remove_waiting_transaction(tx_hash)
                 if receipt["status"] == 1:
                     log.info(
                         "Transaction was mined successfully",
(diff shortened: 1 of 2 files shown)


src/monitoring_service/service.py
```diff
@@ -170,14 +170,30 @@ def _process_new_blocks(self, last_block: BlockNumber) -> None:
...
+
+        In theory it's not necessary to check all pending transactions, but only the one with the
+        smallest nonce, and continue from there when this one is mined and confirmed. However,
+        as it is not expected that this list becomes to big this isn't optimized currently.
+        """
+
         for tx_hash in self.context.db.get_waiting_transactions():
             receipt = self.web3.eth.getTransactionReceipt(tx_hash)
-
             if receipt is not None:
```
karlb: I usually prefer the "return early" style:
```python
if receipt is None:
    continue
```
That's also more consistent with the check that follow.

src/monitoring_service/service.py
```diff
@@ -169,14 +170,30 @@ def _process_new_blocks(self, last_block: BlockNumber) -> None:
...
-                self.context.db.remove_waiting_transaction(tx_hash)
+                tx_block = receipt.get("blockNumber")
+                if tx_block is None:
+                    return
```
karlb:
```suggestion
                    continue
```

src/monitoring_service/service.py
```diff
@@ -169,14 +170,30 @@ def _process_new_blocks(self, last_block: BlockNumber) -> None:
...
 
+                confirmation_block = tx_block + self.context.required_confirmations
+                if self.context.last_known_block < confirmation_block:
+                    return
```
karlb:
```suggestion
                    continue
```

src/monitoring_service/service.py
```diff
@@ -167,14 +167,30 @@ def _process_new_blocks(self, last_block: BlockNumber) -> None:
...
             receipt = self.web3.eth.getTransactionReceipt(tx_hash)
-
             if receipt is not None:
-                self.context.db.remove_waiting_transaction(tx_hash)
+                tx_block = receipt.get("blockNumber")
+                if tx_block is None:
+                    return
 
+                confirmation_block = tx_block + self.context.required_confirmations
+                if self.context.last_known_block < confirmation_block:
```
karlb: This should not be `last_known_block`, but the current block (`web3.eth.blockNumber`). Those are usually `confirmation_blocks` apart and the actual logic behind this check is also less clear when using `last_known_block`.

src/monitoring_service/service.py
palango: Ouch, that was bad

tests/monitoring/monitoring_service/test_service.py
```diff
@@ -0,0 +1,25 @@
...
+def test_check_pending_transactions(web3, wait_for_blocks, monitoring_service: MonitoringService):
+    required_confirmations = 3
+
+    monitoring_service.context.required_confirmations = required_confirmations
+    monitoring_service.database.add_waiting_transaction(waiting_tx_hash=make_transaction_hash())
+
+    for tx_status in (0, 1):
+        tx_receipt = {"blockNumber": web3.eth.blockNumber, "status": tx_status}
+        with patch.object(web3.eth, "getTransactionReceipt", Mock(return_value=tx_receipt)):
+            with patch.object(
```
karlb: You can use multiple context managers with a single `with`, which reduced the nesting.

src/monitoring_service/service.py
palango: Good point

PR #508, opened 2019-08-08
Update to eth-utils 1.6.4 and fix new type warnings

palango:


karlb:
Won't this cause conflicts with raiden, which uses an older eth-utils?

palango:
Ah, I just wanted to proudly present the PR 🤷‍♂

src/pathfinding_service/api.py
```diff
@@ -362,10 +367,10 @@ def get(self) -> Tuple[dict, int]:
             "payment_address": to_checksum_address(self.pathfinding_service.address),
         }
         if info["message"] == DEFAULT_INFO_MESSAGE:
-            info["message"] = info["message"] + to_checksum_address(
+            info["message"] = info["message"] + to_checksum_address(  # type: ignore
```
karlb: Why do we need the type ignore?

src/pathfinding_service/database.py
```diff
@@ -216,7 +218,9 @@ def get_feedback_token(
 
         if token:
             return FeedbackToken(
-                token_network_address=decode_hex(token["token_network_address"]),
+                token_network_address=TokenNetworkAddress(
+                    decode_hex(token["token_network_address"])
```
karlb: Above we use `TokenNetworkAddress(to_canonical_address(...))`. Using the same would be more consistent.

src/raiden_libs/blockchain.py
```diff
@@ -147,14 +148,17 @@ def get_blockchain_events(
 
     events: List[Event] = []
     for event_dict in registry_events:
+        token_network_address = TokenNetworkAddress(
+            decode_hex(event_dict["args"]["token_network_address"])
```
karlb: Again, `to_canonical_address` would be more consistent then using `decode_hex` which is just possible due to the typecast.

src/raiden_libs/blockchain.py
```diff
@@ -218,19 +222,21 @@ def get_monitoring_blockchain_events(
         if event_name == MonitoringServiceEvent.NEW_BALANCE_PROOF_RECEIVED:
             events.append(
                 ReceiveMonitoringNewBalanceProofEvent(
-                    token_network_address=decode_hex(event["args"]["token_network_address"]),
+                    token_network_address=TokenNetworkAddress(
+                        decode_hex(event["args"]["token_network_address"])
```
karlb: Again, to_canonical_address would be more consistent then using decode_hex which is just possible due to the typecast.

src/raiden_libs/matrix.py
```diff
@@ -284,6 +284,9 @@ def _handle_message(self, room: Any, event: dict) -> bool:
             return False
 
         for message in messages:
+            if message.sender is None:
```
karlb: Can that happen? If not, I would prefer an assert.

src/pathfinding_service/api.py
palango: Without that I get:
```
src/pathfinding_service/api.py:370: error: Unsupported left operand type for + ("object")
```

src/pathfinding_service/database.py
palango: True, I missed that during my automatic replace.

src/raiden_libs/matrix.py
palango: Just checked, it cannot happen. This is checked in `raiden_libs.matrix.deserialize_messages`

src/pathfinding_service/api.py
karlb: I would prefer
```py
info["message"] = str(info["message"]) + ...
```
or
```py
assert isinstance(info['message'], str)
```
if mypy understands that. Type ignores are always a bit distracting.

palango: 👍

Issue #504, opened 2019-08-08
Avoid attacks by sending MRs after closing


karlb:
A client can theoretically send many MRs after a channel has closes. I must be impossible (or at least expensive) to make the MSes call `monitor` repeatedly with updated BPs by doing this. Otherwise an attacker could waste the MS` gas repeatedly with just a single channel.

Countermeasure: Ignore MRs that are sent significantly after closing the channel. The point in time where MRs start to be ignored must be between
* The confirmation of the ChannelClosed event in the client. It is reasonable to send MRs up to this point.
* The time at which MSes start to call `monitor`. Before, we can be sure that no gas has been wasted.

A simple choice would be the timestamp of (ChannelClosed Event + 20 blocks).

See also https://github.com/raiden-network/raiden-services/pull/502#discussion_r311468046.

PR #510, opened 2019-08-08
Ignore MRs sent after the channel has been closed

karlb:
Closes https://github.com/raiden-network/raiden-services/issues/504.

### DIFF
--- a/src/monitoring_service/constants.py
+++ b/src/monitoring_service/constants.py
@@ -12,5 +12,5 @@ KEEP_MRS_WITHOUT_CHANNEL: timedelta = timedelta(minutes=15)
 MATRIX_RATE_LIMIT_ALLOWED_BYTES = 5_000_000
 MATRIX_RATE_LIMIT_RESET_INTERVAL = timedelta(minutes=5)
 
-# Number of blocks after the close, during which MR are still being accepted
+# Number of blocks after the close, during which MRs are still being accepted
 CHANNEL_CLOSE_MARGIN: int = 10
--- a/src/monitoring_service/constants.py
+++ b/src/monitoring_service/constants.py
@@ -11,3 +11,6 @@ KEEP_MRS_WITHOUT_CHANNEL: timedelta = timedelta(minutes=15)
 # bursts of messages, this is only enforced as an average over 5 minutes.
 MATRIX_RATE_LIMIT_ALLOWED_BYTES = 5_000_000
 MATRIX_RATE_LIMIT_RESET_INTERVAL = timedelta(minutes=5)
+
+# Number of blocks after the close, during which MR are still being accepted
+CHANNEL_CLOSE_MARGIN: int = 10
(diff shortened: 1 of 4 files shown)


src/monitoring_service/constants.py
```diff
@@ -11,3 +11,6 @@
 # bursts of messages, this is only enforced as an average over 5 minutes.
 MATRIX_RATE_LIMIT_ALLOWED_BYTES = 5_000_000
 MATRIX_RATE_LIMIT_RESET_INTERVAL = timedelta(minutes=5)
+
+# Number of blocks after the close, during which MR are still being accepted
```
palango: nit: s/MR/MRs

src/monitoring_service/database.py
```diff
@@ -244,6 +244,22 @@ def load_state(self) -> MonitoringServiceState:
         )
         return ms_state
 
+    def channel_close_age(
+        self, token_network_address: TokenNetworkAddress, channel_id: ChannelID
+    ) -> Optional[int]:
+        """ How many blocks ago was the given channel closed? """
+        row = self.conn.execute(
+            """
+                SELECT blockchain.latest_known_block, channel.closing_block
```
palango: SQL question: how do I know here that `identifier = ? AND token_network_address = ?` are related to the channel table?

src/request_collector/server.py
```diff
@@ -94,6 +94,21 @@ def on_monitor_request(self, request_monitoring: RequestMonitoring) -> None:
...
+        # updating the BP after the MS has already called `monitor`, see
+        # https://github.com/raiden-network/raiden-services/issues/504.
+        close_age = self.state_db.channel_close_age(
+            token_network_address=monitor_request.token_network_address,
+            channel_id=monitor_request.channel_identifier,
+        )
+        # This is x blocks after that event is already confirmed, so that should be plenty!
+        if close_age is not None and close_age >= CHANNEL_CLOSE_MARGIN:
+            log.warning(
+                "MR for long closed channel", monitor_request=monitor_request, close_age=close_age
```
palango: Maybe add a verb like *ignoring* or *skipping* to the message?

src/request_collector/server.py
```diff
@@ -121,5 +136,4 @@ def on_monitor_request(self, request_monitoring: RequestMonitoring) -> None:
             reward_amount=monitor_request.reward_amount,
         )
 
-        with self.state_db.conn:
-            self.state_db.upsert_monitor_request(monitor_request)
+        self.state_db.upsert_monitor_request(monitor_request)
```
palango: Why can we remove the transaction here?

src/monitoring_service/database.py
karlb: Yes, because only that table has columns of that name. But I can make it clearer by prefixing the columns with the table name.

src/request_collector/server.py
karlb: It is only a single operation. Those are atomic anyway.

PR #519, opened 2019-08-19
Update to latest Raiden and contracts

palango:
Update client and contracts deps

### TODO:
- [ ] The UserAddressManager now has `.start()` and `.stop()` methods that need to be called to `start` / `stop` the event listener. 
  -  `start` should be called just before login is performed on the client. Stop somewhere before the client it disconnected

⚠️ Don't merge before the hackathon, as we don't want to deploy this yet.

### DIFF
--- a/tests/pathfinding/fixtures/network_service.py
+++ b/tests/pathfinding/fixtures/network_service.py
@@ -4,6 +4,7 @@ from typing import Callable, Dict, Generator, List
 from unittest.mock import Mock, patch
 
 import pytest
+from eth_utils import to_canonical_address
 from web3 import Web3
 from web3.contract import Contract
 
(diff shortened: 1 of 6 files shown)


tests/pathfinding/fixtures/network_service.py
```diff
@@ -23,6 +24,11 @@
 from raiden_contracts.constants import CONTRACT_TOKEN_NETWORK_REGISTRY, CONTRACT_USER_DEPOSIT
 
 
+@pytest.fixture(scope="session")
+def pfs_address(create_service_account) -> Address:
```
karlb: Why do we need this? We can always just read the address from the PFS object created by another fixture?

PR #521, opened 2019-08-20
Graph pruning

### LINKED ISSUE #520: Long blocking route search
## Problem Definition

There were two recent examples where the PFS seemed to get stuck while looking for routes.

It's not clear to me if this is based on an endless loop or just long search in bigger networks.

- The networks had higher than usual amount of channels, the attached logs has at least 65 channels in the queried token network
- The routes found don't match the topology described in the scenario file.

Scenario config: https://gist.github.com/czepluch/94e387c0aac852825bdf70dd4cf237ef

Logs:
[hang4.log](https://github.com/raiden-network/raiden-services/files/3516460/hang4.log)

palango:
It showed in #520 that in bigger token networks the search was blocking
the PFS for longer intervals. The main reason for that was that the path
finding was trying many routes where participants were offline.

This can easily be improved by pruning the graph on unreachable nodes
before the routing, so that the graph on which the search is done
consists only of reachable nodes.

This can be improved further in the future by also pruning channels with
to little capacity, etc.

Fixes #520

### DIFF
--- a/src/pathfinding_service/api.py
+++ b/src/pathfinding_service/api.py
@@ -18,7 +18,6 @@ from flask_restful import Api, Resource
 from gevent.pywsgi import WSGIServer
 from marshmallow import fields
 from marshmallow_dataclass import add_schema
-from networkx.exception import NetworkXNoPath, NodeNotFound
 from web3 import Web3
 
 import pathfinding_service.exceptions as exceptions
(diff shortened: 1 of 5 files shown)


src/pathfinding_service/model/token_network.py
```diff
@@ -30,6 +31,23 @@
...
+    for p1, p2 in graph.edges:
+        nodes_online = (
+            address_to_reachability[p1] == AddressReachability.REACHABLE
+            and address_to_reachability[p2] == AddressReachability.REACHABLE
+        )
+        if nodes_online:
+            result.add_edge(p1, p2, view=graph[p1][p2]["view"])
+            result.add_edge(p2, p1, view=graph[p2][p1]["view"])
+
+    return result
```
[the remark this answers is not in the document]
Dominik1999: maybe pruned_graph?

src/pathfinding_service/api.py
```diff
@@ -133,17 +132,17 @@ def post(self, token_network_address: str) -> Tuple[dict, int]:
             value = getattr(path_req, arg)
             if value is not None:
                 optional_args[arg] = value
-        try:
```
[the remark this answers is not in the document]
Dominik1999: ah ok found where we do that now

src/raiden_libs/utils.py
```diff
@@ -15,7 +15,7 @@ def public_key_to_address(public_key: PublicKey) -> Address:
 def private_key_to_address(private_key: Union[str, bytes]) -> Address:
     """ Converts a private key to an Ethereum address. """
     if isinstance(private_key, str):
-        private_key = to_bytes(hexstr=private_key)
+        private_key = decode_hex(private_key)
```
Dominik1999: why is this in the PR?

tests/pathfinding/test_graphs.py
```diff
@@ -137,8 +138,8 @@ def test_capacity_check(
     """ Test that the mediation fees are included in the capacity check """
     # First get a path without mediation fees. This must return the shortest path: 4->1->0
     paths = token_network_model.get_paths(
-        addresses[4],
-        addresses[0],
+        source=addresses[4],
```
Dominik1999: nice that you think about these things

src/raiden_libs/utils.py
palango: Was just a drive by fix.

PR #529, opened 2019-09-05
Fix message deserialization

palango:
Follow-up to https://github.com/raiden-network/raiden-services/pull/528 as it didn't fix the message deserialization.

### DIFF
--- a/src/raiden_libs/matrix.py
+++ b/src/raiden_libs/matrix.py
@@ -1,4 +1,3 @@
-import json
 import sys
 from datetime import datetime, timedelta
 from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Union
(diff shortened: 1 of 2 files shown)


src/raiden_libs/matrix.py
```diff
@@ -75,17 +74,6 @@ def check_and_count(self, sender: Address, added_bytes: int) -> bool:
         return True
 
 
-def message_from_dict(data: Dict[str, Any]) -> Message:
-    if "_type" not in data:
-        raise SerializationError("Invalid message data. Can not find the data type") from None
-    if data["_type"] not in VALID_MESSAGE_TYPES:
```
palango: Not sure if we should check somewhere that the message is one of the types in `SERVICE_MESSAGES`

karlb: Messages of unexpected types will just not be handled in `service.handle_message`. I don't see a need for checks in addition to that. This would mean we can remove `SERVICE_MESSAGES` and `VALID_MESSAGE_TYPES`.

PR #536, opened 2019-09-10
Improve fee estimate

### LINKED ISSUE #535: Wrong fee estimation
While writing a scenario for proprtional fees I saw wrong fee estimations from the PFS.

See client issue https://github.com/raiden-network/raiden/issues/4844

And the spec https://raiden-network-specification.readthedocs.io/en/latest/mediation_fees.html

palango:
Fixes #535

Before we applied the proportional fee on the incoming amount twice,
without taking the incoming fee into account. This commit changes that
and adds tests.

### DIFF
--- a/src/pathfinding_service/model/channel_view.py
+++ b/src/pathfinding_service/model/channel_view.py
@@ -69,9 +69,13 @@ class ChannelView:
     ) -> Optional[FeeAmount]:
         """Returns the mediation fee for this channel when transferring the given amount"""
         try:
-            return self.fee_schedule_sender.fee_payer(amount, balance)
+            imbalance_fee = self.fee_schedule_sender.imbalance_fee(amount=amount, balance=balance)
         except UndefinedMediationFee:
             return None
+
+        flat_fee = self.fee_schedule_sender.flat
+        prop_fee = int(round(amount * self.fee_schedule_sender.proportional / 1e6))
+        return FeeAmount(flat_fee + prop_fee + imbalance_fee)
 
     def backwards_fee_receiver(
         self, balance: Balance, amount: PaymentWithFeeAmount
@@ -91,7 +95,7 @@ class ChannelView:
 
         try:
             imbalance_fee = self.fee_schedule_receiver.imbalance_fee(
-                amount=PaymentWithFeeAmount(amount - fee_in(imbalance_fee=FeeAmount(0))),
+                amount=PaymentWithFeeAmount(-(amount - fee_in(imbalance_fee=FeeAmount(0)))),
                 balance=balance,
             )
         except UndefinedMediationFee:
--- a/src/pathfinding_service/model/channel_view.py
+++ b/src/pathfinding_service/model/channel_view.py
@@ -8,6 +8,7 @@ from marshmallow_dataclass import add_schema
 
 from pathfinding_service.constants import DEFAULT_REVEAL_TIMEOUT
 from pathfinding_service.exceptions import InvalidPFSFeeUpdate
+from raiden.exceptions import UndefinedMediationFee
 from raiden.transfer.mediated_transfer.mediation_fee import FeeScheduleState as FeeScheduleRaiden
 from raiden.utils.typing import (
     Address,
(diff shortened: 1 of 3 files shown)


palango:
@karlb Please have a look

src/pathfinding_service/model/channel_view.py
```diff
@@ -68,7 +69,26 @@ def forward_fee_sender(self, amount: PaymentWithFeeAmount) -> FeeAmount:
 
     def forward_fee_receiver(self, amount: PaymentWithFeeAmount) -> FeeAmount:
         """Return the mediation fee for this channel when receiving the given amount"""
-        return self.fee_schedule_receiver.fee_payee(amount, Balance(self.capacity))
+        temp_fee = round(
```
karlb: Do we need forward fee calculation at all in the PFS?

src/pathfinding_service/model/channel_view.py
```diff
@@ -62,13 +64,40 @@ def update_capacity(
...
+                        (amount + self.fee_schedule_receiver.flat + imbalance_fee)
+                        / (1 - self.fee_schedule_receiver.proportional / 1e6)
+                    )
+                    - amount
+                )
+            )
+
+        try:
+            imbalance_fee = self.fee_schedule_receiver.imbalance_fee(
+                amount=PaymentWithFeeAmount(amount - fee_in(imbalance_fee=FeeAmount(0))),
```
karlb: The sign has to be flipped on this one:
```suggestion
                amount=PaymentWithFeeAmount(-(amount - fee_in(imbalance_fee=FeeAmount(0)))),
```

src/pathfinding_service/model/token_network.py
```diff
@@ -63,6 +65,39 @@ def window(seq: Sequence, n: int = 2) -> Iterable[tuple]:
         yield result
 
 
+def backwards_fee_sender(
```
karlb: Isn't this the same as in `channel_view.py`? Any reason for the duplication?

src/pathfinding_service/model/token_network.py
```diff
@@ -76,19 +111,36 @@ def __init__(
...
-                ) + self.G[mediator][next_node]["view"].forward_fee_sender(total)
-                total += fee_for_this_mediator
-                self.fees.append(fee_for_this_mediator)
+                view_in: ChannelView = self.G[prev_node][mediator]["view"]
+                view_out: ChannelView = self.G[mediator][next_node]["view"]
+
+                fee_out = view_out.backwards_fee_sender(
+                    balance=Balance(view_out.capacity), amount=total
+                )
+                if fee_out is None:
```
karlb: Depending on whether (and how) we share this code with the client, we could just let the `UndefinedMediationFee` exception bubble up and skip the special case, here.

src/pathfinding_service/model/token_network.py
palango: Nice find, I forgot to delete them when I moved them to the channel view.

src/pathfinding_service/model/token_network.py
palango: Good point, as it looks now we could remove all `try except` blocks in the functions and catch it toplevel. But let's discuss tomorrow how we can reuse the client code.

tests/pathfinding/test_fee_schedule.py
```diff
@@ -131,6 +131,8 @@ def test_fees_in_routing():
     # Now let's try imbalance fees
     tn.set_fee(2, 3, imbalance_penalty=[(TA(0), FA(0)), (TA(200), FA(200))])
     assert tn.estimate_fee(1, 3) == 10
+
+    tn.set_fee(2, 3, imbalance_penalty=[(TA(0), FA(200)), (TA(200), FA(0))])
     assert tn.estimate_fee(3, 1) == -10
```
karlb: Why do you flip both the `imbalance_penalty` and the transfer direction at the same time? This makes it harder to compare it with the assert above.

Why not do
```python
# Now let's try imbalance fees
tn.set_fee(2, 3, imbalance_penalty=[(TA(0), FA(0)), (TA(200), FA(200))])
assert tn.estimate_fee(1, 3) == 10
assert tn.estimate_fee(3, 1) == -10
```
?

I'm also surprised that your code passes. After flipping both direction and fee schedule, I would have expected the same fee as before. Where is the error in my reasoning?

palango: This is weird, the test passes like this as well:
```py
    tn.set_fee(2, 3, imbalance_penalty=[(TA(0), FA(0)), (TA(200), FA(200))])
    assert tn.estimate_fee(1, 3) == 10
    assert tn.estimate_fee(3, 1) == 10

    tn.set_fee(2, 3, imbalance_penalty=[(TA(0), FA(200)), (TA(200), FA(0))])
    assert tn.estimate_fee(1, 3) == -10
    assert tn.estimate_fee(3, 1) == -10
```

src/pathfinding_service/model/token_network.py
palango: Let's keep this code short term to focus on the scenarios. Medium term we want to re-use the code from the client, so that there's only one implementation.

PR #546, opened 2019-09-16
Catch exceptions in event handlers

karlb:
This makes it harder for attackers to take down a MS. Before, even
restarting would not have helped in many cases because the same Event
would have been retried forever. Now, events are discarded when they
throw an exception and the MS continues to run.

To make sure that we don't ignore real errors, the exceptions are logged
and sentry issues are created.

There are no tests for this because I did not see a simple enough way to
exercise this in a meaningful manner.

### DIFF
--- a/src/monitoring_service/service.py
+++ b/src/monitoring_service/service.py
@@ -3,6 +3,7 @@ import time
 from datetime import datetime
 from typing import Callable, Dict
 
+import sentry_sdk
 import structlog
 from web3 import Web3
 from web3.contract import Contract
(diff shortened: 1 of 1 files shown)


src/monitoring_service/service.py
```diff
@@ -50,12 +51,25 @@ def check_gas_reserve(web3: Web3, private_key: str) -> None:
...
+        with sentry_sdk.push_scope() as sentry_scope:
+            sentry_scope.set_tag("event", event.__class__.__name__)
+            try:
+                handler(event, context)
+                log.debug(
+                    "Processed event", num_scheduled_events=context.db.scheduled_event_count()
+                )
+            except Exception as ex:  # pylint: disable=broad-except
+                log.error("Error during event handler", event=event, exc_info=ex)
+                sentry_sdk.capture_exception(ex)
```
palango: We need to check whether or not this works when sentry hasn't been initialized.

karlb: I did a very basic tests and apparently it does:
```
Python 3.7.1 (default, Nov 12 2018, 13:22:25) 
[GCC 6.3.0 20170516] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sentry_sdk
>>> sentry_sdk.capture_exception(Exception('foo'))
>>> 
```

PR #552, opened 2019-09-24
Store data in channels, not views

karlb:
Having the data in channel views is confusing because data that belongs the the channel is duplicated and has to be kept in sync across different views. Now the views are really only views on the channel that contains all data.

The `ChannelView` class became quite long, but I didn't see a way around it without breaking mypy.


src/pathfinding_service/model/channel_view.py
```diff
@@ -37,25 +37,103 @@ def from_raiden(cls, fee_schedule: FeeScheduleRaiden, timestamp: datetime) -> "F
...
-    )
-    capacity: TokenAmount = TokenAmount(0)
-    reveal_timeout: int = DEFAULT_REVEAL_TIMEOUT
-    update_nonce: Nonce = Nonce(0)
-    fee_schedule_sender: FeeSchedule = field(default_factory=FeeSchedule)
-    fee_schedule_receiver: FeeSchedule = field(default_factory=FeeSchedule)
+    fee_schedule1: FeeSchedule = field(default_factory=FeeSchedule)
+    fee_schedule2: FeeSchedule = field(default_factory=FeeSchedule)
+
+    capacity1: TokenAmount = TokenAmount(0)
```
palango: I think it would be nicer to order these in pairs as well.

src/pathfinding_service/model/token_network.py
```diff
@@ -205,38 +205,32 @@ def handle_channel_opened_event(
...
-        # Choosing which direction to add by execution order is not very
-        # robust. We might want to change this to either
-        # * participant1 < participant2 or
-        # * same as in contract (which would require an additional attribute on ChannelView)
-        if channel_view.channel_id not in self.channel_id_to_addresses:
+        # Only add it once per channel, not once per ChannelView
+        if channel_view.participant1 < channel_view.participant2:
             self.channel_id_to_addresses[channel_view.channel_id] = (
                 channel_view.participant1,
                 channel_view.participant2,
```
palango: For some reason I cannot comment below:
I wonder if we should make the graph bidirectional now and store the channel instead of the views.

src/pathfinding_service/schema.sql
```diff
@@ -11,18 +11,28 @@ CREATE TABLE token_network (
...
+    capacity2       HEX_INT NOT NULL,
+    reveal_timeout2 HEX_INT NOT NULL,
+    update_nonce2   HEX_INT,
+
+    -- From PFSFeeUpdate
+    fee_schedule1   JSON,
+    fee_schedule2   JSON,
+
+    PRIMARY KEY (token_network_address, channel_id),
+    CHECK (participant1 < participant2),
```
palango: Is this the same comparison as in python?

src/pathfinding_service/service.py
```diff
@@ -231,20 +230,21 @@ def handle_channel_closed(self, event: ReceiveChannelClosedEvent) -> None:
         log.info("Received ChannelClosed event", event_=event)
 
         token_network.handle_channel_closed_event(channel_identifier=event.channel_identifier)
-        self.database.delete_channel_views(event.channel_identifier)
+        self.database.delete_channel(event.token_network_address, event.channel_identifier)
```
palango: Nice, that could have ended badly.

src/pathfinding_service/service.py
```diff
@@ -231,20 +230,21 @@ def handle_channel_closed(self, event: ReceiveChannelClosedEvent) -> None:
...
 
     def handle_message(self, message: Message) -> None:
+        changed_channel: Optional[Channel]
```
palango: Should you set a default of `None` here or is this implicit?

src/pathfinding_service/model/token_network.py
karlb: I don't think that will work when we search the graph. The effect of traversing depends on the direction, so we need a directed graph AFAIK.

src/pathfinding_service/schema.sql
karlb: This is a good point. Due to the unusual checksumming, we might need to do `lower(participant1) < participant2`. I'll check it.

src/pathfinding_service/service.py
karlb: There is no default. It *will* get initialized or we won't reach a place where it's used. I only needed to add the type declaration because otherwise mypy would assume `Channel` instead of `Optional[Channel]`.

Would you prefer a different way to write it?

palango: Ah, I missed the `return` there. I'd still feel that `changed_channel: Optional[Channel] = None` is more explicit.

src/pathfinding_service/model/token_network.py
palango: Good point.

Issue #555, opened 2019-09-26
Update to contracts 0.32


palango:
Update the contracts package after https://github.com/raiden-network/raiden/issues/4980

PR #557, opened 2019-09-26
Update to latest Raiden and contracts

karlb:
Closes https://github.com/raiden-network/raiden-services/issues/555.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,7 @@ install:
 
 install-dev:
 	pip install -r requirements-dev.txt
+	pip install -e .
 
 dist:
 	python3 setup.py sdist bdist_wheel
--- a/src/monitoring_service/database.py
+++ b/src/monitoring_service/database.py
@@ -84,11 +84,12 @@ class SharedDatabase(BaseDatabase):
         kwargs = {
             key: val
             for key, val in zip(row.keys(), row)
-            if key not in ("non_closing_signer", "saved_at", "waiting_for_channel")
+            if key not in ("saved_at", "waiting_for_channel")
         }
         kwargs["token_network_address"] = to_canonical_address(kwargs["token_network_address"])
         kwargs["msc_address"] = to_canonical_address(kwargs["msc_address"])
         kwargs["closing_signature"] = decode_hex(kwargs["closing_signature"])
+        kwargs["non_closing_participant"] = to_canonical_address(kwargs.pop("non_closing_signer"))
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -29,6 +29,7 @@ install:
 
 install-dev:
 	pip install -r requirements-dev.txt
+	pip install -e .
```
palango: Nice, I wanted to do this for a long time.

src/monitoring_service/states.py
```diff
@@ -1,11 +1,13 @@
 from dataclasses import dataclass, field
 from typing import Iterable, Optional
 
+from eth_typing.evm import HexAddress
```
palango: Mh, a bit ugly to import this as well. Is this what the contracts use?

src/monitoring_service/handlers.py
```diff
@@ -370,6 +370,10 @@ def updated_head_block_event_handler(event: Event, context: Context) -> None:
 
 
 def _is_mr_valid(monitor_request: MonitorRequest, channel: Channel) -> bool:
+    if monitor_request.non_closing_signer != monitor_request.non_closing_participant:
```
palango: Do we have a test for this?

src/monitoring_service/database.py
```diff
@@ -84,11 +84,12 @@ def get_monitor_request(
         kwargs = {
             key: val
             for key, val in zip(row.keys(), row)
-            if key not in ("non_closing_signer", "saved_at", "waiting_for_channel")
```
palango: Was that a bug before?

src/monitoring_service/states.py
karlb: Yes, since they don't have access to Raiden's types.

src/monitoring_service/handlers.py
karlb: I'll add one and move this to the request collector.

src/monitoring_service/database.py
karlb: No, we generate the `non_closing_signer` out of the signature when constructing a `MonitorRequest` object. That column is only added to the DB to enforce the correct uniqueness of MRs.

The current handling of `MonitorRequest`, `UnsignedMonitorRequest`, 'non_closing_signer' and `non_closing_participant` is not ideal anymore, but I wanted to keep the changes small for now.

palango: Ok, great!

Issue #563, opened 2019-10-02
Add as many iterations into PFS imbalance fee calculation as we use in the Raiden client


Dominik1999:
Currently, the Raiden client iterates the imbalance fee calculation one time whereas the corresponding calculation in the pfs does not iterate.

We need to adjust the pfs fee calculation to the client calculation.

See here for reference https://docs.google.com/spreadsheets/d/1nHVCOhMcrM-L8brJj_C_Qb0Paws1qwozcNHU3VFcWEY/edit#gid=1257400215

palango:
There is some code for this in https://github.com/raiden-network/raiden/pull/5024, this needs to be copied to the PFS or be used from the raiden codebase.

PR #565, opened 2019-10-04
Iteration in fee approximation

karlb:
Use `fee_receiver` and `fee_sender` from raiden to get the latest fee calculation. Those are currently in a test utils package and should be moved in a following PR.

Closes https://github.com/raiden-network/raiden-services/issues/563.

### DIFF
--- a/src/pathfinding_service/model/channel.py
+++ b/src/pathfinding_service/model/channel.py
@@ -8,8 +8,7 @@ from marshmallow_dataclass import add_schema
 
 from pathfinding_service.constants import DEFAULT_REVEAL_TIMEOUT
 from pathfinding_service.exceptions import InvalidPFSFeeUpdate
-from raiden.exceptions import UndefinedMediationFee
-from raiden.tests.utils.mediation_fees import fee_receiver
+from raiden.tests.utils.mediation_fees import fee_receiver, fee_sender
 from raiden.transfer.mediated_transfer.mediation_fee import FeeScheduleState as FeeScheduleRaiden
 from raiden.utils.typing import (
     Address,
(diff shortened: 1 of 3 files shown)


karlb:
The PR needs the corresponding raiden PR merged first. I'll update it as soon as that PR is ready.

src/pathfinding_service/model/channel.py
```diff
@@ -8,7 +8,7 @@
 
 from pathfinding_service.constants import DEFAULT_REVEAL_TIMEOUT
 from pathfinding_service.exceptions import InvalidPFSFeeUpdate
-from raiden.exceptions import UndefinedMediationFee
+from raiden.tests.utils.mediation_fees import fee_receiver, fee_sender
```
Dominik1999: ok what I understand is that you take the fee calculation from the Raiden repo to do it in the pfs, now. Is that correct?

karlb: Yes, exactly.

tests/pathfinding/test_fee_schedule.py
```diff
@@ -122,51 +122,52 @@ def test_fees_in_balanced_routing():
     tn.set_fee(2, 3)
 
     # Let's try imbalance fees
+    # When approximation iterations matter, those are given as sums of the steps.
 
     # Incoming channel
-    tn.set_fee(2, 1, imbalance_penalty=[(TA(0), FA(0)), (TA(200), FA(200))])
-    assert tn.estimate_fee(1, 3) == 10
+    tn.set_fee(2, 1, imbalance_penalty=[(TA(0), FA(0)), (TA(2000), FA(200))])
+    assert tn.estimate_fee(1, 3) == 10 + 1
```
Dominik1999: can you add a comment that `+ 1 ` is due to the iteration?

PR #571, opened 2019-10-08
Update raiden version

palango:
Merge after #570 

This updates to the latest raiden packaged version and prepares to release a patch version with that.


palango:
> Does updating Raiden provide any benefit in this case, or is this release mostly to fix docker and the Raiden update is just a clean up?

Mostly the docker fix, but then Konrad also preferred the released version in the container for the bundle.

PR #573, opened 2019-10-08
Configure docker to have services-dev and -test

karlb:
With these settings, we should be able to run our last release on
services-test while the latest commit runs on services-dev. The builder
will need changes in another PR to work as intended.

### DIFF
--- a/deployment/docker-compose.yml
+++ b/deployment/docker-compose.yml
@@ -21,7 +21,7 @@ version: '3.4'
 x-defaults: &defaults
   restart: always
   env_file: .env
-  image: raidennetwork/raiden-services:latest
+  image: raidennetwork/raiden-services:stable
   volumes:
     - /data/state:/state
     - /data/keystore:/keystore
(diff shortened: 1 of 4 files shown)


karlb:
I'm not sure if sentry will autodetect the different hosts or if we have to change the config a bit.

deployment/docker-compose.yml
```diff
@@ -205,7 +205,7 @@ services:
       - /root/.ssh:/root/.ssh:ro
     labels:
       - "traefik.enable=true"
-      - "traefik.frontend.rule=Host: services-dev.raiden.network"
+      - "traefik.frontend.rule=Host: services-test.raiden.network"
```
palango: Do we need to overwrite this in `deployment/docker-compose.latest.yml` as well?

karlb: Good point! I didn't really care much for the build so far, so I missed it. Fixed now.

Issue #575, opened 2019-10-09
Use production matrix servers


palango:
Currently only development server are used.

https://github.com/raiden-network/raiden-services/blob/9acfd0e374f0b1aabd21d3f1c22c17410b7fca1f/src/raiden_libs/matrix.py#L138-L140

We need to either:
- Use production servers on mainnet
- Add a flag to set develop/production mode

PR #577, opened 2019-10-10
Use production matrix servers when on mainnet

karlb:
Closes https://github.com/raiden-network/raiden-services/issues/575

### DIFF
--- a/src/raiden_libs/matrix.py
+++ b/src/raiden_libs/matrix.py
@@ -14,6 +14,7 @@ from monitoring_service.constants import (
     MATRIX_RATE_LIMIT_ALLOWED_BYTES,
     MATRIX_RATE_LIMIT_RESET_INTERVAL,
 )
+from raiden.constants import Environment
 from raiden.exceptions import SerializationError, TransportError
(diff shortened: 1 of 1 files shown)


src/raiden_libs/matrix.py
```diff
@@ -28,10 +29,12 @@
 )
 from raiden.network.transport.utils import timeout_exponential_backoff
 from raiden.settings import (
+    DEFAULT_MATRIX_KNOWN_SERVERS,
     DEFAULT_TRANSPORT_MATRIX_RETRY_INTERVAL,
     DEFAULT_TRANSPORT_RETRIES_BEFORE_BACKOFF,
 )
 from raiden.storage.serialization.serializer import MessageSerializer
+from raiden.utils.cli import get_matrix_servers
```
palango: I'm a bit surprised this test passes https://github.com/raiden-network/raiden-services/pull/576/files#diff-043a8444ad20f9b349ab2ba851e6a229L146

But whatever, if it passes it doesn't seem to be necessary.

PR #579, opened 2019-10-11
Simplify builder

karlb:
The builder will now only care about the services, not about updating
itself, the docker-compose-files or traefik. These changes are rare and
can be done manually.

The builder (or the `update.sh` to be more precise) is also now able to
understand the difference between the `dev` and `test` deployments.

Closes https://github.com/raiden-network/raiden-services/issues/80.

### DIFF
--- a/deployment/builder/app/main.py
+++ b/deployment/builder/app/main.py
@@ -2,56 +2,19 @@ import json
 import os
 import subprocess
 import sys
-from pprint import pprint
-from typing import List
 
 import requests
 from flask import Flask, request
 
-ALL_SERVICE_NAMES = [
-    "pfs-ropsten",
-    "pfs-rinkeby",
-    "pfs-kovan",
-    "pfs-goerli",
-    "pfs-ropsten-with-fee",
-    "pfs-rinkeby-with-fee",
-    "pfs-kovan-with-fee",
-    "pfs-goerli-with-fee",
-    "ms-ropsten",
-    "ms-rinkeby",
-    "ms-kovan",
-    "ms-goerli",
-    "ms-goerli-backup",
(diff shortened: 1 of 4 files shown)


deployment/builder/app/main.py
```diff
@@ -61,77 +24,22 @@ def main():
...
-        raise ImageUpdateError from e
-
-    print_to_stderr("Fetching latest changes: git fetch")
     try:
-        subprocess.check_output(["git", "fetch", "--all"])
-    except subprocess.SubprocessError as e:
-        print_to_stderr(f"Fetching latest changes failed: {e}")
-        raise ImageUpdateError from e
+        tag = data["push_data"]["tag"]
+        os.chdir("/deployment")
```
palango: Is this idempotent?

deployment/docker-compose.yml
```diff
@@ -201,8 +201,9 @@ services:
     build: ./builder
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
-      - /root:/root
-      - /root/.ssh:/root/.ssh:ro
```
palango: Nice, so this wasn't necessary after all.

deployment/builder/app/main.py
karlb: It is an absolute path, so it is.

karlb:
I left an `update_builder.sh` on the server for now:
```
root@services-dev:~/raiden-services/deployment# cat update_builder.sh 
docker-compose -f docker-compose.yml -f docker-compose.latest.yml build builder && docker-compose -f docker-compose.yml -f docker-compose.latest.yml up -d builder
r
```
Definitvely not good documentation, but better than nothing.

PR #586, opened 2019-10-21
Update registration URL for services-dev

karlb:


deployment/docker-compose.latest.yml
```diff
@@ -84,6 +84,10 @@ services:
   msrc-goerli:
     <<: *defaults
 
+  registration-goerli:
```
palango: Do we also need `registration-goerli-backup`?

karlb: No. That one doesn't register any interesting URL because it's only used for the MS, which doesn't need the URL.

PR #587, opened 2019-10-21
Don't restart registration tasks

palango:
They should only run once.


deployment/docker-compose.latest.yml
```diff
@@ -86,6 +86,8 @@ services:
 
   registration-goerli:
     <<: *defaults
+    restart: "no"
```
karlb: You don't need to overwrite "restart". It's already "no".

Issue #590, opened 2019-10-21
Add raiden-contracts version to info endpoint


karlb:
This is helpful when debugging and when checking whether services did already restart with the latest version.

PR #592, opened 2019-10-21
Show contracts version in info endpoint

karlb:
Closes https://github.com/raiden-network/raiden-services/issues/590

### DIFF
--- a/src/pathfinding_service/api.py
+++ b/src/pathfinding_service/api.py
@@ -353,6 +353,7 @@ class FeedbackResource(PathfinderResource):
 
 class InfoResource(PathfinderResource):
     version = pkg_resources.get_distribution("raiden-services").version
+    contracts_version = pkg_resources.get_distribution("raiden-contracts").version
 
     def get(self) -> Tuple[dict, int]:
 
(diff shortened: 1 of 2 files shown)


src/pathfinding_service/api.py
```diff
@@ -353,6 +353,7 @@ def post(
 
 class InfoResource(PathfinderResource):
     version = pkg_resources.get_distribution("raiden-services").version
+    contracts_version = pkg_resources.get_distribution("raiden-contracts").version
```
palango: There's also `get_contracts_deployment_info` which returns a `DeployedContracts` instance which contains a `contracts_version` attribute: https://github.com/raiden-network/raiden-contracts/blob/master/raiden_contracts/contract_manager.py#L43

Not sure if this is better though, but at least atm we can have contract 0.33 but use the 024 contracts.

karlb: The python package version is the more visible part (it's in the dependencies), so I definitively want to have that info. I could add the other number, too, but I'm not sure that extra information outweights the potential confusion of showing two version numbers.

Maybe I'll open up the topic of contracts versioning again and see if we can find an easier to understand mapping between those two version numbers, so that giving one of them is usually sufficient.

palango: Makes sense.

PR #594, opened 2019-10-22
Wait for matrix to completetely start

palango:
Until now we only started matrix but didn't wait for completions of the
initialization.
This could lead to problems when a blockchain event was processed before
the sync token was set.

### DIFF
--- a/src/pathfinding_service/service.py
+++ b/src/pathfinding_service/service.py
@@ -4,10 +4,12 @@ from typing import Dict, Optional
 
 import gevent
 import structlog
+from gevent import Timeout
 from web3 import Web3
 from web3.contract import Contract
 
 from monitoring_service.constants import MAX_FILTER_INTERVAL
+from pathfinding_service.constants import MATRIX_START_TIMEOUT
 from pathfinding_service.database import PFSDatabase
 from pathfinding_service.exceptions import (
     InvalidCapacityUpdate,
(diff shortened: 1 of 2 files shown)


src/pathfinding_service/service.py
```diff
@@ -98,6 +98,8 @@ def __init__(  # pylint: disable=too-many-arguments
             message_received_callback=self.handle_message,
             address_reachability_changed_callback=self.handle_reachability_change,
         )
+        self.matrix_listener.startup_finished.wait(timeout=1)
```
palango: Any opinion on what is a reasonable timeout for this?

src/pathfinding_service/service.py
```diff
@@ -98,6 +98,8 @@ def __init__(  # pylint: disable=too-many-arguments
...
+        log.debug("Started Matrix client")
```
karlb: Ideally, creating a `PathfindingService` would not run anything until `_run` is called. Thus I would expect the `wait` below `self.matrix_listener.start()`.

If we don't want to clean that up right now, this change is the right one.

src/pathfinding_service/service.py
karlb: As long as we don't get a timeout it's high enough and since it is only a timeout and not a fixed wait, it doesn't hurt to set it to a few more seconds. Everything between 1 and 10 seconds seems reasonable.

src/pathfinding_service/service.py
palango: You're right, it's still morning.

src/pathfinding_service/service.py
palango: Fixed.

PR #604, opened 2019-10-24
Trying to run CI on contracts 0.33.3 (don't merge)

pirapira:
This commit uses some commits in github.com/pirapira
to see how CI runs.


pirapira:
Before trying to do this, I should try to update `raiden` up to `raiden:develop` now.

PR #606, opened 2019-10-24
Force adding the user to the known list

rakanalh:
When a channel is created, the PFS service calls `follow_address_presence` with `refresh=True` to:
1. populate the user ID
2. refresh presence

However, populating the user ID would check if that address is known before the transport servers are queried for the user ID. On restart, that list of known addresses is empty so the check would fail and we would end up not registering the user ID. Therefore, this change forces the addition even though the address is unknown to the PFS when the channel is opened.

### DIFF
--- a/src/raiden_libs/matrix.py
+++ b/src/raiden_libs/matrix.py
@@ -208,7 +208,7 @@ class MatrixListener(gevent.Greenlet):
             self.user_manager.add_address(address)
 
             if refresh:
-                self.user_manager.populate_userids_for_address(address)
+                self.user_manager.populate_userids_for_address(address, force=True)
                 self.user_manager.refresh_address_presence(address)
 
             log.debug(


palango:
I think we came up with the current solution because @ulope told us it should be fine and to improve performance.

During startup of the `PathfindingService` we load all channels from the DB and registers all participants for user tracking. At this point however, the matrix client hasn't been started yet.

While the tracking doesn't pull updates from the server, it at least registers all addresses in the `UserManager`:
```py
    def follow_address_presence(self, address: Address, refresh: bool = False) -> None:
        if self.user_manager:
            self.user_manager.add_address(address)

            if refresh:
                self.user_manager.populate_userids_for_address(address)
                self.user_manager.refresh_address_presence(address)
```

If I remember correctly, this information is then used during startup of the matrix client to update all presences. @err508 can you confirm this?

If the above is right, this change shouldn't be necessary.

rakanalh:
> During startup of the PathfindingService we load all channels from the DB and registers all participants for user tracking. At this point however, the matrix client hasn't been started yet.
> If I remember correctly, this information is then used during startup of the matrix client to update all presences. @err508 can you confirm this?

The change uses force because:
1. On startup `address_to_userids` is empty
2. `populate_userids_for_address` would check the existence of an address in an empty list and skip fetching the userids altogether in this case.
3. `refresh_address_presence` would only fetch presence of users whose ids are in the aforementioned list.

rakanalh:
> If the above is right, this change shouldn't be necessary.

I urge you to check

palango:
@err508 Can you comment here

err508:
I think this change is necessary, I was also looked into that part of the code when trying debug why we didn't see the correct presence in the PFS. 

> If I remember correctly, this information is then used during startup of the matrix client to update all presences. @err508 can you confirm this?

What happens if the address gets added in `_address_to_userids`, is that it gets tracked by the `_presence_listener`, nothing else.  This will only update the presence, the next time we receive a presence update event from the server, which iirc only happens if it changes, e.g. a user goes offline or comes online. The `refresh_address_presence` call in the next line will also only use the user_ids that we already know and already track.

palango:
Ok, then what happens if `follow_address_presence` is called without the matrix client being started yet?

PR #607, opened 2019-10-24
Upgrading raiden-contracts to 0.33.3

pirapira:
This switches the ServiceRegistries to the correct instances that can be deprecated.

Also a bug-fix enables deployments on private networks.


palango:
We need to remember to remove the service DBs after this got merged.

pirapira:
@palango are the DBs in the server's filesystem?

palango:
Yes, we can do this once this is merged.

Issue #601, opened 2019-10-24
MS should join all monitoring rooms on all available transport servers.


rakanalh:
## Description

Having that we found out we could potentially have a set of Raiden nodes joining different global rooms (specifically in this case `monitoring` rooms) on different servers. MS should join all these global rooms on all available servers to be able to receive balance proof updates.

rakanalh:
Handled in PR: https://github.com/raiden-network/raiden-services/pull/609

PR #609, opened 2019-10-24
Changes join_global_room to join_global_rooms.

### LINKED ISSUE #600: PFS should join all path finding rooms on all available transport servers.
## Description

Having that we found out we could potentially have a set of Raiden nodes joining different global rooms (specifically in this case `path_finding` rooms) on different servers. The PFS should join all these global rooms on all available servers to:

1. Be able to receive presence updates
2. Be able to receive balance proof updates.

err508:
The MS/PFS iterates through all available servers and joins all PFS/MS rooms.
If no rooms are found, it creates the rooms.
Fixes #600, #601

### DIFF
--- a/src/raiden_libs/matrix.py
+++ b/src/raiden_libs/matrix.py
@@ -1,6 +1,7 @@
 import sys
 from datetime import datetime, timedelta
-from typing import Any, Callable, Dict, Iterable, List, Optional, Union
+from typing import Any, Callable, Dict, Iterable, List, Optional, Sequence
+from urllib.parse import urlparse
 
 import gevent
 import structlog
@@ -17,11 +18,10 @@ from monitoring_service.constants import (
 from raiden.constants import Environment
 from raiden.exceptions import SerializationError, TransportError
 from raiden.messages.abstract import Message, SignedMessage
-from raiden.network.transport.matrix.client import Room
+from raiden.network.transport.matrix.client import GMatrixClient, Room
 from raiden.network.transport.matrix.utils import (
     AddressReachability,
     UserAddressManager,
-    join_global_room,
     login_or_register,
     make_client,
     make_room_alias,
@@ -147,13 +147,13 @@ class MatrixListener(gevent.Greenlet):
             http_retry_timeout=40,
             http_retry_delay=matrix_http_retry_delay,
         )
-        self.broadcast_room: Optional[Room] = None
+        self.broadcast_rooms: List[Room] = []
         self.user_manager: Optional[UserAddressManager] = None
 
         if address_reachability_changed_callback is not None:
             self.user_manager = UserAddressManager(
                 client=self.client,
-                get_user_callable=self._get_user,
+                get_user_callable=self._get_user_from_user_id,
                 address_reachability_changed_callback=address_reachability_changed_callback,
             )
 
@@ -191,14 +191,13 @@ class MatrixListener(gevent.Greenlet):
             raise ConnectionError("Could not login/register to matrix.")
 
         try:
-            room_name = make_room_alias(self.chain_id, self.service_room_suffix)
-            self.broadcast_room = join_global_room(
-                client=self.client, name=room_name, servers=self.available_servers
-            )
+            self.join_global_rooms(client=self.client, available_servers=self.available_servers)
         except (MatrixRequestError, TransportError):
             raise ConnectionError("Could not join monitoring broadcasting room.")
 
-        self.broadcast_room.add_listener(self._handle_message, "m.room.message")
+        # Add listener for global rooms
+        for broadcast_room in self.broadcast_rooms:
+            broadcast_room.add_listener(self._handle_message, "m.room.message")
 
         # Signal that startup is finished
         self.startup_finished.set()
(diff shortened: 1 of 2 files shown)


src/raiden_libs/matrix.py
```diff
@@ -274,3 +272,51 @@ def _handle_message(self, room: Any, event: dict) -> bool:
             self.message_received_callback(message)
 
         return True
+
+    def join_global_rooms(
```
palango: Currently most of these functions reside in `raiden.network.transport.utils`. I'd put this one there as well and add some tests.

src/raiden_libs/matrix.py
```diff
@@ -274,3 +272,51 @@ def _handle_message(self, room: Any, event: dict) -> bool:
...
+        """
+        room_alias_prefix = make_room_alias(self.chain_id, self.service_room_suffix)
+
+        parsed_servers = [
+            urlparse(s).netloc for s in available_servers if urlparse(s).netloc not in {None, ""}
+        ]
+
+        for server in parsed_servers:
+            room_alias_full = f"#{room_alias_prefix}:{server}"
+            log.debug("Trying to join pathfinding room", room_alias_full=room_alias_full)
```
palango: This might also be monitoring room, so we need to adjust the message.

src/raiden_libs/matrix.py
```diff
@@ -274,3 +272,51 @@ def _handle_message(self, room: Any, event: dict) -> bool:
...
+                    try:
+                        global_room = client.create_room(room_alias_full, is_public=True)
+                        log.debug("Created pathfinding room", room=broadcast_room)
+                        self.broadcast_rooms.append(global_room)
+                    except MatrixRequestError:
+                        log.debug(
+                            "Could neither join nor create a global room",
+                            room_alias_full=room_alias_full,
+                        )
+                        TransportError("Could neither join nor create a global room")
```
palango: Why create this and then raise it below, can't we do this in one line?

src/raiden_libs/matrix.py
```diff
@@ -221,18 +220,17 @@ def follow_address_presence(self, address: Address, refresh: bool = False) -> No
...
-
-            # if handed a User instance with displayname set, update the discovery room cache
-            if getattr(user, "displayname", None):
-                assert isinstance(user, User)
-                duser.displayname = user.displayname
-            user = duser
-        elif not isinstance(user, User):
+        for broadcast_room in self.broadcast_rooms:
+            if user_id in broadcast_room._members:  # pylint: disable=protected-access
+                duser: User = broadcast_room._members[user_id]  # pylint: disable=protected-access
```
palango: nit: can we improve the name

src/raiden_libs/matrix.py
```diff
@@ -221,18 +220,17 @@ def follow_address_presence(self, address: Address, refresh: bool = False) -> No
...
-            duser: User = self.broadcast_room._members[user_id]  # pylint: disable=protected-access
-
-            # if handed a User instance with displayname set, update the discovery room cache
-            if getattr(user, "displayname", None):
-                assert isinstance(user, User)
-                duser.displayname = user.displayname
-            user = duser
-        elif not isinstance(user, User):
+        for broadcast_room in self.broadcast_rooms:
+            if user_id in broadcast_room._members:  # pylint: disable=protected-access
```
LefterisJP: Design question. Why are you guys accessing a private attribute of the Matrix `Room` class? They got a function for that: https://github.com/matrix-org/matrix-python-sdk/blob/a469d531b6879c5687718937a0bfcb443155ce43/matrix_client/room.py#L480

src/raiden_libs/matrix.py
```diff
@@ -274,3 +272,51 @@ def _handle_message(self, room: Any, event: dict) -> bool:
             self.message_received_callback(message)
 
         return True
+
+    def join_global_rooms(
+        self, client: GMatrixClient, available_servers: Sequence[str] = ()
+    ) -> None:
+        """Join or create a global public room with given name on all available servers.
+        If can't, create a public room with the name on each server.
```
LefterisJP: nitpick: If can't -> If not possible  ... or If you can't. Also "with the name" to "with the given name"

src/raiden_libs/matrix.py
LefterisJP: +1 ... also since the log string and the exception string is the same perhaps add it in a variable and stuff it into both?

Dominik1999:
when we can merge this, I will trigger a scenario run. Then we can use the weekend to let the scenarios run 5-6 times and if, inshalla, everything is green we produce another RC.

LefterisJP:
@Dominik1999 We got some more open bug/issues that are not yet figured out. We should probably hold the RC till these bugs are figured out?

Dominik1999:
sure, let's discuss in standup.

src/raiden_libs/matrix.py
err508: I think it is confusing to have `join_global_room` and `join_global_rooms` in `raiden.network.transport.utils`. As we would like to keep both and `join_global_rooms` is specific for the services, I would leave it here.

palango: My point is that the services have minimal matrix related code and no integration tests for matrix.
Therefore, I think it's better to move this to the other utils (and change the name to clarify) and add some tests with the others.

src/raiden_libs/matrix.py
err508: This  is actually a little quirky, as we want to do a lookup of User by user_id: get_joined_members gives us a list of Users, where _members is already a user_id: User dict. We could add a function to expose the lookup.

src/raiden_libs/matrix.py
err508: The function is supposed to be used _only_ in the services, just like the matrix_listener it belongs to. Is it okay If we just add the tests in raiden.matrix like with the rest of the services stuff but leave the function where it is used?

src/raiden_libs/matrix.py
LefterisJP: Please do that.

And still as you already know accessing private attributes of a class is really bad practise and will probably come bite us in the future. On the other hand we are using a deprecated library so ... yeah .. :man_shrugging:

palango:
I'm Ok with merging that now to see if it solves the presence issues and unblock other people. but there are still two issues that should be solved:
- https://github.com/raiden-network/raiden-services/pull/609#discussion_r338977099
- https://github.com/raiden-network/raiden-services/pull/609#discussion_r338974733

@err508 Can you create issues for those.

Also as written in RC, adding commit for review items instead of amending all changes make re-review a lot easier.

Dominik1999:
@err508 did you create issues for the comments @palango had above? Or are those issues already solved and merged somehow?

palango:
@err508 :

> @err508 did you create issues for the comments @palango had above? Or are those issues already solved and merged somehow?

err508:
@palango 
I added it to https://github.com/raiden-network/raiden/issues/5149 but for the private attributes I would not change that, as we should move to a new transport client soon.

src/raiden_libs/matrix.py
palango: And here we are: https://github.com/raiden-network/raiden-services/issues/630#issuecomment-554373344

PR #615, opened 2019-11-07
Increase matrix timeout

palango:
This is a short term fix, in the future we should parallelize in
`join_global_rooms`.


src/raiden_libs/constants.py
```diff
@@ -3,4 +3,4 @@
 UDC_SECURITY_MARGIN_FACTOR_MS: float = 1.1
 UDC_SECURITY_MARGIN_FACTOR_PFS: float = 2
 
-MATRIX_START_TIMEOUT = 10
+MATRIX_START_TIMEOUT = 30
```
Dominik1999: can you add a comment, that this is seconds?

Dominik1999: and do we have a test for that?

palango: A test for what?

Issue #633, opened 2019-11-19
Allow passing the URL of the matrix server list


karlb:
Our test environment should not use the same matrix servers as our development environment. To make this possible, we should allow taking either a list of servers or the URL of a server list file as an argument. I'd prefer the former because it is easier to use. Would that cover all use cases?

PR #635, opened 2019-11-19
Allow setting the matrix servers via CLI

karlb:
Closes https://github.com/raiden-network/raiden-services/issues/633


src/pathfinding_service/cli.py
```diff
@@ -59,6 +59,12 @@
     type=str,
     help="Place for a personal message to the customers",
 )
+@click.option(
+    "--matrix-server",
+    type=str,
+    multiple=True,
+    help="Use this matrix server instead of the default ones. Include protocol in argument.",
```
palango: Should we mention that multiple values are possible?

karlb: I thought about it but also wanted to keep the help text concise. Since we don't even know a single case where we want to pass multiple servers, I removed it. Not sure if that was the right call. The cli help does not indicate the option of repeating that parameter, unfortunately.

Issue #636, opened 2019-11-19
ValueError: x_list must be in strictly ascending order!


palango:
## Problem Definition

In the BF1 scenario that @manuelwedler ran, some transfer failed. In the PFS I found this logs:
```
Exception on /api/v1/0xDa1fBc048f503635950058953f5c60FC1F564ee6/paths [POST]
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/flask/app.py", line 1832, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.7/site-packages/flask/app.py", line 1818, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/local/lib/python3.7/site-packages/flask_restful/__init__.py", line 458, in wrapper
    resp = resource(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/flask/views.py", line 88, in view
    return self.dispatch_request(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/flask_restful/__init__.py", line 573, in dispatch_request
    resp = meth(*args, **kwargs)
  File "/services/src/pathfinding_service/api.py", line 142, in post
    **optional_args,
  File "/services/src/pathfinding_service/model/token_network.py", line 437, in get_paths
    fee_penalty=fee_penalty,
  File "/services/src/pathfinding_service/model/token_network.py", line 384, in _get_single_path
    Path(self.G, nodes, value, address_to_reachability) for nodes in all_paths
  File "/services/src/pathfinding_service/model/token_network.py", line 382, in <genexpr>
    p
  File "/services/src/pathfinding_service/model/token_network.py", line 384, in <genexpr>
    Path(self.G, nodes, value, address_to_reachability) for nodes in all_paths
  File "/services/src/pathfinding_service/model/token_network.py", line 81, in __init__
    self._calculate_fees()
  File "/services/src/pathfinding_service/model/token_network.py", line 97, in _calculate_fees
    receivable_amount=view_in.capacity,
  File "/usr/local/lib/python3.7/site-packages/raiden/tests/utils/mediation_fees.py", line 94, in get_amount_with_fees
    cap_fees=schedule_in.cap_fees,
  File "/usr/local/lib/python3.7/site-packages/raiden/transfer/mediated_transfer/mediation_fee.py", line 225, in mediation_fee_backwards_func
    cap_fees=cap_fees,
[...]

PR #646, opened 2019-11-25
Add regression test for `x_list must be in str...`

karlb:
Regression test for
https://github.com/raiden-network/raiden-services/issues/636.

### DIFF
--- a/tests/pathfinding/test_token_network.py
+++ b/tests/pathfinding/test_token_network.py
@@ -113,9 +113,10 @@ def test_path_without_capacity(token_network_model: TokenNetwork, addresses: Lis
     )
 
     token_network_model.G[addresses[1]][addresses[2]]["view"].channel.capacity1 = 100
-    Path(
+    path = Path(
         G=token_network_model.G,
         nodes=[addresses[0], addresses[1], addresses[2]],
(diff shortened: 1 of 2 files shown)


tests/pathfinding/test_token_network.py
```diff
@@ -92,3 +92,30 @@ def test_graph_pruning():
...
+    )
+    token_network_model.handle_channel_opened_event(
+        channel_identifier=ChannelID(2),
+        participant1=addresses[1],
+        participant2=addresses[2],
+        settle_timeout=15,
+    )
+
+    token_network_model.G[addresses[1]][addresses[2]]["view"].channel.capacity1 = 100
+    Path(
```
palango: Maybe we should also check that `Path.valid is True`? I'm scared that we simplify the constructor in the future this this becomes a useless test.

karlb: The path is not valid. It can't receive a single token. But now it's invalid without throwing an exception.

palango: Yeah, then I mean a check that `Path.valid is False`. My point is that we check more than just running thew constructor, because we might simplify/change that soon.

Issue #643, opened 2019-11-25
Add profiling option


karlb:
We should be able to create profiling data both for PFSs running on services-dev.raiden.network and inside tests. Due to using gevent, we can't use the normal profiling tools.

- [x] Find out what Raiden is using for profiling, if that is not suitable for some reason, investigate https://pypi.org/project/GreenletProfiler/
- [x] Add option to save profile for running PFS (exclude startup from profile?)
- [x] Add a way to profile specific tests

Probably also useful:
- [x] Add option to save profile for running MS
- [x] Add option to save profile for running RC
- [ ] Log warnings when services reach high load
- [x] Profile one local PFS while running a scenario (see https://github.com/raiden-network/raiden-services/issues/651)

Sp 5

PR #653, opened 2019-11-26
Add profiling options to services and tests

karlb:
This will create stack profiling data for use with https://github.com/brendangregg/FlameGraph.

Will only work reliably after https://github.com/raiden-network/raiden-services/pull/652 is merged.

Closes https://github.com/raiden-network/raiden-services/issues/643

### DIFF
--- a/src/raiden_libs/cli.py
+++ b/src/raiden_libs/cli.py
@@ -1,3 +1,4 @@
+import datetime
 import json
 import logging
 import os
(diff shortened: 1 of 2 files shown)


src/raiden_libs/cli.py
```diff
@@ -60,6 +61,22 @@ def validate_address(
     return to_canonical_address(value)
 
 
+def start_profiler(output_dir: Optional[str]) -> Optional[Any]:
+    """ Start and return a profiler is `output_dir` is not None
+    """
```
palango: Can you join that with the last line?

src/raiden_libs/cli.py
```diff
@@ -60,6 +61,22 @@ def validate_address(
...
+    if not output_dir:
+        return None
+
+    from raiden.utils.profiling.sampler import SignalSampler, FlameGraphCollector
+
+    os.makedirs(output_dir, exist_ok=True)
+    now = datetime.datetime.now()
+    stack_path = os.path.join(output_dir, f"{now:%Y%m%d_%H%M}_stack.data")
+    stack_stream = open(stack_path, "w")
```
palango: I think we should close the handle somewhere

src/raiden_libs/cli.py
```diff
@@ -102,12 +119,27 @@ def decorator(func: Callable) -> Callable:
                     default=False,
                     help="Enable or disable logging in JSON format",
                 ),
+                click.option(
+                    "--flamegraph",
+                    help="Directory in which to save a stack profile to produce flame graphs.",
```
palango: Does this hurt performance? If so, should we hide or mention that it's not recommended?

karlb: It surely does, although I don't know how much. But option help texts are not made for long explanations, so I would not add it.

src/raiden_libs/cli.py
karlb: That is done by `FlameGraphCollector.stop`.

src/raiden_libs/cli.py
karlb: Done.

PR #656, opened 2019-11-26
PFS performance logging and tweaks

karlb:
This is not solving a specific problem, but the changes are simple and should be an improvement.

### DIFF
--- a/src/pathfinding_service/service.py
+++ b/src/pathfinding_service/service.py
@@ -169,6 +169,7 @@ class PathfindingService(gevent.Greenlet):
         before_process = time.time()
         for event in events:
             self.handle_event(event)
+            gevent.idle()  # Allow answering requests in between events
 
         if events:
             log.info(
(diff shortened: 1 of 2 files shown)


src/pathfinding_service/service.py
```diff
@@ -143,23 +146,41 @@ def _run(self) -> None:  # pylint: disable=method-hidden
...
-            gevent.joinall(
-                set([self.matrix_listener]), timeout=self._poll_interval, raise_error=True
-            )
+            # Let tests waiting for this event know that we're done with processing
+            self.updated.set()
+            self.updated.clear()
+
+            # Sleep, then collect errors from greenlets
+            gevent.sleep(self._poll_interval)
+            gevent.joinall(set([self.matrix_listener]), timeout=0, raise_error=True)
```
palango: What does `timeout=0`  imply?

karlb: We don't actually wait for all greenlets (the matrix_listener won't finish unless it has an error), we just collect all greenlets that are done or threw an exception right now and then continue.

It's similar a similar concept to `gevent.sleep(0)`.

PR #657, opened 2019-11-26
Revert join all rooms

### LINKED ISSUE #634: Revert joining of different broadcast rooms
During the investigation of the presence problem @ulope unified the broadcasting rooms (meaning all aliases link tot he same underlying room)

Therefore, we can revert https://github.com/raiden-network/raiden-services/pull/609 which should also fix the problems described in https://github.com/raiden-network/raiden-services/issues/630#issuecomment-554373344

[SP 3]

palango:
Fixes #634, https://github.com/raiden-network/raiden-services/issues/630

This reverts https://github.com/raiden-network/raiden-services/pull/609 as we now have one alias for the same broadcast room on all servers.

### DIFF
--- a/src/raiden_libs/matrix.py
+++ b/src/raiden_libs/matrix.py
@@ -1,7 +1,6 @@
 import sys
 from datetime import datetime, timedelta
-from typing import Any, Callable, Dict, Iterable, List, Optional, Sequence
-from urllib.parse import urlparse
+from typing import Any, Callable, Dict, Iterable, List, Optional, Union
 
 import gevent
 import structlog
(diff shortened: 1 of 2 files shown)


src/raiden_libs/matrix.py
```diff
@@ -225,12 +229,12 @@ def follow_address_presence(self, address: Address, refresh: bool = False) -> No
 
     def _get_user_from_user_id(self, user_id: str) -> User:
         """Creates an User from an user_id, if none, or fetch a cached User """
-        for broadcast_room in self.broadcast_rooms:
-            if user_id in broadcast_room._members:  # pylint: disable=protected-access
-                user: User = broadcast_room._members[user_id]  # pylint: disable=protected-access
-                break
+        if (  # pylint: disable=protected-access
```
hackaugusto: It's a better idea to not make `_broadcast_room` a private member.

hackaugusto: Also, shouldn't this be an `assert self._broadcast_room`?

palango: > It's a better idea to not make `_broadcast_room` a private member.

I think this is about accessing `Room._members`, not `_broadcast_room`.

palango: Regarding the assert, I changed it now. But I'm not sure about it.

palango: @ulope Can you throw a look at this and check whether it makes sense?

ulope: Yeah looks ok.

Maybe it would make that you also start using the `DisplayNameCache`, although using the broacasting room members as the "cache" should already give you 90% of the benefit.

PR #666, opened 2019-11-28
Rework path validity checking

palango:
This reworks the path validity checking a bit. Now it checks first if
the path has enough capacity for the payment without fees before
calculating fees.
This should improve speed and also give clearer errors when payments
fail because of missing capacity.

### DIFF
--- a/src/pathfinding_service/model/token_network.py
+++ b/src/pathfinding_service/model/token_network.py
@@ -79,11 +79,17 @@ class Path:
         self.value = value
         self.address_to_reachability = address_to_reachability
         self.fees: List[FeeAmount] = []
-        self._calculate_fees()
+        self._is_valid = self._check_validity()
 
-        log.debug("Creating Path object", nodes=nodes, is_valid=self.is_valid, fees=self.fees)
+        log.debug("Creating Path object", nodes=nodes, is_valid=self._is_valid, fees=self.fees)
 
-    def _calculate_fees(self) -> None:
+    def _calculate_fees(self) -> bool:
+        """ Calcluates fees backwards for this path.
+
+        Note: Stateful, should only be called once!
+
+        Returns ``False``, if the fee calculation cannot be done.
+        """
         total = PaymentWithFeeAmount(self.value)
         for prev_node, mediator, next_node in reversed(list(window(self.nodes, 3))):
             view_in: ChannelView = self.G[prev_node][mediator]["view"]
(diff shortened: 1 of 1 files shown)


src/pathfinding_service/model/token_network.py
```diff
@@ -198,8 +183,51 @@ def is_valid(self) -> bool:
                 )
                 return False
 
+        # check capacity with fees
+        valid_fees = self._calculate_fees()
+        if not valid_fees:
+            return False
+
+        fees = self.fees + [FeeAmount(0)]  # The hop to the target does not incur mediation fees
+        for edge, fee in zip(edges, fees):
```
karlb: This is already implicitly done in `self._calculate_fees`, isn't it?

src/pathfinding_service/model/token_network.py
```diff
@@ -198,8 +183,51 @@ def is_valid(self) -> bool:
...
+    def edge_attrs(self) -> Iterable[dict]:
+        return (self.G[node1][node2] for node1, node2 in zip(self.nodes[:-1], self.nodes[1:]))
+
+    def to_dict(self) -> dict:
+        return dict(
+            path=[to_checksum_address(node) for node in self.nodes], estimated_fee=sum(self.fees),
+        )
+
+    @property
+    def is_valid(self) -> bool:
```
karlb: We don't need this as a property, anymore. We can just rename `_is_valid` to `is_valid`.

src/pathfinding_service/model/token_network.py
```diff
@@ -79,11 +79,17 @@ def __init__(
...
         self.address_to_reachability = address_to_reachability
         self.fees: List[FeeAmount] = []
-        self._calculate_fees()
+        self._is_valid = self._check_validity()
 
-        log.debug("Creating Path object", nodes=nodes, is_valid=self.is_valid, fees=self.fees)
+        log.debug("Creating Path object", nodes=nodes, is_valid=self._is_valid, fees=self.fees)
 
-    def _calculate_fees(self) -> None:
+    def _calculate_fees(self) -> bool:
```
karlb: If we made this
```suggestion
    def _calculate_fees(self) -> Optional[FeeAmount]:
```
would not modify any state, anymore.

src/pathfinding_service/model/token_network.py
```diff
@@ -124,62 +130,41 @@ def _calculate_fees(self) -> None:
...
-    def to_dict(self) -> dict:
-        return dict(
-            path=[to_checksum_address(node) for node in self.nodes], estimated_fee=sum(self.fees),
-        )
+        return True
 
-    @property
-    def is_valid(self) -> bool:
-        """ Check capacity and settle timeout
+    def _check_validity(self) -> bool:
```
karlb: Similarly, we can remove the side effect here:
```suggestion
    def _calc_fees_if_valid(self) -> Optional[FeeAmount]:
```

src/pathfinding_service/model/token_network.py
```diff
@@ -190,22 +190,23 @@ def _check_validity(self) -> bool:
...
-        valid_fees = self._calculate_fees()
-        if not valid_fees:
+        fees = self._calculate_fees()
+        if not fees:
             log.debug("Path invalid because of invalid fee calculation",)
-            return False
+            return None
 
-        return True
+        return fees
```
karlb: If we moved the `log` to L132, we could directly do
```python
return self._calculate_fees()
```
This would also allow adding more details to the log, if necessary.

PR #674, opened 2019-12-02
Provide specific error message for common errors

karlb:
This should make it easier to known what's going on when we get "no
route found" errors.


karlb:
@hackaugusto reviewed it on my machine.

PR #686, opened 2019-12-10
Save token network graph on NoRouteFound errors

karlb:
This should make it easier to debug cases where we should not get
NoRouteFound.

### DIFF
--- a/src/pathfinding_service/api.py
+++ b/src/pathfinding_service/api.py
@@ -1,4 +1,5 @@
 import collections
+import os
 from dataclasses import dataclass, field
 from datetime import datetime
 from typing import Any, ClassVar, Dict, List, Optional, Tuple, Type, TypeVar, cast
@@ -117,6 +118,13 @@ class PathsResource(PathfinderResource):
         self.debug_mode = debug_mode
 
     def post(self, token_network_address: str) -> Tuple[dict, int]:
+        graph_dir: Optional[str]
+        if self.debug_mode:
+            graph_dir = "/tmp/pfs_graphs"  # TODO: make this configurable
+            os.makedirs(graph_dir, exist_ok=True)
+        else:
+            graph_dir = None
+
         token_network = self._validate_token_network_argument(token_network_address)
         path_req = self._parse_post(PathRequest)
         process_payment(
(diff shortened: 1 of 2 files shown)


src/pathfinding_service/model/token_network.py
```diff
@@ -458,6 +459,40 @@ def check_path_request_errors(
...
+                        "{} [{}]\n".format(
+                            to_checksum_address(node),
+                            ""
+                            if address_to_reachability[node] == AddressReachability.REACHABLE
+                            else "style=filled",
+                        )
+                    )
+            for edge_attr in self.G.edges.values():
+                f.write(
+                    '"{}" -> "{}" [label={}]\n'.format(
```
palango: Does graphviz handle bidirectional graphs?

src/pathfinding_service/api.py
```diff
@@ -117,6 +118,13 @@ def __init__(self, debug_mode: bool, **kwargs: Any):
         self.debug_mode = debug_mode
 
     def post(self, token_network_address: str) -> Tuple[dict, int]:
+        graph_dir: Optional[str]
+        if self.debug_mode:
+            graph_dir = "/tmp/pfs_graphs"  # TODO: make this configurable
```
palango: Do we need to mount `/tmp` in the docer containers?

src/pathfinding_service/model/token_network.py
karlb: I've modeled it as a directed graph with a pair of edges between the nodes and that works.

src/pathfinding_service/api.py
karlb: I don't think we should do that. We should make that path configurable and do a bind mount for it. Until then we'll have fetch it from inside the container. Something like `docker run "cat /tmp/pfs_graphs/*.log"` should do it.

karlb: We could also bind `/tmp/pfs_graphs`, but really think binding all of `/tmp` is a bad idea.

palango: Ah cool, I didn't know about `docker run "cat /tmp/pfs_graphs/*.log"`.

konradkonrad:
What's up with this PR @karlb ? Do you still want to get this merged?

karlb:
> What's up with this PR @karlb ? Do you still want to get this merged?

Looks like all path problems could be explained with "presence is messed up", so far. So this seems to be not necessary. I'll just close it.

PR #691, opened 2019-12-18
Show better errors when there is no available path in the graph

palango:


tests/pathfinding/test_token_network.py
```diff
@@ -167,10 +167,26 @@ def test_check_path_request_errors(token_network_model, addresses):
...
     token_network_model.G.edges[a[0], a[1]]["view"].capacity = 100
+
     assert token_network_model.check_path_request_errors(a[0], a[2], 100, reachability).startswith(
         "Target does not have a channel with sufficient capacity"
     )
     token_network_model.G.edges[a[1], a[2]]["view"].capacity = 100
 
     # Must return `None` when no errors could be found
-    assert token_network_model.check_path_request_errors(a[0], a[2], 100, reachability) is None
+    assert token_network_model.check_path_request_errors(a[0], a[2], 100, reachability,) is None
```
karlb: Where does the trailing comma come from?

src/pathfinding_service/model/token_network.py
```diff
@@ -456,6 +456,11 @@ def check_path_request_errors(
                 value,
             )
 
+        try:
+            next(nx.shortest_simple_paths(G=self.G, source=source, target=target, weight="weight"))
```
karlb: No need to include the `weight`, here.

tests/pathfinding/test_token_network.py
```diff
@@ -167,10 +167,26 @@ def test_check_path_request_errors(token_network_model, addresses):
...
+
+    # Check error when there is no route in the graph
+    token_network_model.handle_channel_opened_event(
+        channel_identifier=ChannelID(2),
+        participant1=a[3],
+        participant2=a[4],
+        settle_timeout=BlockTimeout(15),
+    )
```
karlb: Can't we get the same error just by doing
```
reachability[a[1]] = AddressReachability.UNREACHABLE
```
instead of creating an additional channel?

tests/pathfinding/test_token_network.py
palango: Before I added another parameter and that required reformatting, and then I forgot to remove it.

PR #692, opened 2020-01-09
Collect addresses before updating presence

palango:
As a base for discussion.

My hope is that this makes startup faster, by reducing the calls to `follow_address_presence`.

However, it also improves the latency from channel opening to a usable channel, but in general (the synced state) this should be negligible.


karlb:
This does not bring sufficient benefit while make the functions less pleasant to read.

PR #694, opened 2020-01-09
Fix lint

palango:
Necessary for following PRs to be built.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -81,7 +81,8 @@ disable=
         ungrouped-imports,
         # checks handled by flake8
         line-too-long,
-        wildcard-import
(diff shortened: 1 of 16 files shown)


pylintrc
```diff
@@ -259,6 +260,9 @@ good-names=i,
            p2,
            G,   # graph (networkx uses G for graph as mathematicians do)
            T,   # generic TypeVar used by mypy
+           to,
+           Schema,
+           id,
```
karlb: Do we really want this? `id` is also a builtin.

palango: Ok, I renamed it to `uuid`

PR #695, opened 2020-01-09
Fix channel delete

palango:
Merge after #694 

This fixes a problem in `DB.delete_channel` and adds add test for it.

### DIFF
--- a/src/pathfinding_service/database.py
+++ b/src/pathfinding_service/database.py
@@ -190,10 +190,11 @@ class PFSDatabase(BaseDatabase):
     def delete_channel(
         self, token_network_address: TokenNetworkAddress, channel_id: ChannelID
     ) -> None:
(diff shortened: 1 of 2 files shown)


src/pathfinding_service/database.py
```diff
@@ -190,10 +190,11 @@ def get_channels(self) -> Iterator[Channel]:
     def delete_channel(
         self, token_network_address: TokenNetworkAddress, channel_id: ChannelID
     ) -> None:
-        self.conn.execute(
+        cursor = self.conn.execute(
             "DELETE FROM channel WHERE token_network_address = ? AND channel_id = ?",
-            [token_network_address, channel_id],
+            [to_checksum_address(token_network_address), hex256(channel_id)],
```
rakanalh: This seems to be the fix

PR #697, opened 2020-01-10
Optimize channel deletion

palango:
Remove loading a channel just to delete it then. 

Instead, `delete_channel` expects to delete 0 or 1 row now.

### DIFF
--- a/src/pathfinding_service/service.py
+++ b/src/pathfinding_service/service.py
@@ -257,16 +257,8 @@ class PathfindingService(gevent.Greenlet):
 
         log.info("Received ChannelClosed event", event_=event)
 
-        channel = self.database.get_channel(event.token_network_address, event.channel_identifier)
-        if channel:
-            token_network.handle_channel_closed_event(event.channel_identifier)
-            self.database.delete_channel(event.token_network_address, event.channel_identifier)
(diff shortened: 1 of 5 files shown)


src/pathfinding_service/service.py
```diff
@@ -257,16 +257,8 @@ def handle_channel_closed(self, event: ReceiveChannelClosedEvent) -> None:
...
-            token_network.handle_channel_closed_event(event.channel_identifier)
-            self.database.delete_channel(event.token_network_address, event.channel_identifier)
-        else:
-            log.error(
-                "Received ChannelClosed event for unknown channel",
-                token_network_address=event.token_network_address,
-                channel_identifier=event.channel_identifier,
-            )
+        token_network.handle_channel_closed_event(event.channel_identifier)
+        self.database.delete_channel(event.token_network_address, event.channel_identifier)
```
karlb: I was thinking of
```python
if self.database.delete_channel(event.token_network_address, event.channel_identifier):
    token_network.handle_channel_closed_event(event.channel_identifier)
```
with the delete returning the row count and no except in `TokenNetwork.handle_channel_closed_event`. But this works fine, too.

PR #702, opened 2020-01-14
Using persistent data for Path debug endpoint

hackaugusto:
Fixes https://github.com/raiden-network/raiden/issues/5503

If the PFS is restarted while a scenario is running the data of the PFS
requests will be lost, which has led to scenario failures. This commit
changes the debug endpoint to use data persisted in the database, so
that restarts will no lead to a problem.

Note: This is relying on the feedback data, which does not contain all
the requests. Specifically, requests which failed with a validation
error are not included in the table, therefore are still successtible to
loss of data.

### DIFF
--- a/src/pathfinding_service/api.py
+++ b/src/pathfinding_service/api.py
@@ -45,7 +45,7 @@ from raiden_libs.marshmallow import ChecksumAddress, HexedBytes
 log = structlog.get_logger(__name__)
 T = TypeVar("T")
 # list stores max 200 last requests
-last_requests: collections.deque = collections.deque([], maxlen=200)
+last_failed_requests: collections.deque = collections.deque([], maxlen=200)
 
 
 class ApiWithErrorHandler(Api):
@@ -137,7 +137,7 @@ class PathsResource(PathfinderResource):
         if error:
             # this is for assertion via the scenario player
             if self.debug_mode:
-                last_requests.append(
+                last_failed_requests.append(
                     dict(
                         token_network_address=to_checksum_address(token_network_address),
                         source=to_checksum_address(path_req.from_),
@@ -170,7 +170,7 @@ class PathsResource(PathfinderResource):
         # this is for assertion via the scenario player
         if len(paths) == 0:
             if self.debug_mode:
-                last_requests.append(
+                last_failed_requests.append(
                     dict(
                         token_network_address=to_checksum_address(token_network_address),
                         source=to_checksum_address(path_req.from_),
@@ -182,17 +182,6 @@ class PathsResource(PathfinderResource):
                 from_=to_checksum_address(path_req.from_),
                 to=to_checksum_address(path_req.to),
                 value=path_req.value,
-            )
-
-        # this is for assertion via the scenario player
-        if self.debug_mode:
-            last_requests.append(
-                dict(
-                    token_network_address=to_checksum_address(token_network_address),
-                    source=to_checksum_address(path_req.from_),
-                    target=to_checksum_address(path_req.to),
-                    routes=paths,
-                )
             )
 
         # Create a feedback token and store it to the DB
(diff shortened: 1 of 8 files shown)


src/pathfinding_service/database.py
```diff
@@ -212,13 +212,18 @@ def get_token_networks(self) -> Iterator[TokenNetwork]:
                 token_network_address=TokenNetworkAddress(to_canonical_address(row[0]))
             )
 
-    def prepare_feedback(self, token: FeedbackToken, route: List[Address]) -> None:
+    def prepare_feedback(
+        self, token: FeedbackToken, route: List[Address], estimated_fee: int
```
palango: the type of `estimated_fee` should be `TokenAmount`

src/pathfinding_service/database.py
```diff
@@ -248,6 +253,35 @@ def update_feedback(self, token: FeedbackToken, route: List[Address], successful
 
         return updated_rows
 
+    def get_feedback_routes(
+        self, token_network_address: str, source_address: str, target_address: str = None
```
palango: The types here should be `TokenNetworkAddress` and `Address` I think.

src/pathfinding_service/schema.sql
```diff
@@ -65,7 +65,10 @@ CREATE TABLE feedback (
     token_id CHAR(32) NOT NULL,
     creation_time TIMESTAMP NOT NULL,
     token_network_address CHAR(42) NOT NULL,
+    source_address CHAR(42) NOT NULL,
+    target_address CHAR(42) NOT NULL,
     route TEXT NOT NULL,
+    estimated_fee INTEGER NOT NULL,
```
palango: int might not be big enough here, so we need to hex-encode.

src/pathfinding_service/api.py
```diff
@@ -430,6 +421,28 @@ def get(  # pylint: disable=no-self-use
                     dict(source=req["source"], target=req["target"], routes=req["routes"])
                 )
 
+        decoded_target_address: Optional[Address] = None
+        if target_address:
+            decoded_target_address = Address(to_canonical_address(target_address))
```
palango: Is `Address` needed here? `to_canonical_address` should return an `Address` already

src/pathfinding_service/api.py
```diff
@@ -430,6 +421,28 @@ def get(  # pylint: disable=no-self-use
...
+
+        feedback_routes = list(
```
palango: Seems like you don't need `list` here. The loop below should work on an iterator as well and save some memory.

src/pathfinding_service/api.py
```diff
@@ -430,6 +421,28 @@ def get(  # pylint: disable=no-self-use
...
+            self.pathfinding_service.database.get_feedback_routes(
+                TokenNetworkAddress(to_canonical_address(token_network_address)),
+                Address(to_canonical_address(source_address)),
```
palango: Same with `Address` as above

src/pathfinding_service/database.py
```diff
@@ -212,13 +212,18 @@ def get_token_networks(self) -> Iterator[TokenNetwork]:
...
+        self, token: FeedbackToken, route: List[Address], estimated_fee: TokenAmount
+    ) -> None:
         hexed_route = [to_checksum_address(e) for e in route]
         token_dict = dict(
             token_id=token.uuid.hex,
             creation_time=token.creation_time,
             token_network_address=to_checksum_address(token.token_network_address),
             route=json.dumps(hexed_route),
+            estimated_fee=hex256(estimated_fee),
+            source_address=route[0],
```
palango: These need to be wrapped in `to_checksum_address`

src/pathfinding_service/api.py
palango: Fixed

src/pathfinding_service/api.py
palango: Fixed

src/pathfinding_service/database.py
palango: Fixed

PR #707, opened 2020-01-20
Improving logging and linting

Dominik1999:
Improves logging for the "Checking IOU" step to better debug #5684

### DIFF
--- a/src/pathfinding_service/api.py
+++ b/src/pathfinding_service/api.py
@@ -226,7 +226,7 @@ def process_payment(  # pylint: disable=too-many-branches
 
     log.debug(
         "Checking IOU",
-        sender=iou.sender,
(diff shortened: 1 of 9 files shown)


src/pathfinding_service/api.py
```diff
@@ -224,7 +224,8 @@ def process_payment(  # pylint: disable=too-many-branches
     if iou is None:
         raise exceptions.MissingIOU
 
-    log.debug("Checking IOU", iou=iou)
+    log.debug("Checking IOU", sender=iou.sender, total_amount=iou.amount)
```
palango: This will still be an unformatted address, you should wrap it in `to_checksum_address`.

palango: I'd also log the remaining information

PR #709, opened 2020-01-21
Upgrade to latest raiden

palango:
After https://github.com/raiden-network/raiden/pull/5728

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-git+https://github.com/palango/raiden.git@4dca1d2f35aaa94505cf82e3a2b890bf2106b7df
+git+https://github.com/raiden-network/raiden.git@61a0b71462a71496d93d8c2ab7cd3415455a937f
 raiden-contracts==0.34
 
 structlog==19.1.0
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -1,5 +1,5 @@
-git+https://github.com/raiden-network/raiden.git@f9bbbc0aeadf2ce1a54f3ed22de948133bc461df
-raiden-contracts==0.33.3
+git+https://github.com/palango/raiden.git@4dca1d2f35aaa94505cf82e3a2b890bf2106b7df
```
Dominik1999: is the the correct link? It is not raiden-network but palango, now?

tests/libs/test_matrix.py
```diff
@@ -112,12 +112,6 @@ def test_deserialize_messages_valid_messages(request_monitoring_message):
     assert len(messages) == 2
 
 
-def test_matrix_http_retry_delay():
```
Dominik1999: we don't need that test anymore?

requirements.txt
palango: This was the only way to test this before the PR in raiden is merged, I'll change that back. Just wanted CI to run already.

tests/libs/test_matrix.py
palango: Fixed

src/raiden_libs/matrix.py
```diff
@@ -120,8 +123,8 @@ def deserialize_messages(
 def matrix_http_retry_delay() -> Iterable[float]:
     return timeout_exponential_backoff(
         DEFAULT_TRANSPORT_RETRIES_BEFORE_BACKOFF,
-        int(DEFAULT_TRANSPORT_MATRIX_RETRY_INTERVAL / 5),
```
palango: @karlb Any idea why `mypy` didn't catch this.

PR #715, opened 2020-02-05
Remove Rakan from auto assign

ulope:


.github/auto_assign.yml
```diff
@@ -7,7 +7,6 @@ addAssignees: false
 # A list of reviewers to be added to pull requests (GitHub user name)
 reviewers:
   - hackaugusto
-  - rakanalh
   - palango
   - karlb
 #  - LefterisJP
```
palango: And remove lefteris

PR #716, opened 2020-02-05
Enable `no-implicit-optional`

palango:
This makes function signatures a bit more explicit.


setup.cfg
```diff
@@ -30,7 +30,9 @@ disallow_untyped_defs = True
 warn_unused_configs = True
 warn_unused_ignores = True
 warn_unreachable = True
+warn_redundant_casts = True
```
karlb: I like this!

PR #723, opened 2020-02-07
Use `AsyncResult` to raise errors while waiting

karlb:
We were waiting on `event`s that never came when the greenlet we're
waiting for threw an exception at an unfortunate time. Because we we're
stuck in the `wait`, we did not event collect and raise the exception
from the underlying greenlet.

If should probably have split some of the changes in `matrix.py` into a
separate commit, but I don't think it's worth the effort to pull the
changes apart, now.

### DIFF
--- a/src/pathfinding_service/service.py
+++ b/src/pathfinding_service/service.py
@@ -105,7 +105,7 @@ class PathfindingService(gevent.Greenlet):
 
         self.token_networks = self._load_token_networks()
         self.updated = gevent.event.Event()  # set whenever blocks are processed
-        self.startup_finished = gevent.event.Event()
+        self.startup_finished = gevent.event.AsyncResult()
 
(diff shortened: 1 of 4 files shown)


src/pathfinding_service/service.py
```diff
@@ -105,7 +105,7 @@ def __init__(  # pylint: disable=too-many-arguments
 
         self.token_networks = self._load_token_networks()
         self.updated = gevent.event.Event()  # set whenever blocks are processed
-        self.startup_finished = gevent.event.Event()
+        self.startup_finished = gevent.event.AsyncResult()
```
palango: Why do we need an `AsyncResult` here?

karlb: Events don't have a `get` that can raise exceptions, so they don't help handling the error case.

Issue #721, opened 2020-02-06
Monitoring transactions sent late on goerli


palango:
On goerli we have quite hard conditions for the MS, the minimum settle timeout is just 40 blocks, which leaves only 8 blocks (20% of 40) for the MS to act on chain in the scenarios.

In some scenarios we saw surprising numbers, so we need to look into that. A different idea is to add a gas price strategy as we use in raiden to the MS to get the transactions to be mined faster.

PR #725, opened 2020-02-12
Trigger scheduled events earlier

palango:
Fixes https://github.com/raiden-network/raiden-services/issues/721

We used to trigger scheduled events based on the confirmed block, which lets valuable time pass. Especially in the scenarios with short settle timeouts the confirmation timeout of 5 blocks led to some flakyness.

This is possible because triggered events only rely on block number, and not on certain events that might change during a chain reorg.

### DIFF
--- a/src/monitoring_service/handlers.py
+++ b/src/monitoring_service/handlers.py
@@ -54,14 +54,11 @@ class Context:
         return self.ms_state.blockchain_state.latest_committed_block
 
     @property
-    def latest_block(self) -> BlockNumber:
-        """ The latest block. """
-        return self.web3.eth.blockNumber
-
-    @property
     def latest_confirmed_block(self) -> BlockNumber:
-        """ The latest confirmed block. """
-        return BlockNumber(self.latest_block - self.required_confirmations)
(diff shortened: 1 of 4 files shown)


src/monitoring_service/handlers.py
```diff
@@ -53,10 +53,15 @@ def latest_committed_block(self) -> BlockNumber:
         has been written to the DB. """
         return self.ms_state.blockchain_state.latest_committed_block
 
+    @property
+    def latest_block(self) -> BlockNumber:
```
karlb: Should we call this `latest_unconfirmed_block` to make sure that it is not misunderstood?

If `self.web3.eth.blockNumber` does an RPC call, I am against making it a property, since we expect attributes to be fast, non-blocking, non-switching.

src/monitoring_service/handlers.py
```diff
@@ -53,10 +53,15 @@ def latest_committed_block(self) -> BlockNumber:
...
+        """ The latest block. """
```
karlb: The comment is the same as the function name and thus not helpful.

PR #727, opened 2020-02-17
change event name to processed due to PR #5734 in raiden

fredo:
With PR [#5734](https://github.com/raiden-network/raiden/pull/5734) the synced event how it is used here is renamed to `processed`. Thats what is desired here in the PFS which is the current sync to be processed.


karlb:
Looks like you have run black without `--line-length 99`. You can also use `make black` to run it with these settings.

You will also need to update the raiden dependency after merging https://github.com/raiden-network/raiden/pull/5734 to get it to run, I assume.

PR #729, opened 2020-02-25
Update to web3 5.5.1

palango:
Update web3 and fix problems that showed up. The biggest one is the
change of the eth-tester chain id from 1 to 61.


palango:
@karlb I had to remove xdist for now, as it crashes one node.

PR #737, opened 2020-03-06
Fix reward claim

### LINKED ISSUE #734: In MS Scenarios - No RewardClaimed event found for this channel
## Problem Definition 

In [MS1](https://github.com/raiden-network/raiden/blob/develop/raiden/tests/scenarios/ms1_simple_monitoring.yaml), [MS2](https://github.com/raiden-network/raiden/blob/develop/raiden/tests/scenarios/ms2_simple_monitoring.yaml) and [BF1](https://github.com/raiden-network/raiden/blob/develop/raiden/tests/scenarios/bf1_basic_functionality.yaml) we want to see the Raiden monitoring service to react properly. Monitoring works and the Monitoring Service reacts as expected. 

However, those scenarios fail constantly when the Monitoring Service tries to claim its reward with 

```
ScenarioAssertionError(\"No RewardClaimed event found for this channel.\") scenario_player.exceptions.legacy.ScenarioAssertionError: No RewardClaimed event found for this channel.",
```

The calculation of the point in time at which this reward is to be claimed seems wrong.

palango:
Fixes #734 

This fixes the claim of the MS reward by pushing the transaction back one block. This is necessary because of the same parity problem we already discovered in https://github.com/raiden-network/raiden-services/pull/728

As we have no time pressure to claim early, adding one block doesn't hurt.

I locally tested this with MS1 and it fixed the run.

### DIFF
--- a/src/monitoring_service/handlers.py
+++ b/src/monitoring_service/handlers.py
@@ -329,11 +329,16 @@ def monitor_new_balance_proof_event_handler(event: Event, context: Context) -> N
 
         context.database.upsert_channel(channel)
 
-    # check if this was our update, if so schedule the call
-    # of `claimReward`
-    # it will be checked there that our update was the latest one
+    # check if this was our update, if so schedule the call of
+    # `claimReward` it will be checked there that our update was the latest one
     if event.ms_address == context.ms_state.address:
         assert channel.closing_block is not None, "closing_block not set"
+        # Transactions go into the next mined block, so we could trigger one block
+        # before the `claim` call is allowed to succeed to include it in the
+        # first possible block.
+        # Unfortunately, parity does the gas estimation on the current block
+        # instead of the next one, so we have to wait for the first allowed
+        # block to be finished to send the transaction successfully on parity.
         trigger_block = BlockNumber(channel.closing_block + channel.settle_timeout + 1)
 
         # trigger the claim reward action by an event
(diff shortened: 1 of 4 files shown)


src/monitoring_service/handlers.py
```diff
@@ -329,25 +329,40 @@ def monitor_new_balance_proof_event_handler(event: Event, context: Context) -> N
...
     if event.ms_address == context.ms_state.address:
         assert channel.closing_block is not None, "closing_block not set"
-        trigger_block = BlockNumber(channel.closing_block + channel.settle_timeout)
+        # Transactions go into the next mined block, so we could trigger one block
+        # before the `claim` call is allowed to succeed to include it in the
+        # first possible block.
+        # Unfortunately, parity does the gas estimation on the current block
+        # instead of the next one, so we have to wait for the first allowed
+        # block to be finished to send the transaction successfully on parity.
+        trigger_block = BlockNumber(channel.closing_block + channel.settle_timeout + 1)
```
palango: This is the real fix.

tests/monitoring/monitoring_service/test_end_to_end.py
```diff
@@ -242,36 +242,25 @@ def test_e2e(  # pylint: disable=too-many-arguments,too-many-locals
...
-
     assert [e.event for e in query()] == [MonitoringServiceEvent.NEW_BALANCE_PROOF_RECEIVED]
 
     # wait for settle timeout
-    wait_for_blocks(2)  # timeout is 5, but we've already waited 3 blocks before
+    # timeout is 5, but we've already waited 3 blocks before. Additionally one block is
+    # added to handle parity running gas estimation on current instead of next.
+    wait_for_blocks(3)
 
-    token_network.functions.settleChannel(
```
palango: This was not necessary for the test

PR #739, opened 2020-03-09
Add contract infos to registration script

palango:
Logs some more info during registration

### DIFF
--- a/src/raiden_libs/register_service.py
+++ b/src/raiden_libs/register_service.py
@@ -62,7 +62,7 @@ def main(
     The address that is registered is derived from the supplied private key.
     It also sets or updates the URL of the services deployment.
     """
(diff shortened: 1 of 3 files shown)


src/raiden_libs/register_service.py
```diff
@@ -62,6 +62,12 @@ def main(
     The address that is registered is derived from the supplied private key.
     It also sets or updates the URL of the services deployment.
     """
+    log.info("Web3 client", node_address=web3.provider.endpoint_uri)
```
karlb: We usually use `node_address` for raiden nodes, not ethereum nodes.
```suggestion
    log.info("Using RPC endpoint", url=web3.provider.endpoint_uri)
```

PR #742, opened 2020-03-24
Adds disclaimer at startup

### LINKED ISSUE #741: Add Disclaimers when starting up the PFS and the MS
## Description

state that I know the risks of running a pfs or a ms 

**So that I can**
not sue bble or any other party when things end up unexpectedly using the software for the alpha release.

## Acceptance Criteria

- [ ] Get disclaimer from legal department
- [x] Implement disclaimer at startup for MS, see [Raiden implementation](https://github.com/raiden-network/raiden/blame/develop/raiden/ui/cli.py#L563), together with a `--accept-disclaimer` cli option
- [x] Implement disclaimer at startup for PFS, see [Raiden implementation](https://github.com/raiden-network/raiden/blame/develop/raiden/ui/cli.py#L563), together with a `--accept-disclaimer` cli option
- [ ] Adjust Raiden Service Bundle to start with `--accept-disclaimer` or show disclaimer at start-up

Dominik1999:
Fixes #741 

We need a disclaimer at startup like in the Raiden Client for legal reasons. The disclaimer is a placeholder and subject to change. 

Do not merge. Things will break, since PFS and MS will restart and cannot automatically accept the disclaimer. We need to adjust the script.


src/monitoring_service/cli.py
```diff
@@ -57,9 +65,52 @@ def main(  # pylint: disable=too-many-arguments
     confirmations: BlockNumber,
     min_reward: int,
     debug_shell: bool,
+    accept_disclaimer: bool,
 ) -> int:
     """ The Monitoring service for the Raiden Network. """
     log.info("Starting Raiden Monitoring Service")
+    click.secho(
+        textwrap.dedent(
```
karlb: Could you put the disclaimer into a variable outside of this function? The code is a bit unpleasant to read with this big text inside it.

Dominik1999: yes, will do

src/pathfinding_service/constants.py
```diff
@@ -26,3 +27,37 @@
 
 MAX_AGE_OF_IOU_REQUESTS: timedelta = timedelta(hours=1)
 MAX_AGE_OF_FEEDBACK_REQUESTS: timedelta = timedelta(minutes=10)
+
+DISCLAIMER: str = textwrap.dedent(
```
palango: This still exists twice

Dominik1999: yes, I am unsure if we get 2 or just 1 disclaimer from Jim. I will adjust accordingly

Dominik1999:
> Is it sufficient to have it for MS and PFS, or do we need it for all executables?

We would need that for all executables to be used by third parties

karlb:
> > Is it sufficient to have it for MS and PFS, or do we need it for all executables?
> 
> We would need that for all executables to be used by third parties

Then we might want to add that to `common_options`, so that the RC and the fee claiming script get the disclaimer, too. At least if we settle on a single disclaimer.

Dominik1999:
I already added that to the RC by hand. But your approach seems even better.

deployment/.env.template
```diff
@@ -15,17 +15,21 @@ PFS_HOST=0.0.0.0
 PFS_KEYSTORE_FILE=/keystore/<FILENAME>
 PFS_PASSWORD=<PASSWORD>
 PFS_LOG_LEVEL=DEBUG
+# PFS_ACCEPT_DISCLAIMER=TRUE
```
palango: Should we have the disclaimer somewhere so that we add a link here?

src/monitoring_service/cli.py
```diff
@@ -57,9 +65,19 @@ def main(  # pylint: disable=too-many-arguments
     confirmations: BlockNumber,
     min_reward: int,
     debug_shell: bool,
+    accept_disclaimer: bool,
 ) -> int:
     """ The Monitoring service for the Raiden Network. """
     log.info("Starting Raiden Monitoring Service")
+    click.secho(
```
palango: Why doesn't this get put on one line?

src/pathfinding_service/cli.py
```diff
@@ -1,10 +1,4 @@
 """Console script for pathfinding_service."""
-from gevent import monkey, config  # isort:skip # noqa
```
palango: This should stay at the beginning of the file

src/request_collector/cli.py
```diff
@@ -1,7 +1,4 @@
 from gevent import monkey  # isort:skip # noqa
-
-monkey.patch_all(subprocess=False, thread=False)  # isort:skip # noqa
```
palango: ditto

src/request_collector/cli.py
```diff
@@ -40,11 +40,28 @@
     multiple=True,
     help="Use this matrix server instead of the default ones. Include protocol in argument.",
 )
+@click.option(
```
palango: Can you move this into `common_options`?

Dominik1999: It looks like we are going to use 2 different disclaimers. Then this is not needed, right?

src/monitoring_service/cli.py
Dominik1999: I do not understand your comment, can we have a short call about that?

karlb: I think he just talks about the formatting. If you run black on that file, this line and the following ones should be merged into a single line.

Dominik1999: ah I see, but my black tells me otherwise :)
at least I did run black several times on that file and it did not change. I can change by hand though.

src/pathfinding_service/cli.py
```diff
@@ -31,6 +27,12 @@
 from raiden_libs.blockchain import get_web3_provider_info
 from raiden_libs.cli import blockchain_options, common_options, setup_sentry
 
+# pylint: disable=too-many-locals
+# there were some issues with the 'thread' resolver, remove it from the options
+config.resolver = ["dnspython", "ares", "block"]  # noqa
```
palango: This is still wrong

src/pathfinding_service/cli.py
```diff
@@ -81,9 +90,19 @@ def main(  # pylint: disable-msg=too-many-arguments
     info_message: str,
     enable_debug: bool,
     matrix_server: List[str],
+    accept_disclaimer: bool,
 ) -> int:
     """ The Pathfinding service for the Raiden Network. """
     log.info("Starting Raiden Pathfinding Service")
+    click.secho(
```
palango: Please reformat

src/pathfinding_service/cli.py
Dominik1999: in line 30?

src/pathfinding_service/cli.py
```diff
@@ -81,9 +90,19 @@ def main(  # pylint: disable-msg=too-many-arguments
...
+        DISCLAIMER, fg="yellow",
+    )
+    if not accept_disclaimer:
+        click.confirm(
+            "\nHave you read, understood and hereby accept the above "
```
palango: You could also make the message a constant and use it everywhere

src/pathfinding_service/cli.py
palango: Ah no, this is about the gevent monkey patching. This needs to happen first

src/pathfinding_service/cli.py
Dominik1999: indeed, done

src/pathfinding_service/cli.py
Dominik1999: hmm it works if I put it there. Any idea where I should put it instead?

deployment/.env.template
palango: @Dominik1999

deployment/.env.template
```diff
@@ -15,17 +15,21 @@ PFS_HOST=0.0.0.0
...
+# MS_ACCEPT_DISCLAIMER=TRUE
 
 MSRC_KEYSTORE_FILE=/keystore/<FILENAME>
 MSRC_PASSWORD=<PASSWORD>
 MSRC_LOG_LEVEL=DEBUG
+# MSRC_ACCEPT_DISCLAIMER=TRUE
 
 MS_BACKUP_KEYSTORE_FILE=/keystore/<FILENAME_BACKUP>
 MS_BACKUP_PASSWORD=<PASSWORD_BACKUP>
+# MS_ACCEPT_DISCLAIMER=TRUE
```
palango: I don't think we need that one.

src/monitoring_service/cli.py
```diff
@@ -58,9 +67,15 @@ def main(  # pylint: disable=too-many-arguments
     confirmations: BlockNumber,
     min_reward: int,
     debug_shell: bool,
+    accept_disclaimer: bool,
 ) -> int:
     """ The Monitoring service for the Raiden Network. """
     log.info("Starting Raiden Monitoring Service")
+    click.secho(MS_DISCLAIMER, fg="yellow")
+    if not accept_disclaimer:
+        click.confirm(
```
palango: This might fit on one line as well

src/pathfinding_service/cli.py
palango: It should go where it was before, right at the top of the file

src/pathfinding_service/cli.py
```diff
@@ -81,9 +91,15 @@ def main(  # pylint: disable-msg=too-many-arguments
     info_message: str,
     enable_debug: bool,
     matrix_server: List[str],
+    accept_disclaimer: bool,
 ) -> int:
     """ The Pathfinding service for the Raiden Network. """
     log.info("Starting Raiden Pathfinding Service")
+    click.secho(PFS_DISCLAIMER, fg="yellow")
+    if not accept_disclaimer:
+        click.confirm(
```
palango: see above

src/raiden_libs/constants.py
```diff
@@ -4,3 +4,7 @@
 UDC_SECURITY_MARGIN_FACTOR_PFS: float = 2
 
 MATRIX_START_TIMEOUT = 30  # in seconds
+
+CONFIRMATION_OF_UNDERSTANDING = (
+    "\nHave you read, understood and hereby accept the above disclaimer and privacy warning?"
```
palango: Do we need the newline in front here?

src/request_collector/cli.py
```diff
@@ -10,9 +7,14 @@
 import structlog
 from request_collector.server import RequestCollector
 
+from monitoring_service.constants import MS_DISCLAIMER
 from monitoring_service.database import SharedDatabase
 from raiden.utils.cli import NetworkChoiceType
 from raiden_libs.cli import common_options, setup_sentry
+from raiden_libs.constants import CONFIRMATION_OF_UNDERSTANDING
+
+monkey.patch_all(subprocess=False, thread=False)  # isort:skip # noqa
```
palango: Same as above

src/request_collector/cli.py
```diff
@@ -40,11 +42,24 @@
...
+def main(
+    private_key: str, state_db: str, matrix_server: List[str], accept_disclaimer: bool
+) -> int:
     """ The request collector for the monitoring service. """
     log.info("Starting Raiden Monitoring Request Collector")
-
+    click.secho(MS_DISCLAIMER, fg="yellow")
+    if not accept_disclaimer:
+        click.confirm(
+            CONFIRMATION_OF_UNDERSTANDING, abort=True,
```
palango: see above

src/monitoring_service/cli.py

src/request_collector/cli.py
Dominik1999: ok done

deployment/.env.template
Dominik1999: Yes, can we simply link to the constants.py on gitHub?

deployment/.env.template
Dominik1999: ok removed. How does the backup MS start then?

deployment/.env.template
Dominik1999: atm we do not need to link that. We need to come up with a holistic view on which disclaimers to put where

palango:
This needs updates to the deployments after this is merged.

PR #745, opened 2020-03-24
Update raiden, fix typing problems

### LINKED ISSUE #744: Update services to latest raiden
Besides the usual updates this will require some typing changes, as web3 now exports its type definitions.

palango:
Fixes #744

### DIFF
--- a/src/raiden_libs/blockchain.py
+++ b/src/raiden_libs/blockchain.py
@@ -48,7 +48,7 @@ def get_web3_provider_info(web3: Web3) -> str:
         if endpoint is not None:
             return str(endpoint)
 
-    return ""
+    raise RuntimeError(f"Unsupported web3 provider {provider!r}")
 
 
 def create_registry_event_topics(contract_manager: ContractManager) -> List:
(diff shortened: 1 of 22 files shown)


src/raiden_libs/blockchain.py
```diff
@@ -35,15 +36,31 @@
...
+    Currently works only with `HTTPProvider`. Needs to be adapted when new procviders
+    are added.
+    """
+    provider = web3.provider
+    if isinstance(provider, HTTPProvider):
+        endpoint = provider.endpoint_uri
+        if endpoint is not None:
+            return str(endpoint)
+
+    return ""
```
karlb: Should we raise an error here, instead?

tests/monitoring/monitoring_service/test_handlers.py
```diff
@@ -169,7 +169,7 @@ def test_channel_opened_event_handler_adds_channel(context: Context):
 
 def test_channel_closed_event_handler_closes_existing_channel(context: Context):
     context = setup_state_with_open_channel(context)
-    context.web3.eth.blockNumber = BlockNumber(60)
+    context.web3.eth.blockNumber = BlockNumber(60)  # type: ignore
```
karlb: What is the typing problem, here?

palango: error: Property "blockNumber" defined in "Eth" is read-only

src/raiden_libs/blockchain.py
palango: Good idea

PR #748, opened 2020-03-24
Add `withdraw` command and rename `register_service `

### LINKED ISSUE #743: Add withdraw command to register_service.py
We need a way to withdraw tokens that have been deposited by the service registry. This can't be easily done via Etherscan, since each deposit gets its own instance of the `Deposit` contract, so that those are not verified.

karlb:
Closes #743.

### DIFF
--- a/deployment/docker-compose.yml
+++ b/deployment/docker-compose.yml
@@ -213,7 +213,7 @@ services:
   registration-goerli:
     <<: *defaults
     restart: "no"
-    command: ["python3", "-m", "raiden_libs.register_service"]
+    command: ["python3", "-m", "raiden_libs.service_registry"]
     environment:
       - RDN_REGISTRY_LOG_LEVEL=DEBUG
       - RDN_REGISTRY_KEYSTORE_FILE=${MS_KEYSTORE_FILE}
@@ -227,7 +227,7 @@ services:
   registration-goerli-backup:
     <<: *defaults
     restart: "no"
-    command: ["python3", "-m", "raiden_libs.register_service"]
+    command: ["python3", "-m", "raiden_libs.service_registry"]
     environment:
       - RDN_REGISTRY_LOG_LEVEL=DEBUG
       - RDN_REGISTRY_KEYSTORE_FILE=${MS_BACKUP_KEYSTORE_FILE}
--- a/deployment/docker-compose.yml
+++ b/deployment/docker-compose.yml
@@ -213,7 +213,7 @@ services:
   registration-goerli:
     <<: *defaults
     restart: "no"
-    command: ["python3", "-m", "raiden_libs.service_registry"]
+    command: ["python3", "-m", "raiden_libs.service_registry", "register"]
     environment:
       - RDN_REGISTRY_LOG_LEVEL=DEBUG
       - RDN_REGISTRY_KEYSTORE_FILE=${MS_KEYSTORE_FILE}
@@ -227,7 +227,7 @@ services:
   registration-goerli-backup:
     <<: *defaults
     restart: "no"
-    command: ["python3", "-m", "raiden_libs.service_registry"]
+    command: ["python3", "-m", "raiden_libs.service_registry", "register"]
     environment:
       - RDN_REGISTRY_LOG_LEVEL=DEBUG
       - RDN_REGISTRY_KEYSTORE_FILE=${MS_BACKUP_KEYSTORE_FILE}
--- a/src/raiden_libs/cli.py
+++ b/src/raiden_libs/cli.py
@@ -24,6 +24,7 @@ from web3.middleware import geth_poa_middleware
 from pathfinding_service.middleware import http_retry_with_backoff_middleware
 from raiden.utils.typing import Address, BlockNumber, ChainID
 from raiden_contracts.constants import (
+    CONTRACT_DEPOSIT,
     CONTRACT_MONITORING_SERVICE,
     CONTRACT_ONE_TO_N,
     CONTRACT_SERVICE_REGISTRY,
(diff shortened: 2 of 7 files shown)


karlb:
Do we have any documentation on that command that needs updating? We don't have any in this repo.

palango:
> Do we have any documentation on that command that needs updating? We don't have any in this repo.

Doesn't seem like it, so far it was for internal use only. Wonder if we should add it here or to the RSB...

src/raiden_libs/cli.py
```diff
@@ -187,7 +188,9 @@ def blockchain_options(contracts: List[str]) -> Callable:
         CONTRACT_MONITORING_SERVICE: "monitor-contract",
         CONTRACT_ONE_TO_N: "one-to-n-contract",
         CONTRACT_SERVICE_REGISTRY: "service-registry-contract",
+        CONTRACT_DEPOSIT: "service-deposit-contract",
```
palango: what about `service-registry-deposit-contract`?

src/raiden_libs/contract_info.py
```diff
@@ -44,8 +44,9 @@ def get_contract_addresses_and_start_block(
 
     # Get deployed addresses for those contracts which have no overwrites
     addresses = {
-        c: address_overwrites.get(
-            c, to_canonical_address(contract_data["contracts"][c]["address"])
+        c: (
+            address_overwrites.get(c)
+            or to_canonical_address(contract_data["contracts"][c]["address"])
```
palango: Is there a difference here?

src/raiden_libs/service_registry.py
```diff
@@ -180,5 +187,55 @@ def update_service_url(
         )
 
 
+@blockchain_options(contracts=[CONTRACT_DEPOSIT])
+@cli.command("withdraw")
+@click.option(
+    "--to", type=str, callback=validate_address, help="Target address for withdrawn tokens"
```
palango: Is this automatically required?

src/raiden_libs/service_registry.py
```diff
@@ -180,5 +187,55 @@ def update_service_url(
...
+    deposit_contract = contracts[CONTRACT_DEPOSIT]
+
+    # Check usage of correct key
+    withdrawer = deposit_contract.functions.withdrawer().call()
+    caller_address = private_key_to_address(private_key)
+    if to_canonical_address(withdrawer) != caller_address:
+        log.error(
+            "You must used the key used to deposit when withdrawing",
+            expected=withdrawer,
+            actual=caller_address,
```
palango: This should be checksum encoded

src/raiden_libs/service_registry.py
```diff
@@ -180,5 +187,55 @@ def update_service_url(
...
+        )
+        sys.exit(1)
+
+    # Can we withdraw already?
+    release_at = deposit_contract.functions.release_at().call()
+    if time.time() < release_at:
+        log.error("Too early to withdraw")
```
palango: Would be nice to log how much time is missing...

src/raiden_libs/service_registry.py
```diff
@@ -180,5 +187,55 @@ def update_service_url(
...
+        sys.exit(1)
+
+    # Withdraw now!
+    log.info("Withdrawing", to=to_checksum_address(to))
+    deposit_contract.functions.withdraw(to).transact({"from": caller_address})
```
palango: Can you use `checked_transact` here?

deployment/docker-compose.yml
```diff
@@ -213,7 +213,7 @@ services:
   registration-goerli:
     <<: *defaults
     restart: "no"
-    command: ["python3", "-m", "raiden_libs.register_service"]
+    command: ["python3", "-m", "raiden_libs.service_registry"]
```
palango: Don't we need to add `register` here?

src/raiden_libs/cli.py
karlb: I thought "user-deposit" and "service-deposit" were a nice pair, and it's a bit shorter. But I don't really mind either way.

src/raiden_libs/contract_info.py
karlb: Yes, a very subtle one. In the old version the fallback was evaluated in every case, causing a key error for the deposit contract.

src/raiden_libs/cli.py
palango: Fine for me as well. BTW, how do people get the address? Do they have to filtere the events themselves?

Can we log it during registration at least?

src/raiden_libs/service_registry.py
karlb: No, it isn't. I'll change it. I first had it as an `argument`, which is always required. But then I noticed we don't use arguments anywhere and made it an option, again.

src/raiden_libs/service_registry.py
karlb: Right, I lost that when introducing the `caller_address` variable.

deployment/docker-compose.yml
karlb: Yes, well spotted!

src/raiden_libs/service_registry.py
karlb: Done.

src/raiden_libs/service_registry.py
karlb: Done.

karlb:
Great feedback! I've updated the code and also included one more cleanup (see last commit).

src/raiden_libs/cli.py
karlb: I used Etherscan when testing. I wanted to get this quickly done, so I skipped event filtering and automated tests. Both make sense in later PRs.

PR #756, opened 2020-03-26
Add contract suffix to token netwrok registry CLI option

palango:
Mostly for consistency (and my OCD 🙈 )


palango:
> Do we need use that in any of the other repos right now (SP, service bundle) where it needs to be updated?

SP doesn't use the services directly. I don#t think the RSB either. But we need to check there anyways after Dominiks changes are merged.

Issue #782, opened 2020-05-19
Adaptive block chain filtering


palango:
Filtering events on mainnet is quite a lot slower than on testnets.

This leads to problems as fixed in https://github.com/raiden-network/raiden/pull/6092

We need to do something similar for the services as well.

PR #783, opened 2020-05-19
Adaptive blockchain syncing

palango:
This PR adds adaptive blockchain syncing to the services similar to what https://github.com/raiden-network/raiden/pull/6092 did in Raiden.

Mid-term we should try to merge the two systems as there is a lot of duplicated effort.

This is the minimal impact PR I could build, I have some refactorings at hand but are hesitant to merge them so close to release.

This measures the time the `eth_getLogs` RPC call takes and adjusts the query range accordingly. In comparison to https://github.com/raiden-network/raiden/pull/6092 I chose to reduce the range more aggressively when a timeout is hit.

Resolves https://github.com/raiden-network/raiden-services/issues/782

### DIFF
--- a/src/monitoring_service/constants.py
+++ b/src/monitoring_service/constants.py
@@ -1,8 +1,11 @@
 import textwrap
 from datetime import timedelta
 
-DEFAULT_REQUIRED_CONFIRMATIONS: int = 10
-MAX_FILTER_INTERVAL: int = 100_000
+from raiden.utils.typing import BlockTimeout
+
+DEFAULT_FILTER_INTERVAL: BlockTimeout = BlockTimeout(1_000)
+MAX_FILTER_INTERVAL: BlockTimeout = BlockTimeout(100_000)
+MIN_FILTER_INTERVAL: BlockTimeout = BlockTimeout(2)
 DEFAULT_GAS_BUFFER_FACTOR: int = 10
 DEFAULT_GAS_CHECK_BLOCKS: int = 100
 KEEP_MRS_WITHOUT_CHANNEL: timedelta = timedelta(minutes=15)
--- a/src/monitoring_service/service.py
+++ b/src/monitoring_service/service.py
@@ -7,6 +7,7 @@ import sentry_sdk
 import structlog
 from eth_typing import Hash32
 from eth_utils import to_canonical_address
+from requests.exceptions import ReadTimeout
 from web3 import Web3
 from web3.contract import Contract
 from web3.exceptions import TransactionNotFound
@@ -17,9 +18,11 @@ from monitoring_service.constants import (
     DEFAULT_GAS_CHECK_BLOCKS,
     KEEP_MRS_WITHOUT_CHANNEL,
     MAX_FILTER_INTERVAL,
+    MIN_FILTER_INTERVAL,
 )
 from monitoring_service.database import Database
 from monitoring_service.handlers import HANDLERS, Context
+from raiden.constants import ETH_GET_LOGS_THRESHOLD_FAST, ETH_GET_LOGS_THRESHOLD_SLOW
 from raiden.utils.typing import BlockNumber, BlockTimeout, ChainID, MonitoringServiceAddress
 from raiden_contracts.constants import (
     CONTRACT_MONITORING_SERVICE,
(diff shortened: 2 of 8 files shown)


src/monitoring_service/constants.py
```diff
@@ -1,8 +1,11 @@
 import textwrap
 from datetime import timedelta
 
-DEFAULT_REQUIRED_CONFIRMATIONS: int = 10
-MAX_FILTER_INTERVAL: int = 100_000
+from raiden.utils.typing import BlockTimeout
+
+DEFAULT_FILTER_INTERVAL: BlockTimeout = BlockTimeout(1_000)
+MAX_FILTER_INTERVAL: BlockTimeout = BlockTimeout(100_000)
+MIN_FILTER_INTERVAL: BlockTimeout = BlockTimeout(2)
```
ulope, line 8: Nothing to do with your PR but that type alias name is pretty confusing...

src/monitoring_service/service.py
```diff
@@ -147,40 +150,65 @@ def start(
...
+        from_block = BlockNumber(blockchain_state.latest_committed_block + 1)
+        to_block = min(
+            latest_confirmed_block,
+            # decrement by one, as `from_block` is inclusive
+            BlockNumber(from_block + blockchain_state.current_event_filter_interval - 1),
         )
 
+        try:
+            before_query = time.monotonic()
+            new_chain_state, events = get_blockchain_events(
```
hackaugusto, line 173: This is kinda problematic:

- the code is using the standard library deepcopy, which we know it is slow
- the code is doing a single request per address, this is not bad for the MS but to any other node that is using the same Ethereum node.
- the time check here is for the above plus the decoding of events, which I would guess is also slow

so the measurement here is not the threadshold of a single query, and it has lots of moving pieces in the middle, which makes it hard to reason about.

can you please fix these problems?

src/monitoring_service/constants.py
palango: Definitively, I forget the exact name every time.

src/pathfinding_service/service.py
```diff
@@ -168,26 +169,66 @@ def _process_new_blocks(self, to_block: BlockNumber) -> None:
...
-            web3=self.web3,
-            contract_manager=CONTRACT_MANAGER,
-            chain_state=self.blockchain_state,
-            to_block=to_block,
+        # increment by one, as `latest_committed_block` has been queried last time already
+        from_block = BlockNumber(self.blockchain_state.latest_committed_block + 1)
+        to_block = min(
+            latest_confirmed_block,
+            # decrement by one, as `from_block` is inclusive
+            BlockNumber(from_block + self.blockchain_state.current_event_filter_interval - 1),
```
hackaugusto, line 177: the `-1` is not needed here AFAIU

src/pathfinding_service/service.py
```diff
@@ -168,26 +169,66 @@ def _process_new_blocks(self, to_block: BlockNumber) -> None:
...
         if events:
             log.info(
                 "Processed blocks",
-                from_block=self.blockchain_state.latest_committed_block + 1,
+                from_block=from_block,
                 to_block=to_block,
-                getting=round(before_process - before_get, 2),
-                processing=round(time.time() - before_process, 2),
-                total_duration=round(time.time() - start, 2),
+                interval=to_block - from_block,
```
hackaugusto, line 228: you probably also want to do the `+1` here? https://github.com/raiden-network/raiden-services/pull/783/files#diff-15eaee2c72a6c3d020822b921b44d79aR167

palango: Ah, this is just a leftover. I'll remove it

src/monitoring_service/service.py
palango: 1. The `BlockchainState` object has only tow layers, and it's only copied once for each new block. So I don't think it's urgent. It's already fixed in https://github.com/raiden-network/raiden-services/pull/785
2. This is true, but I would prefer to not touch that now but improve it later
3. It measures the whole process, doing the queries and decoding the events. I don't see why this is worse than just adapting to the actual RPC call, it's just a different piece of work we expect to be finished in the given amount of time.

Issue #672, opened 2019-12-02
Improve PFS Performance - Optimize blockchain polling


Dominik1999:
Optimize `event_abi_to_log_topic` because 5.77% of the time is spent here.
  - Moved to https://github.com/raiden-network/raiden-services/issues/794
We might save up to ~40% of the time by batching of event polling.

First:

- [x] finalize https://github.com/raiden-network/raiden/pull/5003 in Raiden
- [ ] batch event polling like in Raiden here https://github.com/raiden-network/raiden/pull/5003
- [x] combine multiple `get_logs` calls to one

Related to #651

palango:
Further improvements tracked in #793

PR #785, opened 2020-05-20
WIP: Further refactor

palango:
Follow-up to #783 

During the work on adaptive syncing I saw some opportunities to refactor stuff to reduce duplicated code, but didn't want to merge it back then.

This is mostly about removing the token network addresses from the blockchain state class. This make the code paths of MS and PFS similar and will allow to factor out the code there.

We now need to query the token network from the DB every time a new block is queried, but this only happens once per new block, so that shouldn't hurt.

## TODO
- [x] Factor out common blockchain filter code
- [x] Fix/adapt tests

### DIFF
--- a/src/monitoring_service/service.py
+++ b/src/monitoring_service/service.py
@@ -7,7 +7,6 @@ import sentry_sdk
 import structlog
 from eth_typing import Hash32
 from eth_utils import to_canonical_address
-from requests.exceptions import ReadTimeout
 from web3 import Web3
 from web3.contract import Contract
 from web3.exceptions import TransactionNotFound
@@ -17,12 +16,9 @@ from monitoring_service.constants import (
     DEFAULT_GAS_BUFFER_FACTOR,
     DEFAULT_GAS_CHECK_BLOCKS,
     KEEP_MRS_WITHOUT_CHANNEL,
-    MAX_FILTER_INTERVAL,
-    MIN_FILTER_INTERVAL,
 )
 from monitoring_service.database import Database
 from monitoring_service.handlers import HANDLERS, Context
-from raiden.constants import ETH_GET_LOGS_THRESHOLD_FAST, ETH_GET_LOGS_THRESHOLD_SLOW
 from raiden.utils.typing import BlockNumber, BlockTimeout, ChainID, MonitoringServiceAddress
 from raiden_contracts.constants import (
     CONTRACT_MONITORING_SERVICE,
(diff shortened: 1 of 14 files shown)


src/monitoring_service/service.py
```diff
@@ -199,18 +202,6 @@ def _process_new_blocks(self, latest_confirmed_block: BlockNumber) -> None:
...
             return
 
-        # If a new token network was found we need to write it to the DB, otherwise
-        # the constraints for new channels will not be constrained. But only update
-        # the network addresses here, all else is done later.
-        token_networks_changed = (
-            blockchain_state.token_network_addresses != new_chain_state.token_network_addresses
-        )
-        if token_networks_changed:
-            blockchain_state.token_network_addresses = new_chain_state.token_network_addresses
```
palango, line 209: I think this stuff is way nicer to do via an event handler

src/raiden_libs/blockchain.py
```diff
@@ -157,15 +156,15 @@ def parse_token_network_event(event: dict) -> Optional[Event]:
 def get_blockchain_events(
     web3: Web3,
     contract_manager: ContractManager,
+    token_network_addresses: List[TokenNetworkAddress],
     chain_state: BlockchainState,
     from_block: BlockNumber,
     to_block: BlockNumber,
-) -> Tuple[BlockchainState, List[Event]]:
+) -> List[Event]:
```
karlb, line 174: Really nice to get rid of the BlockchainState here!

src/raiden_libs/blockchain.py
```diff
@@ -157,15 +156,15 @@ def parse_token_network_event(event: dict) -> Optional[Event]:
 def get_blockchain_events(
     web3: Web3,
     contract_manager: ContractManager,
+    token_network_addresses: List[TokenNetworkAddress],
```
karlb, line 170: The list gets mutated in the function. Should we make that more obvious? Maybe a note in the docstring.

karlb: Did you change anything about this?

karlb: I see you did. It's just not in this function but one function further outside.

src/raiden_libs/blockchain.py
```diff
@@ -296,3 +309,71 @@ def get_pessimistic_udc_balance(
         udc.functions.effectiveBalance(address).call(block_identifier=BlockNumber(block))
         for block in range(from_block, to_block + 1)
     )
+
+
+def get_blockchain_events_adaptive(
```
karlb, line 314: This somewhat duplicates what we have in Raiden, but I guess extracting just the reusable part is not trivial.

palango: This is tracked in https://github.com/raiden-network/raiden-services/issues/672

PR #793, opened 2020-06-02
Batch the RPC request for token network events

palango:
On deployments with multiple token networks this lowers the number of
RPC requests by a big margin.

As we are interested in all events we do not need the filter topics. With them gone we can query multiple contracts at the same time. For now I chose to only do this for the token networks, as it is the easiest to do.

Part of https://github.com/raiden-network/raiden-services/issues/672

### DIFF
--- a/src/raiden_libs/blockchain.py
+++ b/src/raiden_libs/blockchain.py
@@ -1,5 +1,5 @@
 import time
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Union
 
 import structlog
 from eth_abi.codec import ABICodec
@@ -84,9 +84,8 @@ def decode_event(
 def query_blockchain_events(
     web3: Web3,
     contract_manager: ContractManager,
-    contract_address: Address,
+    contract_addresses: Union[Address, List[Address]],
     contract_name: str,
-    topics: List,
     from_block: BlockNumber,
     to_block: BlockNumber,
 ) -> List[Dict]:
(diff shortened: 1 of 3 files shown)


src/raiden_libs/blockchain.py
```diff
@@ -259,9 +256,9 @@ def get_monitoring_blockchain_events(
     monitoring_service_events = query_blockchain_events(
         web3=web3,
         contract_manager=contract_manager,
-        contract_address=monitor_contract_address,
+        contract_addresses=monitor_contract_address,
         contract_name=CONTRACT_MONITORING_SERVICE,
-        topics=[None],
+        # topics=[None],
```
karlb, line 261: Do we want to keep this line?

src/raiden_libs/blockchain.py
```diff
@@ -95,9 +94,9 @@ def query_blockchain_events(
     Args:
         web3: A Web3 instance
         contract_manager: A contract manager
-        contract_address: The address of the contract to be filtered, can be `None`
+        contract_addresses: The address of the contract to be filtered, can be a single address
```
karlb, line 97: This single-address-or-list is a bit strange. We could always take a list and call it with single element lists where necessary. If we keep both options, I would probably prefer the singular for the parameter name. No need to change this if you prefer it this way, though.

src/raiden_libs/blockchain.py
palango: Nice catch, this shouldn't be there.

src/raiden_libs/blockchain.py
palango: Yeah, I was unhappy with that as well, I chose the union because the single address is still common, but on the second thought it seems like the better idea to only have a list as the parameter.

Issue #763, opened 2020-03-30
Provide healthcheck in service repo


karlb:
There are basic healthchecks in the raiden service bundle repo. Having these checks in this repo would be preferable, since that would allow the service to test and update the checks as needed.

palango:
Related to https://github.com/raiden-network/raiden-services/issues/650

Issue #650, opened 2019-11-26
Add APIs for healtchchecking the services


hackaugusto:
## Problem Definition

Both the MS and the PFS must have healtcheck APIs for proper provisioning checks.

karlb:
How should such a thing look for the MS, which does not really have an API? Updating a file with a timestamp to show that it's still alive?

hackaugusto:
> How should such a thing look for the MS, which does not really have an API? Updating a file with a timestamp to show that it's still alive?

Just add an endpoint `/health` that return a valid response if the system is properly initialized. Having a simple HTTP interface makes it easier to integrate with existing DevOps tools

karlb:
https://github.com/raiden-network/raiden-services/issues/763 is more detailed.

Issue #501, opened 2019-08-06
MS price feedback


palango:
The MS' need a way to let the clients know their expected reward for monitoring.

There are some ideas on how to accomplish that:
- Open a broadcasting channel in which the MS post
- Attach the information as matrix metadata to the MS matrix identities

@ulope Can you provide some input here?

ulope:
Both seem reasonable but I think we should limit the use of broadcasting rooms.
The metadata approach seems simpler to implement as well.

How often can those fees change? I'm asking because if we go with the metadata approach we need some mechanism for the client to periodically refresh that information (I'm not sure if there is an event generated in the sync response when the displayname (which is what we'd be 'abusing' for this) changes).

However thinking about it now, wouldn't this also be a good candidate to use a REST API for? Similar to how the PFS does it?

Issue #7, opened 2018-03-29
Service monitoring/Dashboard


palango:
## Raiden Service Bundle operators want to see metrics of their services
The Raiden Services, Pathfinding and Monitoring, are being operated by external providers. Those providers want to see a dashboard showing relevant and interesting information. The dashboard can be in parts but should not be in general publicly accessible. 

## Get familiar with the Raiden Service Bundle
- Specification of the Monitoring Service https://raiden-network-specification.readthedocs.io/en/latest/monitoring_service.html
- Specification of the Pathfinding Service https://raiden-network-specification.readthedocs.io/en/latest/pathfinding_service.html
- Services Repo on GitHub https://github.com/raiden-network/raiden-services

## Plan in detail
Please plan this issue here in small steps such that implementation can be iterative. Please sketch a design of the dashboard and the necessary inputs first. 

Bonus: Our colleagues from eth.events, our community manager Boris and our friends from dappNode can give you feedback or ideas which information should be shown on the dashboard. @czepluch can provide contacts.

this was initially planned by the team https://github.com/raiden-network/raiden-services/issues/7#issuecomment-635385656

- [ ] Sketch a design of the dashboard - what data should we show there?
- [ ] From the data, deduct the necessary endpoints in the PFS and MS
- [ ] ...

++++++
Original Issue
The running PFS should provide some insights, e.g. requests per minute, network topology...

Users want to see how many requests were processed and how many IOUs collected. Maybe we can create an easy and secure endpoint for that?

palango:
It might make sense to talk to Nils about that and maybe implement a prototype while he is still in Berlin.

ulope:
Since we already use Prometheus in a couple of places it would make sense to continue that here. 

The Python client library is also pretty easy to use.

konradkonrad:
- specify what data to provide
- create a `_stats` or the like endpoint

ulope:
Depending if we want to go with prometheus or provide a generic api endpoint that can then reformatted the python client library could be helpful to get something working quickly: https://github.com/prometheus/client_python

fredo:
- [ ] Talk to the current providers what kind of metrics are desired
- [ ] investigate prometheus approach (maybe with the help of @ulope @nlsdfnbch )
- [ ] new docker container in docker-compose set up 
- [ ] accessible by new subdomain

nlsdfnbch:
[...]

PR #797, opened 2020-06-05
Add API to MS

palango:
This adds a REST API to the MS.

This is done in preparation for https://github.com/raiden-network/raiden-services/issues/7 for which we need a working API to export prometheus stats. However, it's also useful for https://github.com/raiden-network/raiden-services/issues/501, https://github.com/raiden-network/raiden-services/issues/650 and https://github.com/raiden-network/raiden-services/issues/763

The changes include:
- `mypy` fixes to changes caused by ne `__init__.py` file
- Make `MonitoringService` a `greenlet`
- Add `MsApi` class with currently only a `/info` endpoint similar to the PFS'
- Add an API test
- Move common test stuff to libs


src/pathfinding_service/cli.py
```diff
@@ -128,7 +130,7 @@ def main(  # pylint: disable=too-many-arguments,too-many-locals
             raise Exception("PFS did not start within time.")
 
         log.debug("Starting API")
-        api = ServiceApi(
+        api = PfsApi(
```
karlb, line 133: I would have expected `PFSApi`, but both are a bit awkward in their own way.

src/monitoring_service/api.py
```diff
@@ -0,0 +1,91 @@
...
+                "service_token_address": to_checksum_address(self.service_token_address),
+                "confirmed_block": {
+                    "number": self.monitoring_service.context.ms_state.blockchain_state.latest_committed_block  # noqa
+                },
+            },
+            "version": self.version,
+            "contracts_version": self.contracts_version,
+            "operator": self.api.operator,
+            "message": self.api.info_message,
+            "UTC": datetime.utcnow().isoformat(),
```
karlb, line 51: What are the security implications of providing this information? If it is public, attackers might watch for `latest_committed_block` to be outdated and take advantage of that situation.

src/monitoring_service/service.py
```diff
@@ -78,7 +78,8 @@ def handle_event(event: Event, context: Context) -> None:
                 sentry_sdk.capture_exception(ex)
 
 
-class MonitoringService:  # pylint: disable=too-few-public-methods,too-many-instance-attributes
+class MonitoringService(gevent.Greenlet):
```
karlb, line 81: Please don't inherit from `Greenlet`! It ruins mypy checks and couples the implementation to gevent.

We didn't settle on gevent guidelines yet, but I suggest something like
```python

ms_greenlet = spawn_named("MonitoringService", ms)
...
ms_greenlet.kill()
```

src/monitoring_service/api.py
palango: There might be, even though I'm not aware of any as the MS are not giving any guarantees right now. However, I'd keep the API non-public for now.

src/monitoring_service/cli.py
```diff
@@ -80,24 +113,40 @@ def main(  # pylint: disable=too-many-arguments
...
+
+        if debug_shell:
+            import IPython
+
+            IPython.embed()
+            return 0
 
-    if debug_shell:
-        import IPython
+        task = gevent.spawn(service.start)
```
karlb, line 136: Why not spawn_named?

src/monitoring_service/cli.py
```diff
@@ -80,24 +113,40 @@ def main(  # pylint: disable=too-many-arguments
...
 
-    ms.start()
+        task.get()
+    finally:
+        log.info("Stopping Monitoring Service...")
+        if api:
+            api.stop()
+        if task:
+            task.kill()
+            task.join()
```
karlb, line 149: Wouldn't `task.get()` be even better here? If it stops due to a GreenletExit  generated by `kill`, it should pass without an error, but raise errors that happened during cleanup.

src/monitoring_service/api.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def run(self, host: str, port: int) -> None:
+        self.rest_server = WSGIServer((host, port), self.flask_app)
+        self.rest_server.start()
+
+        log.info("Running endpoint", endpoint=f"http://{host}:{port}")
+
+    def stop(self) -> None:
+        if self.rest_server:
+            self.rest_server.stop()
```
karlb, line 91: If we want to go all in with the `spawn/kill` model, this should be
```python
    def run(self, host: str, port: int) -> None:
        self.rest_server = WSGIServer((host, port), self.flask_app)
        log.info("Running endpoint", endpoint=f"http://{host}:{port}")
        self.rest_server.server_forever()
```
But we can start doing that when we have some gevent guidelines.

palango: Let's do that later...

PR #810, opened 2020-06-23
Use GitHub Actions

palango:
Use GitHub actions for CI

The config is more straight forward, especially with regard to matrix (as in matrix of different parameters) testing.


palango:
@karlb According to https://github.community/t/actions-usage-for-a-public-repo-on-a-large-org/16880 public repos have no limited minutes per month.

PR #817, opened 2020-07-09
Read and process claims in PFS

palango:
First step for https://github.com/raiden-network/raiddit/issues/33


src/pathfinding_service/service.py
```diff
@@ -152,6 +158,50 @@ def _run(self) -> None:  # pylint: disable=method-hidden
...
 
+    def _process_claims_file(self) -> None:
+        if self.claims_file is None:
+            return
+
+        log.info("Reading claims from file", claims_file=self.claims_file)
+        claims_data = [json.loads(line) for line in self.claims_file.readlines()]
+
+        operator = to_canonical_address(claims_data.pop(0)["operator"])
+        claims = [
```
palango, line 169: This and the line above will keep all claims in memory. Ulo had some problems with that with a lot of claims. Should we make this generators instead?

karlb: I've make this more memory efficient. We'll still have the whole graph in memory, but that should be ok.

PR #825, opened 2020-07-29
[Raiddit] Fix raiddit requirements

konradkonrad:
`raiden` and `raiden-contracts` need to be explicitely installed from their respective git branch `raiddit`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-raiden==1.1.0
-
+-e git+https://github.com/raiden-network/raiden@raiddit#egg=raiden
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,3 +1,3 @@
-raiden==1.1.0
-
+-e git+https://github.com/raiden-network/raiden@raiddit#egg=raiden
+-e git+https://github.com/raiden-network/raiden-contracts@raiddit#egg=raiden-contracts
```
palango, line 2: Does that work for you? I had to use `git+https://github.com/raiden-network/raiden.git@raiddit`

konradkonrad: it seems (tm) to work here

PR #836, opened 2020-08-12
Add MS/PFS Prometheus endpoint

ezdac:
This adds an enpoint for Prometheus metrics (`/metrics`) and add RED metrics
as well as metrics on the services rewards claimed.

See #7 for reference!

- [x]  other services are missing
- [x]  `tests/libs/test_matrix.py::test_matrix_lister_smoke_test` fails in the overall test context, but passes individually - looks like some SSL monkey patching problem? Maybe there is no proper teardown in another test? -  *edit:  this is a local (probably MacOS) specific problem, CI-tests seem to work*

Edit:
Now, basic metrics are exposed over the Prometheus endpoint for the MS as well as for the PFS.


src/monitoring_service/metrics.py
```diff
@@ -0,0 +1,50 @@
...
+
+EVENTS_PROCESSING_TIME = Histogram(
+    "events_processing_duration_seconds",
+    "The overall time it takes for an event to get processed",
+    labelnames=["event_type"],
+)
+
+
+@contextmanager
+def collect_event_metrics(event: Event) -> Generator:
```
ulope, line 43: Nitpick: the `Generator` type is generic and should be annotated with the concrete types: https://docs.python.org/3/library/typing.html#typing.Generator

ulope (approved):
LGTM!

ulope:
@ezdac Can you rebase this, so we can merge?

Issue #855, opened 2020-09-10
multiple transactions in a row fail when using infura


fredo:
## Problem Definition
When using infura it is not enough to simply call `waitForTransactionReceipt` before continuing building new transactions. The reason is that it might be the case that infura is using different nodes behind the scenes (a reorg as a failure reason is possible as well). For some reason the following outcome after waiting for a transactionreceipt happens

`web3.eth.blockNumber == transaction_receipt["blockNumber"] - 1` 

if a next transaction gets build (and there are again multiple rpc calls happening which could have a different blocknumber again) and this transaction relies on the state change of the previous transaction, it can fail on several stages. (estimating gas will fail, the nonce is not increased, etc.)

instead of 

```
send_transaction(A)
wait_for_transaction_receipt(A)
send_transaction(B)
```

we should add a safety period like

```
send_transaction(A)
wait_for_transaction_receipt(A)
wait_until_block(transaction_receipt["blockNumber"] + Confirmation_Interval)
send_transaction(B)
```

Adding confirmation interval increases probability that the problem will not occur + decreases probabilty of a reorg causing the failure.

PR #856, opened 2020-09-10
fix transaction failing because prev one isnt in current chain state

fredo:
This fix waits for `transaction_receipt["blockNumber"] + Confirmation interval` before sending a new transaction.
Also it waits on the transaction receipt to have a key "blockNumber". Parity returns a receipt of a non mined transaction with missing "blockNumber" field. 

The fix is done in the service registry script. Should also be checked in other code parts.

fixes: #855

### DIFF
--- a/src/raiden_libs/service_registry.py
+++ b/src/raiden_libs/service_registry.py
@@ -88,8 +88,11 @@ def checked_transact(
     log.info(f"Starting: {task_name}")
     transaction_hash = function_call.transact({"from": sender_address})
 
+    confirmation_msg = ""
+    if wait_confirmation_interval:
+        confirmation_msg = " and waiting for confirmation"
     click.secho(
-        f"\nSending transaction: {task_name}"
+        f"\nSending transaction{confirmation_msg}: {task_name}"
         f"\n\tSee {etherscan_url_for_txhash(web3.eth.chainId, transaction_hash)}"
     )
 
--- a/src/raiden_libs/service_registry.py
+++ b/src/raiden_libs/service_registry.py
@@ -15,6 +15,7 @@ from web3.logs import DISCARD
 from web3.middleware import construct_sign_and_send_raw_middleware
 from web3.types import TxReceipt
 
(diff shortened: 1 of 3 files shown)


src/raiden_libs/service_registry.py
```diff
@@ -273,32 +295,44 @@ def maybe_promt(query: str) -> None:
...
-            checked_transact(
+            receipt = checked_transact(
                 web3=web3,
                 sender_address=service_address,
                 function_call=deposit_token_contract.functions.mint(latest_deposit),
                 task_name="Minting new Test RDN tokens",
+                wait_until_block=wait_until_block,
             )
 
+            wait_until_block = receipt["blockNumber"]
```
karlb, line 306: Shouldn't this be 
```suggestion
            wait_until_block = max(receipt["blockNumber"], wait_until_block)
```
?

karlb (changes requested):

src/raiden_libs/service_registry.py
```diff
@@ -76,8 +77,19 @@ def etherscan_url_for_txhash(chain_id: int, tx_hash: HexBytes) -> str:
...
+    function_call: ContractFunction,
+    task_name: str,
+    wait_until_block: int = 0,
 ) -> TxReceipt:
+
+    current_block = web3.eth.blockNumber
+
+    while current_block < wait_until_block + DEFAULT_NUMBER_OF_BLOCK_CONFIRMATIONS:
+        time.sleep(5)
+        current_block = web3.eth.blockNumber
```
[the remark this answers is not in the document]
fredo: I don't mind separating it into a different function. I just found it convinient to put an inherent wait into the function which is disabled by default

src/raiden_libs/service_registry.py
fredo: Isn't it per definition always the case that  `receipt["blockNumber"] >= wait_until_block`?

karlb: Yes it is, because `checked_transact` blocks until the transaction is mined. Ignore this.

karlb (approved):
Please fix the formatting before merging.

PR #858, opened 2020-09-15
Return PFS broadcast room id in info response

ulope:
This is necessary for
- raiden-network/raiden-service-bundle#187
- raiden-network/raiden#6442

With those PRs the set of federation servers can become disjunct between different Raiden client versions.
Therefore we can't rely on the previous check in the Raiden client if the PFS' server is in the known servers set.

The RoomID is uniquely identifying a specific broadcasting room and can therefore be used as a better information source for this check.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,14 @@
+[tool.black]
+line-length = 99
+target-version = ['py37']
+include = '\.pyi?$'
+exclude = '''
+/(
+    \.git
+  | \.mypy_cache
+  | \.tox
+  | \.venv
+  | build
+  | dist
+)/
+'''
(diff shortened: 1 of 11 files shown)


pyproject.toml
```diff
@@ -0,0 +1,14 @@
+[tool.black]
```
palango, line 1: Can we remove other config when we add this one?

palango (approved):

karlb (approved):

pyproject.toml
karlb: We can remove the `BLACK_PARAMS = --line-length 99` in the Makefile, but as far as I can see nothing else.

PR #862, opened 2020-09-21
Backwards compatible info

palango:
Replacement for https://github.com/raiden-network/raiden-services/pull/861 that is backwards compatible.

### DIFF
--- a/src/monitoring_service/api.py
+++ b/src/monitoring_service/api.py
@@ -57,6 +57,37 @@ class InfoResource(MSResource):
         return info, 200
 
 
+class InfoResource2(MSResource):
+    version = pkg_resources.get_distribution("raiden-services").version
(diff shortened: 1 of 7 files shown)


src/monitoring_service/api.py
```diff
@@ -47,7 +76,7 @@ def get(self) -> Tuple[dict, int]:
                 "confirmed_block": {
                     "number": str(
                         self.monitoring_service.context.ms_state.blockchain_state.latest_committed_block  # noqa
-                    )
+                    )  # noqa
```
karlb, line 79: do we need both `noqa`s?

karlb (approved):
This is one of the cases where code duplication is sensible. That makes it easier to remove the old code when v1 gets unsupported.

PR #870, opened 2020-10-06
Fix metrics endpoint path

ulope:
The metrics endpoint shouldn't be nested below `/api`.


karlb (approved):

karlb:
The path has to be adjusted in the test, too.

PR #878, opened 2020-10-30
Longer timeouts for first matrix sync

karlb:
This has been a problem when starting services that have not been
running in a while or were started for the first time. I still think it
should be faster than this, but I don't see a quick way to optimize,
right now.

### DIFF
--- a/src/pathfinding_service/cli.py
+++ b/src/pathfinding_service/cli.py
@@ -14,7 +14,7 @@ from web3 import Web3
 from web3.contract import Contract
 
 from pathfinding_service.api import PFSApi
-from pathfinding_service.constants import DEFAULT_INFO_MESSAGE, PFS_DISCLAIMER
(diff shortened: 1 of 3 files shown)


src/pathfinding_service/cli.py
```diff
@@ -126,7 +126,7 @@ def main(  # pylint: disable=too-many-arguments,too-many-locals
         service.start()
         log.debug("Waiting for service to start before accepting API requests")
         try:
-            service.startup_finished.get(timeout=60)
+            service.startup_finished.get(timeout=PFS_START_TIMEOUT)
```
palango, line 129: Should we adapt the for the MSRC as well?

palango (approved):

src/pathfinding_service/cli.py
karlb: I only found the `MATRIX_START_TIMEOUT` in the MSRC and not additional timeout like in the PFS.

Issue #768, opened 2020-04-08
Service registration script needs UX improvements


ulope:
## Problem Definition

I was helping Florian from KI Decentralized setting up a test installation of the current RSB release.
When running the registration script the following UX issues came up:

- [ ] The prompt [`The address of the service registry contract used is ... / I have checked that the address is correct and want to continue`](https://github.com/raiden-network/raiden-services/blob/bc5e0a8b43cf64ae06c883b551ab8b5804a51f6c/src/raiden_libs/service_registry.py#L209-L214) isn't actionable for the average RSB user. 
  We don't provide any guidance on how they can verify that it is indeed the correct address. 
- [x] [`The address of the token used is ... / The account balance of that token is ...`](https://github.com/raiden-network/raiden-services/blob/bc5e0a8b43cf64ae06c883b551ab8b5804a51f6c/src/raiden_libs/service_registry.py#L218-L225) does neither specify units nor display the name / symbol of the asset in question (Eth, token, etc.) (The same is true for the log call directly above)
- [x] [`The current required deposit is ...`](https://github.com/raiden-network/raiden-services/blob/bc5e0a8b43cf64ae06c883b551ab8b5804a51f6c/src/raiden_libs/service_registry.py#L246) has the same problem. No units, no asset name / symbol.
- [x] [`Note: The required deposit may change over time.`](https://github.com/raiden-network/raiden-services/blob/bc5e0a8b43cf64ae06c883b551ab8b5804a51f6c/src/raiden_libs/service_registry.py#L247) isn't clear. When is "over time" and how can it change?
- [x] The fact that on testnets tokens are automatically minted should be mentioned before the balance and the question whether to perform the registration are shown.
- [x] Even though the balance of the service token is shown there is no check if it's sufficient for the registration.

karlb:
> The prompt The address of the service registry contract used is ... / I have checked that the address is correct and want to continue isn't actionable for the average RSB user.
We don't provide any guidance on how they can verify that it is indeed the correct address.

@ulope Do you have any suggestion what the guidance could be? I have a hard time coming up with anything that is substantially more helpful.

ulope:
Well yeah, I agree that this can be probably not be adequately explained in a few lines in a console script.
But just linking to etherscan isn't helpful either IMO.
Without pretty detailed knowledge how the Raiden smart contracts and our deployment system works I don't think it's possible for an average user to learn anything there.

Issue #490, opened 2019-07-31
Improve service-registry script


palango:
The `register-service` script is currently quite basic. Some ideas for improvements
- [x] Add an `info` subcommand that shows the remaining time for the slot registration
- [x] Make it possible to prolong a slot in the registration
- [x] Add a command to withdraw the stake of a old slot (#748 )
- [x] Add promts and disclaimer to registration (#758)
- [x] Make withdraw automatically find withdrawable slots
- [x] Check if URL to register is pingable (does that cause circular dependencies during initial registration?)
- [x] Confirm transactions
- [ ] use `RPCClient` from raiden and maybe the smart contract proxies

palango:
See also https://github.com/raiden-network/raiden-services/issues/768

PR #890, opened 2020-11-20
Improve registry script usability

karlb:
A large collection of UI improvements for the service registration. Highlights:
* info command to show past deposits
* extend command to prolong registration
* no need to remember your deposit address, `withdraw` will find it for you

Closes https://github.com/raiden-network/raiden-services/issues/768 and https://github.com/raiden-network/raiden-services/issues/490.

### DIFF
--- a/src/raiden_libs/service_registry.py
+++ b/src/raiden_libs/service_registry.py
@@ -3,7 +3,7 @@ import sys
 import textwrap
 import time
 from datetime import datetime
-from typing import Dict, Optional
+from typing import Any, Dict, List, Optional
 
 import click
 import gevent
@@ -363,12 +363,12 @@ def register_account(
     log.info("Updated infos", current_url=current_url)
 
 
-def find_deposit_contract(
+def find_deposits(
     web3: Web3,
     service_address: Address,
     service_registry_contract: Contract,
     start_block: BlockNumber,
-) -> Address:
+) -> List[Dict[str, Any]]:
     """
     Return the address of the oldest deposit contract which is not withdrawn
     """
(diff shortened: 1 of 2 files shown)


src/raiden_libs/service_registry.py
```diff
@@ -176,15 +190,140 @@ def register(
...
+    # Get required deposit
+    required_deposit = service_registry_contract.functions.currentPrice().call()
+    click.secho(
+        f"\nThe current required deposit is fmt_amount({required_deposit})"
+        "\n\tNote: The required deposit continuously decreases, but "
+        "\n\t      increases significantly after a deposit is made."
+    )
+    if web3.eth.chainId == 1:
+        click.secho("You don't have sufficient tokens", err=True)
+        sys.exit(1)
```
konradkonrad, line 234: I don't understand the flow here:
a) where is the check that the user does not have "sufficient tokens"?
b) why don't we show the token amount the user has?

src/raiden_libs/service_registry.py
```diff
@@ -360,40 +442,106 @@ def maybe_prompt(query: str) -> None:
...
+        abi=CONTRACT_MANAGER.get_contract_abi(CONTRACT_CUSTOM_TOKEN),
+    )
+    fmt_amount = get_token_formatter(deposit_token_contract)
+
+    # Find deposits
+    deposits = find_deposits(web3, service_address, service_registry_contract, start_block)
+    deposits = [d for d in deposits if not d["withdrawn"]]
+    if not deposits:
+        click.echo("No deposits found!", err=True)
+        sys.exit(1)
```
konradkonrad, line 503: Not finding withdrawable deposits does not necessarily constitute an error (`exit > 0`) to me. Does it to you?

src/raiden_libs/service_registry.py
```diff
@@ -176,15 +190,140 @@ def register(
...
+
+    It will round to full tokens according to the decimal definition in the
+    token contract, but it will show at least `min_sig_figures` significant
+    figures. This ensures that small token amount stay readable and won't get
+    rounded to zero.
+    """
+    symbol = token_contract.functions.symbol().call()
+    decimals = token_contract.functions.decimals().call()
+
+    def format_token_amount(amount: float) -> str:
```
konradkonrad, line 207: Neat!

src/raiden_libs/service_registry.py
karlb: If withdraw is called, I would expect a withdraw to happen. It feels similar to an `rm` with a filename that does not exist to me. I do see your point, but prefer the current return code.

src/raiden_libs/service_registry.py
karlb: You are completely right. I moved those lines around a few times too often and apparently mixed up the conditions the last time. Should be better, now.

konradkonrad (approved):
lgtm, thanks!

Issue #379, opened 2019-05-29
Only return feedback token when routes are returned


palango:
Currently a feedback token is returned, even if no routes have been found. In that case the feedback token can be omitted.

PR #893, opened 2020-11-26
Only create feedback token when paths are found

karlb:
Without paths, there cannot be any feedback about them.

Closes https://github.com/raiden-network/raiden-services/issues/379

### DIFF
--- a/src/pathfinding_service/api.py
+++ b/src/pathfinding_service/api.py
@@ -194,10 +194,14 @@ class PathsResource(PathfinderResource):
             )
 
         # Create a feedback token and store it to the DB
-        feedback_token = create_and_store_feedback_tokens(
-            pathfinding_service=self.pathfinding_service,
-            token_network_address=token_network.address,
-            routes=paths,
+        feedback_token = (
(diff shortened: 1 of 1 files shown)


src/pathfinding_service/api.py
```diff
@@ -194,10 +194,14 @@ def post(self, token_network_address: str) -> Tuple[dict, int]:
...
-            token_network_address=token_network.address,
-            routes=paths,
+        feedback_token = (
+            create_and_store_feedback_tokens(
+                pathfinding_service=self.pathfinding_service,
+                token_network_address=token_network.address,
+                routes=paths,
+            )
+            if paths
+            else None
```
palango, line 204: How is `feedback_token.uuid.hex` on `None` gonna work?

Also,  this probably needs adjustments for the clients. Should we add this to the V2 API?

karlb: You're totally right about the `feedback_token.uuid.hex`, of course.

About the client compatibility: I didn't expect the client to look at the feedback token when not route is present. But that might not be the case.

PR #896, opened 2020-12-18
[WIP] Make the PFS use listeners on all homeservers

ezdac:
This will make the PFS connect to all federation servers and gather all presence updates 
only from the federation server of the user to avoid the presence sync problem across servers.

depends on: https://github.com/raiden-network/raiden/pull/6741
**This is a WIP, don't merge yet!**


fredo:
can be closed in favor of #899

Issue #897, opened 2020-12-18
Connect to all federation servers to fetch correct presences


ezdac:
This was the idea of @fredo:

To avoid the presence problem, a node could connect to all possible Matrix federation servers
in order to gather the correct presences for all users in the broadcast room, in order to avoid the presence sync problem 
across federation servers.

This is only viable for the PFS, since this is the most centralised instance in our protocol.
The PFS provides correct paths for initiators, possible presence problems after the path has been suggested has still to be considered between nodes along the path. 

A PR for this is in progress: #896

karlb:
Closed by #896.

PR #899, opened 2020-12-22
listen on presence by respective homeserver

fredo:
This PR is implementing #897 . I modified the WIP of #896 a bit and is to be continued in this PR since @ezdac is on vacation. 

The PFS creates a client for each homeserver in the list and listens on the presence events of each. Since we know that the presence of a user is always correct on their respective homeserver, presences are filtered by the homeservers.

All presences of Homeserver01 will be fetched by the sync of homeserver01 and so on...
This gracefully works around the present bug which happens to occur on forwarding presences due out the federation.

### DIFF
--- a/src/raiden_libs/matrix.py
+++ b/src/raiden_libs/matrix.py
@@ -1,5 +1,6 @@
 import sys
 from datetime import datetime, timedelta
+from functools import partial
 from typing import Callable, Dict, Iterable, List, Optional
 from urllib.parse import urlparse
 
@@ -7,6 +8,7 @@ import gevent
 import structlog
 from eth_utils import to_checksum_address
 from gevent.event import AsyncResult
+from gevent.pool import Pool
 from marshmallow import ValidationError
 from matrix_client.errors import MatrixRequestError
 from matrix_client.user import User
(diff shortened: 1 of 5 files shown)


src/raiden_libs/matrix.py
```diff
@@ -145,15 +145,23 @@ def __init__(
...
+        try:
+            known_servers = get_matrix_servers(
                 DEFAULT_MATRIX_KNOWN_SERVERS[Environment.PRODUCTION]
                 if chain_id == 1
                 else DEFAULT_MATRIX_KNOWN_SERVERS[Environment.DEVELOPMENT]
             )
 
+        except RuntimeError as ex:
+            if servers is None:
+                raise RuntimeError from ex
```
ulope, line 157: When you're not adding anything to an exception you can simply `raise` without any arguments that way the original exception will be re-raised.

src/raiden_libs/utils.py
```diff
@@ -19,3 +25,79 @@ def private_key_to_address(private_key: PrivateKeyType) -> Address:
...
+        def _filter_presence(event: Dict[str, Any], presence_update_id: int) -> None:
+            """
+            The actual presence listener callback. Filters out all presences from own server.
+            Drops presences from other servers. If the client is the "main client",
+            also call presence listener on presences of users of homeservers where no client
+            exists. This is a fallback if the client could not connect to another server
+            """
+            sender_server = event["sender"].split(":")[-1]
+
+            other_clients_servers = [
```
ulope, line 88: Nitpick: A set would be better (but since the list will only contain a few entries it doesn't really matter).

ulope (approved):
Two nitpicks. 
Otherwise LGTM.

PR #909, opened 2021-01-26
Forever reconnect to other clients upon failure

fredo:
This PR introduces the ClientManager which controls the status of matrix clients to their respective home server. Each home server has a `connect_client_worker` which will trigger the creation of a new client if the connection to the homeserver fails. 

A reconnect is described by creating a new client object, logging into the server, joining broadcast rooms and starting the greenlet `listen_forever`.

### DIFF
--- a/src/raiden_libs/matrix.py
+++ b/src/raiden_libs/matrix.py
@@ -1,14 +1,13 @@
 import sys
 from datetime import datetime, timedelta
-from functools import partial
 from typing import Callable, Dict, Iterable, List, Optional
 from urllib.parse import urlparse
 
 import gevent
 import structlog
 from eth_utils import to_checksum_address
-from gevent.event import AsyncResult
-from gevent.pool import Pool
+from gevent.event import AsyncResult, Event
+from gevent.greenlet import Greenlet
 from marshmallow import ValidationError
 from matrix_client.errors import MatrixRequestError
 from matrix_client.user import User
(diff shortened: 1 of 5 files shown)


src/raiden_libs/matrix.py
```diff
@@ -374,3 +287,170 @@ def _handle_message(self, room: Room, message: MatrixMessage) -> List[SignedMess
...
+                    Networks.RINKEBY.value,
+                    Networks.GOERLI.value,
+                    Networks.KOVAN.value,
+                ]
+                else []
+            )
+
+        except RuntimeError as ex:
+            if available_servers is None:
+                raise ex
```
ulope, line 332: To re-raise just `raise` is enough and will preserve the original traceback.

src/raiden_libs/matrix.py
```diff
@@ -374,3 +287,170 @@ def _handle_message(self, room: Room, message: MatrixMessage) -> List[SignedMess
...
+        while not self.stop_event.is_set():
+            stopped_client = self.server_url_to_other_clients.pop(server_url, None)
+            if stopped_client is not None:
+                self.user_manager.remove_client(stopped_client)
+            try:
+                client = self._start_client(server_url)
+                assert client.sync_worker is not None
+                client.sync_worker.get()
+            except (TransportError, ConnectionError):
+                log.debug("Could not connect to server", server_url=server_url)
```
ulope, line 399: Hm unless I'm missing something this looks to me like it will re-connect to the server in each loop without checking if the connection is actually dead?

Also this probably needs a lock so the state of the `server_url_to_other_clients` dict is consistent.

ulope (approved):
LGTM

One question below.

PR #910, opened 2021-01-28
pfs: Refresh presence when receiving message from offline sender

palango:
I've seen in multiple scenario logs that the PFS thinks a certain node is down, but receives messages from it. In that case it makes sense to refresh the presence.

### DIFF
--- a/src/pathfinding_service/service.py
+++ b/src/pathfinding_service/service.py
@@ -25,6 +25,7 @@ from pathfinding_service.typing import DeferableMessage
 from raiden.constants import PATH_FINDING_BROADCASTING_ROOM, UINT256_MAX
 from raiden.messages.abstract import Message
 from raiden.messages.path_finding_service import PFSCapacityUpdate, PFSFeeUpdate
+from raiden.network.transport.matrix import AddressReachability
 from raiden.utils.typing import BlockNumber, BlockTimeout, ChainID, TokenNetworkAddress
 from raiden_contracts.constants import CONTRACT_TOKEN_NETWORK_REGISTRY, CONTRACT_USER_DEPOSIT
 from raiden_contracts.utils.type_aliases import PrivateKey
@@ -307,6 +308,18 @@ class PathfindingService(gevent.Greenlet):
 
                     if changed_channel:
                         self.database.upsert_channel(changed_channel)
+
+                        # If we receive a message from a sender that seems to be offline,
+                        # refresh the presence
+                        sender_presence = (
+                            self.matrix_listener.user_manager.get_address_reachability(
+                                message.sender
+                            )
+                        )
+                        if sender_presence != AddressReachability.REACHABLE:
+                            self.matrix_listener.follow_address_presence(
+                                message.sender, refresh=True
+                            )
 
             except DeferMessage as ex:
                 self.defer_message_until_channel_is_open(ex.deferred_message)


karlb:
Can this happen during normal operation, or should I consider this as yet another workaround for the matrix presence problems?

palango:
> Can this happen during normal operation, or should I consider this as yet another workaround for the matrix presence problems?

I'd say it shouldn't happen during normal operations. It's just some idea that I got when looking at the logs.

But I also have to ask @fredo  whether this will automatically request presence from the correct homeserver

fredo:
This change will not fetch it from the respective homeserver as it uses `track_address_presence` which is not modified to work with the other clients.
Unfortunately, I think it will not even fetch the presence at all if the `user_id` is already added to the UserAddressManager and ReachabilityState is not `AddressReachability.UNKNOWN` (i.e. Unreachable). In this case it will return [here](https://github.com/raiden-network/raiden/blob/ca7dc76936df01d1d1197d5055c3cd8c6708d3fa/raiden/network/transport/matrix/utils.py#L319) and not fetch the presence at all. 

Thinking about the problem itself, could it be the case that the message in the room is received before the presence got propagated. Or is there a distinctive order? As we have seen lately that we have some delays in propagation, could it be the case that it is simply received in the wrong order?
I dont see a reason (yet) why the PFS should care if it receives a message from an offline sender (other than it looks suspicious)

karlb:
>  I dont see a reason (yet) why the PFS should care if it receives a message from an offline sender (other than it looks suspicious)

We could always set the sender's state to online in that case. The only problem I'm afraid of is that this will hide problems, which can make testing and debugging harder.

palango:
> This change will not fetch the presence from the respective homeserver as it uses `track_address_presence` which is not modified to work with the other clients.

Isn't that a problem we should fix then? In the scenarios the PFS will be online when the nodes come online and a channel is opened. The channel opened triggers the call to `follow_address_presence`, and that one should be reliable!

fredo:
> > This change will not fetch the presence from the respective homeserver as it uses `track_address_presence` which is not modified to work with the other clients.
> 
> Isn't that a problem we should fix then? In the scenarios the PFS will be online when the nodes come online and a channel is opened. The channel opened triggers the call to `follow_address_presence`, and that one should be reliable (and therefore ask the homeserver of the user)!

sure thing, that should be the case. Can you create an issue for that? 

On the other hand, as I mentioned this fix would not fetch the presence if the user is already known and the reachability is set to offline. So as a counter measurement it could make sence to call `fetch_presence` directly in the context of this PR. Although I dont think it is necessary to do so

andrevmatos:
Yes, `fetch_presence` is called on a client's server, but seems to be proxied through the federation, as I always got correct presences with it even when presence bug happened.
A fix for this is necessary, there's something very wrong currently with PFS, it's losing presences too often, several scenarios are red because of this, while this doesn't seem to be the widespread presence bug anymore, since we have logs for that on the clients which didn't get triggered, i.e. presence on the federation seem to be working fine with our setPresence workaround, but PFS is still broken.

fredo:
I will have a look into the logs

fredo:
After analyzing the logs, I come to the following conclusion: 

The wrong result (Offline node) does not happen by `get_presence()` rather than a step before in `follow_address_presence`. The call `populate_user_ids()` which is calling the API endpoint `search_user_directory` gives a wrong result, namely an empty list of user ids. Even if the user already joined the global room. 

This leads to the fact, that the PFS does not think the user is offline. It has the reachability state `UNKNOWN` for the user. This happens under the following circumstances. The PFS does get the presence through two occasions. A) Upon calling `follow_address`, it will call `get_presence` for each user id populated. B) through the sync by presence updates from the server.

Obviously, if the node comes online and joins the global room, the PFS will receive a presence update from the user's home server (@fredo 's workaround). Since the user is not registered in the UserAddressManager, this update will be ignored. Eventually, the PFS calls `follow_address_presence` and populates user_id's from its own server. If the return value is empty (meaning the PFS's homeserver does not know the user for whatever reason) then it will not fetch the presence either.

This false behavior comes together with wrong presence behavior and leads me to the assumption that these problems are related to each other. 

## How could the user directory be empty?

From the failing scenario I analyzed  it seems that the call `search_user_directory` is also affected by the presence bug. Or rather by what is causing the presence bug namely a bug in passing events in the federation. If events are being delayed or even dropped, then it might also happen with the event that a user joined a room (`room state event membership`).

### When should the user directory return a given user

From the spec: 
```
Performs a search for users on the homeserver. The homeserver may determine which subset of users are searched, however the homeserver MUST at a minimum consider the users the requesting user shares a room with and those who reside in public rooms (known to the homeserver).
```

This means, as soon as the user is known to be a member of the room it should always be found in the user directory. Further testings approved this assumption. As soon as my homeserver received the membership event I could find a test user in the user directory. 
This being said leaves me with the assumption that the membership event could also be affected by being sent at a later stage. In the failing scenario, we performed the `search_user_directory` call on `t01` three minutes **after** the user joined the global room on `t04` and the result was still an empty list. Performing it now, returns the user correctly.

Note: We do not know if the membership event arrived on time or if the user directory was empty due to some other failure.

## How to deal with it
First of all, it looks to me like the same problem as the presence bug. If the user directory is also affected by this bug, then we could also use the workaround to perform searches on the respective transport servers.

Another alternative could be to track presences of all users in the global room as we can assume that a raiden user will have a channel eventually. This extends @palango 's PR by not only fetching the presence if a message arrives but also assume any user in the global room is a raiden user and this presence can be tracked. The race condition we have (channel open -> follow address, and receiving presence) is also handled in the raiden code base by simply accepting one and therefore not rely on the order of events happening. 

Third alternative but I do not prefer: we could fetch presence for all possible user ids for a given address (number of servers).

konradkonrad:
I believe this PR can be closed, as it was followed up by #912. @fredo @palango ?

PR #912, opened 2021-02-02
Track address by presence updates

fredo:
This is alternative 2 from the discussion in #910 .

Instead of tracking presences by a channel open event, presences are being tracked once we receive a presence update from any user in the global room.

We assume that any user in the global room is a raiden user and will eventually create or already has created at least one channel.

In the future, spamming prevention should be set in place. We assume that spamming a global room will first result in a DoS attack against the server before it hits the PFS.

In order to hold the UserAddressManager in a reasonable size, any `leave` event for a user from a global room (which is effectively a user being purged due to 2 days of inactivity) could be used to untrack the address. This is an optimization and not part of the PR currently.

depends on: #909

### DIFF
--- a/src/raiden_libs/utils.py
+++ b/src/raiden_libs/utils.py
@@ -4,9 +4,13 @@ from uuid import UUID
 from coincurve import PrivateKey, PublicKey
 from eth_utils import keccak
 
-from raiden.network.transport.matrix import AddressReachability
+from raiden.network.transport.matrix import AddressReachability, UserPresence
 from raiden.network.transport.matrix.client import GMatrixClient
-from raiden.network.transport.matrix.utils import DisplayNameCache, UserAddressManager
+from raiden.network.transport.matrix.utils import (
+    DisplayNameCache,
+    UserAddressManager,
+    address_from_userid,
+)
 from raiden.utils.typing import Address, Any, Callable, Dict, Optional
 from raiden_contracts.utils.type_aliases import PrivateKey as PrivateKeyType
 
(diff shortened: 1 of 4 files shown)


karlb:
This has the same five commits as https://github.com/raiden-network/raiden-services/pull/909. I would have expected additional commits for this PR.

src/raiden_libs/utils.py
```diff
@@ -97,3 +101,60 @@ def _filter_presence(event: Dict[str, Any], presence_update_id: int) -> None:
...
+
+    def _presence_listener(self, event: Dict[str, Any], presence_update_id: int) -> None:
+        """
+        Update cached user presence state from Matrix presence events.
+
+        Due to the possibility of nodes using accounts on multiple homeservers a composite
+        address state is synthesised from the cached individual user presence states.
+
+        This is a copy from the super class with a slight change.
+        Any user will automatically added to the whitelist whenever the PFS receives
```
karlb, line 113: Typo: will automatically BE added

src/raiden_libs/utils.py
```diff
@@ -97,3 +101,60 @@ def _filter_presence(event: Dict[str, Any], presence_update_id: int) -> None:
...
+            return
+
+        user_id = event["sender"]
+
+        if event["type"] != "m.presence" or user_id == self._user_id:
+            return
+
+        address = address_from_userid(user_id)
+
+        # Not a user we've whitelisted, skip. This needs to be on the top of
```
karlb, line 127: How does `address is None` relate to being whitelisted? All raiden users should return a non-None value from `address_from_userid`, shouldn't they?

fredo: Maybe the comment is misleading as I had to change the if condition below. The function is mainly a copy from its superclasse but the if condition below. I changed the comment.

karlb (approved):

ezdac:
Should we merge that in order for all the dependent PRs to go through?
If so, let's do a rebase-merge instead of the default squash - otherwise the pendent PRs will get merge conflicts.

fredo:
This PR still depends on #909 which is currently not reviewed yet.

PR #913, opened 2021-02-08
Use legacy resolver until eth circular dependency is fixed

fredo:
Eth packages have an circular dependency which are not compatible to each other. the new dependency resolver will not finish the installation.


palango (approved):
👍

PR #918, opened 2021-02-12
Change MatrixListener to only accept `to_device` messages

### LINKED ISSUE #914: Listen to and process toDevice messages
## Description

PFS needs to accept and process to_device messages by calling the respective handler. In order to be backwards compatible it should also call the handler on room messages. 

Additionally, the handler should be called on all clients. A Raiden node will send to_device messages to the local PFS `MXID`.

### LINKED ISSUE #915: Only join discovery room
PFS and MS dont need to join their message rooms anymore. But they need to join discovery room instead in order to be able to receive presence updates from the users.

The sync filter also needs to be changed to not accept any message from rooms.

The message handler for rooms can therefor be removed. (if its not inherent by the usage of the GMatrixClient class) 

This change will break backwardscompatibility.

ezdac:
Now the MatrixListener's sync-handler will also dispatch
to_device messages without an associated room to the message handler.
The message handler now also supports messages without an associated
room in it's call signature, while the message handling
does not need to be changed.

Resolves #914

We now switch to to-device communication for node-to-service
updates that were formerly handled by the respective service broadcast
room.
The services transport thus don't listen to messages in broadcast
rooms anymore and will only join the global discovery room
in order to still track user presences.

The sync message-handler ignores room messages and will only process
to-device messages.
The sync filter will ignore the discovery room messages for all matrix
clients of the services.

Resolves #915

This PR depends on (/ includes) #912

### DIFF
--- a/src/raiden_libs/matrix.py
+++ b/src/raiden_libs/matrix.py
@@ -16,7 +16,7 @@ from monitoring_service.constants import (
     MATRIX_RATE_LIMIT_ALLOWED_BYTES,
     MATRIX_RATE_LIMIT_RESET_INTERVAL,
 )
-from raiden.constants import DeviceIDs, Environment, Networks
+from raiden.constants import DeviceIDs, Environment, MatrixMessageType, Networks
 from raiden.exceptions import SerializationError, TransportError
 from raiden.messages.abstract import Message, SignedMessage
 from raiden.network.transport.matrix.client import (
@@ -212,9 +212,6 @@ class MatrixListener(gevent.Greenlet):
     def _handle_matrix_sync(self, messages: MatrixSyncMessages) -> bool:
         all_messages: List[Message] = list()
         for room, room_messages in messages:
-            # Ignore toDevice messages
-            if not room:
-                continue
 
             for text in room_messages:
                 all_messages.extend(self._handle_message(room, text))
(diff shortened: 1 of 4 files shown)


src/raiden_libs/matrix.py
```diff
@@ -374,3 +271,170 @@ def _handle_message(self, room: Room, message: MatrixMessage) -> List[SignedMess
...
+    def _start_client(self, server_url: str) -> GMatrixClient:
+        assert self.user_manager
+        if self.stop_event.is_set():
+            raise TransportError()
+
+        if server_url == self.main_client.api.base_url:
+            client = self.main_client
+        else:
+            client = make_client(
+                handle_messages_callback=lambda messages: True,
```
fredo, line 394: In order for the other clients to also listen on toDevice messages you need to pass the handler here as well

fredo (changes requested):

src/raiden_libs/matrix.py
ezdac: The newest commit now passes the main-client's handler to all sub-clients.
Note that the message handler accepts to-device as well as room-messages for all clients.
If duplicated message handling at this point is a problem, the handlers would have to be separated,
so that only the main client listens on the room.

fredo (approved):
LGTM

ezdac:
I also added the implementation for #915 in this PR, to keep the number of dependent PR's lower.
Still this would need a rebase on top of #921.
Let's wait until #909 and #912 is merged to move forward with this.

fredo:
could you rebase to the current master? I merged a bunch of PRs

ezdac:
Could you please review the  HEAD-commit concerning #915 - this got added after you initial review

fredo (approved):
looks good !

PR #921, opened 2021-02-15
Use own device id

fredo:
PFS and MS need use their own device ID in order to receive messages intended for them. 

depends on: #912


fredo:
I needed to merge that with an older PR as it already used the curren Raiden develop branch

Issue #928, opened 2021-02-22
Return matrix user ids along with route


karlb:
We want raiden clients not to rely on the presence information to reduce the load on matrix servers. To communicate with other nodes via toDevice messages, the Raiden client needs to know the other user's matrix user id. Provide these user ids for all nodes in a route along with the route response.

PR #929, opened 2021-02-22
Include matrix user ids in route response

karlb:
I'm using the protected `_address_to_userids` to make this work. It
would make sense to provide a public way to get this information during
a future refactoring.

Closes https://github.com/raiden-network/raiden-services/issues/928

### DIFF
--- a/src/pathfinding_service/model/token_network.py
+++ b/src/pathfinding_service/model/token_network.py
@@ -32,6 +32,7 @@ from raiden.utils.typing import (
     TokenAmount,
     TokenNetworkAddress,
 )
+from src.pathfinding_service.exceptions import InconsistentInternalState
 
 log = structlog.get_logger(__name__)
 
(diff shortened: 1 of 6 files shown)


src/pathfinding_service/model/token_network.py
```diff
@@ -210,6 +210,10 @@ def to_dict(self) -> dict:
         assert self.fees is not None
         return dict(
             path=[to_checksum_address(node) for node in self.nodes],
+            matrix_users=[
+                self.reachability_state._address_to_userids[node]  # pylint: disable=W0212
```
konradkonrad, line 214: I believe there can be some race conditions hidden in here. I think this should at least be guarded against `KeyError`s.

konradkonrad (approved):
Other than potential key errors on `self.reachability_state._address_to_userids[node]` this looks good to me!

src/pathfinding_service/model/token_network.py
karlb: I'll return a descriptive 500 in that case, now.

PR #931, opened 2021-02-23
Fix bad imports from `src`

karlb:
I relied on my autoimporter, but that apparently messed up in a smart
enough way to fool CI.


konradkonrad (approved):

konradkonrad:
Tests still fail though...

karlb:
> Tests still fail though...

Raiden updated structlog and this causes the services to use a more recent structlog, too. That has more stricter typing which can't be properly checked with our current declaration. I'll silence this type warning.

PR #933, opened 2021-02-23
Return matrix user ids in dict

karlb:
Returning a mapping instead of a list should make this more pleasant to
use in the client.


konradkonrad (approved):
:+1:

ulope (approved):
LGTM

Issue #930, opened 2021-02-23
API Endpoint to give presence / Matrix user id for given address


fredo:
## Problem Definition

With removing the global presence room users need to know about the presence of their direct peers upon direct payments. The PFS has this information.

## Outcome
The PFS will provide an endpoint for providing the matrix user id which is online to the requester such that they can act based on this information.

Input: Address of requested user id
Output: Last seen Matrix user id and presence status

PR #936, opened 2021-02-24
Add endpoint user

fredo:
This PR adds an endpoint where the PFS returns a `user_id` for a given address if this user is online derived from synapse's presence state. If there is no user online then a `404` will be returned.

This endpoint is specifically going to be used for direct payments but can also be used for other use cases.

There are additional minor fixes in the tests as they did not use the correct interface for `get_user_ids_for_address` in the user manager as well irregular matrix user ids. 

fixes: #930

### DIFF
--- a/src/pathfinding_service/api.py
+++ b/src/pathfinding_service/api.py
@@ -133,6 +133,7 @@ class PathsResource(PathfinderResource):
             value=path_req.value,
             reachability_state=self.pathfinding_service.matrix_listener.user_manager,
         )
+
         if error:
             # this is for assertion via the scenario player
             if self.debug_mode:
@@ -192,14 +193,12 @@ class PathsResource(PathfinderResource):
                 to=to_checksum_address(path_req.to),
                 value=path_req.value,
             )
-
         # Create a feedback token and store it to the DB
         feedback_token = create_and_store_feedback_tokens(
             pathfinding_service=self.pathfinding_service,
             token_network_address=token_network.address,
             routes=paths,
         )
-
         return (
             {"result": [p.to_dict() for p in paths], "feedback_token": feedback_token.uuid.hex},
             200,
@@ -459,6 +458,32 @@ class InfoResource2(PathfinderResource):
         return info, 200
 
 
+class UserResource(PathfinderResource):
+    def get(self, user_address: str) -> Tuple[Dict[str, str], int]:
+        address = self._validate_address_argument(user_address)
+        user_manager = self.pathfinding_service.matrix_listener.user_manager
+        user_ids = user_manager.get_userids_for_address(address)
+        for user_id in user_ids:
+            if user_manager.get_userid_presence(user_id) in [
+                UserPresence.ONLINE,
+                UserPresence.UNAVAILABLE,
+            ]:
+                return {"user_id": user_id}, 200
+
+        raise exceptions.AddressNotOnline(address=user_address)
+
+    @staticmethod
+    def _validate_address_argument(address: str) -> Address:
+
+        if not is_checksum_address(address):
+            raise exceptions.InvalidAddress(
+                msg="The address needs to be a valid checksummed Ethereum address",
+                address=address,
+            )
+
+        return to_canonical_address(address)
+
+
 class SuggestPartnerResource(PathfinderResource):
 
     cache: Dict[str, Tuple[list, datetime]] = {}
@@ -639,6 +664,7 @@ class PFSApi:
             ),
             ("/v1/info", InfoResource, {}, "info"),
             ("/v2/info", InfoResource2, {}, "info2"),
+            ("/v1/user/<user_address>", UserResource, {}, "user"),
         ]
 
         if debug_mode:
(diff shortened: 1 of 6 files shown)


src/pathfinding_service/api.py
```diff
@@ -192,14 +193,13 @@ def post(self, token_network_address: str) -> Tuple[dict, int]:
...
             )
-
         # Create a feedback token and store it to the DB
         feedback_token = create_and_store_feedback_tokens(
             pathfinding_service=self.pathfinding_service,
             token_network_address=token_network.address,
             routes=paths,
         )
-
+        print([p.to_dict() for p in paths])
```
karlb, line 202: Debugging output?

src/pathfinding_service/api.py
```diff
@@ -459,6 +459,32 @@ def get(self) -> Tuple[dict, int]:
...
 
+class UserResource(PathfinderResource):
+    def get(self, user_address: str) -> Tuple[Dict[str, str], int]:
+        address = self._validate_address_argument(user_address)
+        user_manager = self.pathfinding_service.matrix_listener.user_manager
+        user_ids = user_manager.get_userids_for_address(address)
+        for user_id in user_ids:
+            if user_manager.get_userid_presence(user_id) in [
+                UserPresence.ONLINE,
+                UserPresence.UNAVAILABLE,
```
karlb, line 470: What does UNAVAILABLE mean, again? It doesn't sound like it should be considered the same way as ONLINE.

src/pathfinding_service/model/token_network.py
```diff
@@ -207,12 +223,11 @@ def estimated_fee(self) -> FeeAmount:
...
 
     def to_dict(self) -> dict:
-        assert self.fees is not None
-        user_id_of = self.reachability_state._address_to_userids  # pylint: disable=W0212
+        assert self.is_valid
         try:
             return dict(
                 path=[to_checksum_address(node) for node in self.nodes],
-                matrix_users={to_checksum_address(node): user_id_of[node] for node in self.nodes},
+                matrix_users=self.matrix_users,
```
karlb, line 230: Thanks for fixing this!

src/pathfinding_service/model/token_network.py
```diff
@@ -140,6 +141,32 @@ def _calculate_fees(self) -> Optional[List[FeeAmount]]:
...
+                if self.reachability_state.get_userid_presence(user_id) in [
+                    UserPresence.ONLINE,
+                    UserPresence.UNAVAILABLE,
+                ]:
+                    node_user_id = user_id
+                    # if a reachable user is found we arbitrarily choose
+                    # this user for the given address. There should not be another user online
+                    break
+
+            if node_user_id is None:
```
karlb, line 160: This is always true when this code is reached, isn't it?

If this `if` can be removed, then line 149 is also unnecessary.

src/pathfinding_service/model/token_network.py
```diff
@@ -140,6 +141,32 @@ def _calculate_fees(self) -> Optional[List[FeeAmount]]:
...
+        user_ids: Dict[str, str] = {}
+        for node in self.nodes:
+            node_user_ids = self.reachability_state.get_userids_for_address(node)
+            node_user_id = None
+            for user_id in node_user_ids:
+                if self.reachability_state.get_userid_presence(user_id) in [
+                    UserPresence.ONLINE,
+                    UserPresence.UNAVAILABLE,
+                ]:
+                    node_user_id = user_id
```
karlb, line 155: You could directly add to `user_ids` here instead of in line 167.
```suggestion
                    user_ids[to_checksum_address(node)] = node_user_id
```

src/pathfinding_service/typing.py
```diff
@@ -15,4 +16,10 @@ class AddressReachabilityProtocol(Protocol):
     def get_address_reachability(self, address: Address) -> AddressReachability:
         ...
 
+    def get_userid_presence(self, user_id: str) -> UserPresence:
+        ...
+
+    def get_userids_for_address(self, address: Address) -> Set[str]:
+        ...
+
     _address_to_userids: Dict[Address, Set[str]]
```
karlb, line 25: Now that you have added the functions above, `_address_to_userids` can be removed from this protocol.

src/pathfinding_service/api.py
fredo: It does. Unavailable is something like `invisible` but it is considered to be online. A bit weird naming though.

src/pathfinding_service/api.py
fredo: oh yes! qwill remove it

src/pathfinding_service/model/token_network.py
fredo: In line 155 `node_user_id` is assigned if the user_id is online

src/pathfinding_service/model/token_network.py
fredo: I rearranged the function a bit, please have a look

src/pathfinding_service/model/token_network.py
karlb: I've seen that, but I paired the `break` with the wrong `for` while writing that comment.

src/pathfinding_service/model/token_network.py
```diff
@@ -140,6 +141,32 @@ def _calculate_fees(self) -> Optional[List[FeeAmount]]:
...
+                if self.reachability_state.get_userid_presence(user_id) in [
+                    UserPresence.ONLINE,
+                    UserPresence.UNAVAILABLE,
+                ]:
+                    user_ids[checksummed_address] = user_id
+                    # if a reachable user is found we arbitrarily choose
+                    # this user for the given address. There should not be another user online
+                    break
+
+            if checksummed_address not in user_ids:
```
karlb, line 160: This could else be an `else` to the `for` loop instead of an `if`. But it's fine either way.

karlb (approved):

PR #937, opened 2021-02-25
Release 0.14.1: Disable thread logging, reset resolver settings, pin raiden

konradkonrad:
This sets up the version for a new release `0.14.1`

Patch changes:

- This fixes an issue when setting up `sentry`, where the gevent threading module monkeypatching wasn't completed, while the `sentry` setup already generated log events with thread logging enabled.

- This also removes a patch from the DNS resolver, that lead to problems when trying to resolve addresses from a VPN NS from within docker.

- Furthermore, the `raiden` dependency, formerly linked to the `develop` at github, is now linked to a specific commit. This is done to remove version ambiguity for the next release of `raiden-services`.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.14.0
+current_version = 0.14.1
 commit = True
 tag = False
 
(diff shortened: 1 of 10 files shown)


karlb (approved):
Let's see if problems with the thread resolver reappear somewhere.

Issue #939, opened 2021-02-26
Rename address metadata


fredo:
When returning the route as well as the `user` endpoint should return its information in the same and more transport generic data structure. 

Suggestion 

Endpoint `v1/<address>/metadata` or `v1/address/<address>`

Returning {"matrix_user_id": xxx} 

Route information should rename the field `matrix_users` to more generic `address_metadata` and the values of the dictionary  should contain the same as the return value of the endpoint

PR #942, opened 2021-03-02
add capabilities to address_metadata_endpoint

fredo:
This PR adds capabilities to the address metadata

It also renames the endpoint to `/v1/address/<checksummed_address>/metadata`. 

The return value in case of success looks like
```
{"user_id": <user_id>, "capabilities": <capabilities>}
```

Its the simplistic consistent approach to have capabilities somewhat fetchable from the PFS. It needs to be sorted out whether , how and when these information are actually necessary. 

Also the route metadata information are renamed to `address_metadata` and come in the format 
```
{<address>: {"user_id": <user_id>, "capabilities": <capabilities>}}
```

Maybe this could be defined as its own type in the Raiden client and then be used by the services.
 

fixes: #939

### DIFF
--- a/src/pathfinding_service/api.py
+++ b/src/pathfinding_service/api.py
@@ -1,7 +1,7 @@
 import collections
 from dataclasses import dataclass, field
 from datetime import MINYEAR, datetime
-from typing import Any, ClassVar, Dict, List, Optional, Tuple, Type, TypeVar, cast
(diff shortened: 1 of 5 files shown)


fredo:
I'm open for renaming to better names. Maybe for the route as `addresses_metadata` or `route_metadata`

konradkonrad (approved):
LGTM :+1:

PR #951, opened 2021-03-26
Allow selection of unstable env

karlb:
Part of https://github.com/raiden-network/light-client/issues/2521

PR #955, opened 2021-04-07
Allow selection of unstable env - fix setup

ezdac:
This PR is fixing an issue there was with #951.
It will include the commit of the fixed `raiden-contracts` version (https://github.com/raiden-network/raiden-contracts/pull/1448) and will not use the `--no-use-pep517` pip flag.
(see https://github.com/raiden-network/raiden-services/pull/951#issuecomment-814959179)


ezdac:
To me this looks good - but I would wait until the contracts are merged, since there are open questions and the commit hash is used in the requirements here.

karlb (approved):

PR #965, opened 2021-04-15
fix uncaught error in connecting to matrix server

fredo:
there was a `RaidenUnrecoverableError` uncaught from join_broadcast_room. I removed the call and joined directly by the `client.join_room` call


karlb (approved):
Nice, that's even simpler. I guess we can even remove the `join_broadcast_room` function from raiden, now.

PR #968, opened 2021-04-16
move UserAddressManager from raiden to raiden_libs

netcriptus:


fredo (approved):
LGTM in general. 

Probably the UserAddressManager can be slimmed out a lot as there are many functions which are probably not used and needed anymore. From my side, we can put that into a separate issue and discuss first what to remove here and what the requirements for the UAM are now.

the UAM has grown to be quite complex but how it is used now, mostly as a state keeper of presences, we can probably simplify that a lot.

fredo (commented):
Wouldn't it make sense to move the MultiClientUAM also in the same file or is there a reason why they are located differently?

PR #969, opened 2021-04-21
fix unhandled MatrixRequestError in PFS

fredo:
The failing scenarios described in [SP failure: lock expired](https://github.com/raiden-network/scenario-player/issues/667) were partly due to an unhandled exception in the PFS causing the PFS to stop syncing with a homeserver.

The bug happens when a server is currently unavailable and returning a `404` during `sync`. This happens when a Matrix server restarts. While the Matrix API translates this request error to a `MatrixRequestError`, [`GMatrixClient.listen_forever()`](https://github.com/raiden-network/raiden/blob/6455900101bdadade572689e83170b3a55ebc283/raiden/network/transport/matrix/client.py#L387) only handles `MatrixRequestError` when `error_code >= 500`. A `404` will be raised again and thrown back. This was an unhandled edge case in [`ClientManager.connect_client_forever()`](https://github.com/raiden-network/raiden-services/blob/fde698e3c8c21292bd95ac49ac3da87313fe9744/src/raiden_libs/matrix.py#L384)) and thus stopping this greenlet. Additionally, the greenlet exception was only handled during stop of the PFS which can happen at a much later time. Killing this greenlet will stop trying to reconnect to the homeserver. This leaves the PFS unconnected to the homeserver and thus not being able to receive messages and presence updates for its local users.

### solution 

The PFS will now handle the `MatrixRequestError`. To avoid situations where there is another unexpected exception raised, any other exception will be handled by logging and gracefully shutting down the PFS. Upon shutdown the exception in the greenlet will be raised again. The PFS will then be restarted. There should not be another unhandled exception though.

### DIFF
--- a/src/raiden_libs/matrix.py
+++ b/src/raiden_libs/matrix.py
@@ -301,7 +301,6 @@ class ClientManager:
         self.broadcast_room_id: Optional[RoomID] = None
         self.broadcast_room: Optional[Room] = None
         self.chain_id = chain_id
-        self.startup_finished = AsyncResult()
         self.stop_event = Event()
         self.stop_event.set()
 
@@ -356,10 +355,7 @@ class ClientManager:
         self.user_manager = user_manager
         try:
             self._start_client(self.main_client.api.base_url)
-        except (TransportError, ConnectionError):
-            # When the sync worker fails, waiting for startup_finished does not
-            # make any sense.
-            self.startup_finished.set()
+        except (TransportError, ConnectionError, MatrixRequestError):
             return
 
         for server_url in [
@@ -391,8 +387,20 @@ class ClientManager:
                 client = self._start_client(server_url)
                 assert client.sync_worker is not None
                 client.sync_worker.get()
-            except (TransportError, ConnectionError):
-                log.debug("Could not connect to server", server_url=server_url)
+            except (TransportError, ConnectionError) as ex:
+                log.error("Could not connect to server", server_url=server_url, exception=ex)
+            except MatrixRequestError as ex:
+                log.error(
+                    "A Matrix Error occurred during sync", server_url=server_url, exception=ex
+                )
+            except Exception as ex:
+                log.error(
+                    "An unhandled Exception occurred during sync. Restarting PFS.",
+                    server_url=server_url,
+                    exception=ex,
+                )
+                self.main_client.stop()
+                raise
 
     def _start_client(self, server_url: str) -> GMatrixClient:
         assert self.user_manager
--- a/src/raiden_libs/matrix.py
+++ b/src/raiden_libs/matrix.py
@@ -355,7 +355,12 @@ class ClientManager:
         self.user_manager = user_manager
         try:
             self._start_client(self.main_client.api.base_url)
-        except (TransportError, ConnectionError, MatrixRequestError):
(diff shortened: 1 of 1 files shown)


src/raiden_libs/matrix.py
```diff
@@ -356,10 +355,7 @@ def start(self, user_manager: MultiClientUserAddressManager) -> None:
         self.user_manager = user_manager
         try:
             self._start_client(self.main_client.api.base_url)
-        except (TransportError, ConnectionError):
-            # When the sync worker fails, waiting for startup_finished does not
-            # make any sense.
-            self.startup_finished.set()
+        except (TransportError, ConnectionError, MatrixRequestError):
             return
```
ulope, line 359: Is this correct?
If we can't connect just silently return?

Hm, I assume the assert in `MatrixListener._run()` will fail. But then don't we deadlock on the `joinall` on the `startup_finished_greenlet`?

src/raiden_libs/matrix.py
```diff
@@ -391,8 +387,20 @@ def connect_client_forever(self, server_url: str) -> None:
                 client = self._start_client(server_url)
                 assert client.sync_worker is not None
                 client.sync_worker.get()
-            except (TransportError, ConnectionError):
-                log.debug("Could not connect to server", server_url=server_url)
+            except (TransportError, ConnectionError) as ex:
+                log.error("Could not connect to server", server_url=server_url, exception=ex)
+            except MatrixRequestError as ex:
```
ulope, line 392: This means that we assume `MatrixRequestError`s are transient. 
But I guess it's fine. If the server is out of action more long-term it's clients need to roam to a different server anyway.

ulope (approved):
Looks good.

A question (unrelated to the changes) below though.

src/raiden_libs/matrix.py
fredo: As this is the code base for the PFS to connect to all servers, this particular connection attempt should try as long as it lives even though the error might be persistent. Because the PFS is connected to all other homeservers with its own client there is no need to roam.

ulope: Yeah, I meant the Raiden nodes on a more long term "broken" homeserver will eventually roam to a different one.

src/raiden_libs/matrix.py
fredo: It will actually be handled in `MatrixListener._run()`. The main client is different from the other clients. If the main client fails for whatever reason the PFS shuts down and restarts. The forever sync loop only exists for the other clients.

When it silently returns the run method will end. the joinall has a timeout set to 0 and thus should not dead lock should it?

fredo: I added a log for better verbosity

ulope: > It will actually be handled in MatrixListener._run()

Yeah I got that. But it feels a bit convoluted.

> the joinall has a timeout set to 0 and thus should not dead lock should it?

Ah right. 
Well it's not related to this PR so let's merge it.

src/raiden_libs/matrix.py
fredo: I think thats the reason why a 404 is not handled in the MatrixClient

PR #971, opened 2021-04-22
add displayname to metadata #6954

netcriptus:


src/pathfinding_service/api.py
```diff
@@ -176,6 +176,7 @@ def post(self, token_network_address: str) -> Tuple[dict, int]:
             value=path_req.value,
             reachability_state=self.pathfinding_service.matrix_listener.user_manager,
             max_paths=path_req.max_paths,
+            user_manager=self.pathfinding_service.matrix_listener.user_manager,
```
fredo, line 179: If I see correctly, you replaced the reachability state with the user_manager. Why do we need to keep both?

src/pathfinding_service/model/token_network.py
```diff
@@ -148,17 +154,18 @@ def _get_address_metadata(
...
+                if self.reachability_state.get_userid_presence(user.user_id) in [
                     UserPresence.ONLINE,
                     UserPresence.UNAVAILABLE,
                 ]:
                     capabilities = self.reachability_state.get_address_capabilities(node)
                     metadata[checksummed_address] = {
-                        "user_id": user_id,
+                        "user_id": user.user_id,
                         "capabilities": capabilities,
+                        "displayname": user.displayname or user.get_display_name(),
```
fredo, line 168: `user.get_displayname()` implicitly returns `user.displayname` if it is set. So you calling the first is sufficient.

src/pathfinding_service/model/token_network.py
```diff
@@ -411,6 +418,7 @@ def _get_single_path(  # pylint: disable=too-many-arguments, too-many-locals
         visited: Dict[ChannelID, float],
         disallowed_paths: List[List[Address]],
         fee_penalty: float,
+        user_manager: MultiClientUserAddressManager,
```
fredo, line 421: duplication of reachability state here as well

src/pathfinding_service/model/token_network.py
```diff
@@ -492,6 +503,7 @@ def get_paths(  # pylint: disable=too-many-arguments, too-many-locals
         target: Address,
         value: PaymentAmount,
         max_paths: int,
+        user_manager: MultiClientUserAddressManager,
         reachability_state: AddressReachabilityProtocol,
```
fredo, line 507: same here

fredo (changes requested):
In general looks good to me.

The duplication of reachability_state and user_manager is a bit confusing to me. 

One problem which might occur is, if we do not have the displayname ready for the whole route, the API request will take a long time as the PFS will fetch the displaynames for every user. We could think about if it makes sense to fetch the displayname upon receiving the presence update directly if it is not sent with the update.

Furthermore as a side note, the `UserAddressManager` can probably simplified a lot as it is much more complex than we need it now. For example, we only need to store one user / presence update per address at a time. But we can also clean this up in a follow-up PR.

src/pathfinding_service/model/token_network.py
```diff
@@ -148,17 +148,18 @@ def _get_address_metadata(
         # Check node reachabilities
         metadata: Dict[str, Dict[str, Union[str, PeerCapabilities]]] = {}
         for node in self.nodes:
-            node_user_ids = self.reachability_state.get_userids_for_address(node)
+            node_users = self.reachability_state.get_users_from_address(node)
```
fredo, line 151: `get_users_from_address()` will perform a search on the server to find users that belong to an address. We should only rely on the presence updates received by the server reactively and not perform a search on the server. Only take what we have in the state.

src/raiden_libs/user_address.py
```diff
@@ -158,13 +158,16 @@ def populate_userids_for_address(self, address: Address, force: bool = False) ->
...
                 address,
-                (
-                    user.user_id
-                    for user in self._client.search_user_directory(to_normalized_address(address))
-                    if self._validate_userid_signature(user)
-                ),
+                (user.user_id for user in self.get_users_from_address(address)),
             )
 
+    def get_users_from_address(self, address: Address) -> List[User]:
```
fredo, line 164: As mentioned above, we should not use search_user_directory at all. That has several reason which have been introduced by the presenceless changes.

1. The Raiden node only asks on demand for a user_id where to send messages to. It expects only one user_id
2. A Raiden node is never intended to be online at two servers simultaneously
3. The PFS should only have the one source of truth, that is the received presence information. The only way where two users could be online simultaneously is if the user roams between servers. Then the PFS should always accept the latest received presence update.
4. Most likely the search user directory call should fail anyways as the users do not share a room with the PFS

fredo (changes requested):
We need to do some simplifications already here I guess. See my comments

src/raiden_libs/user_address.py
netcriptus: I'm a bit confused about the expected behaviour. This function came to be as a refactoring of a very nested list comprehension, and also so it could be reused, since the only way I found around the code to access the displayname is to have access to a user object.

All the things around this function suggests that we can expect that one address will point to more than one use, from this function (that used to be a comprehension) to the name of the function where it was extracted from, `populate_userids_for_address`.

So I remain with 2 questions:
1. Can we be really sure that `search_user_directory` will only return one user, despite the evidence?
2. If we shouldn't ever user the user directory, what part of the code is fetching and storing the displayname?

fredo: As per spec [presence updates](https://matrix.org/docs/spec/client_server/r0.6.0#m-presence) come with the displayname if there is any. We have seen that this is not always but most of the times the case. So with any presence event arriving in the sync response, it should contain the respective displayname of the user. If not and if there isnt any displayname in the displaycache it should probably be fetched right away by the `user.get_display_name()` call. 

`search_user_directory` should never be called and it probably never will return a user for an address as the PFS user does not share a room nor a public room with it.

Nevertheless, the displayname should already arrive with the presence update. The respetctive user object could be created right away.

src/pathfinding_service/model/token_network.py
```diff
@@ -155,10 +155,16 @@ def _get_address_metadata(
                     UserPresence.ONLINE,
                     UserPresence.UNAVAILABLE,
                 ]:
+                    displayname = (
+                        self.reachability_state._displayname_cache.userid_to_displayname.get(
```
fredo, line 159: doesn't mypy complain that you access an unknown field from `AddressReachabilityProtocol` ?

fredo (approved):
LGTM, thanks for the changes. I left one question though

src/pathfinding_service/model/token_network.py
netcriptus: No, MyPy is somehow fine with this line.

PR #988, opened 2021-05-05
Return user capabilities as url

netcriptus:
Capabilities are returned as a dictionary in the AddressMetadata. To comply with the outer representation in the matrix profile as well, we could comply and also pass the capabilities as the avatar_url string

### DIFF
--- a/src/raiden_libs/user_address.py
+++ b/src/raiden_libs/user_address.py
@@ -22,7 +22,7 @@ from raiden.network.transport.matrix.utils import (
     address_from_userid,
     validate_userid_signature,
 )
-from raiden.utils.typing import Address, Any, FrozenSet, PeerCapabilities, Set, Union
+from raiden.utils.typing import Address, Any, FrozenSet, Set, Union
 
 log = structlog.get_logger(__name__)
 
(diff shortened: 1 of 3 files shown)


fredo (approved):
LGTM thanks!

ulope:
Hm, is this really an improvement?
The only reason we use the craptastic url encoding in matrix is because we're abusing the avatar url field (and I'm increasingly convinced that it was a mistake to not just use base64 encoded json there too).

Adding this weird encoding to an otherwise completely JSON based REST-API doesn't seem right...

(Sorry for jumping in this late, only saw this now.)

netcriptus:
If I get this right, we're simply keeping the convention. We get the capabilities as a url from the matrix as well, so in my understanding it would be logical to not make the users expect a dictionary from a source but a unparsed url from another.

Also, this url is still encapsulated in a JSON object, so the concept of this being a JSON based REST-API is still there, only the content of one field changed to comply to what we get from a different service.

fredo:
That's right. It was a request by the LC team to keep the convention in place. Nonetheless, we could decide to change the encoding at all in the future as the PFS becomes more and more the source of user information.

PR #992, opened 2021-05-07
V0.15.0

fredo:
Release Branch


ulope:
The reason for the test failure is this: https://bugs.python.org/issue43176

PR #998, opened 2021-05-10
Add last_seen data to offline nodes

### LINKED ISSUE #991: Return last seen timestamp
In order to deactivate inactive users we need to know when a user was last seen online. Unfortunately, until now we cannot ask for user presence directly from the matrix server as we do not share a room with each other.
But a user purger could take advantage of the endpoint provided by its PFS to see presence. What's missing here is a timestamp of when this last change happened. Especially (and maybe only), when the user turned offline. The PFS returns a `404` and should additionally return the number in seconds since when the user is offline.

This gives a user purger the possibility to deactivate inactive users based on their offline period. 

Other possible solutions would involve the adaption of synapse itself

1) allow presence to be fetched based on the list provided in the presence router
2) introduce a background job to automatically deactivate users with an inactivity period longer than set in the synapse config

For now it is faster to provide our own solution.

netcriptus:
Closes #991 

Adding `last_seen` information to offline nodes. This will help the decision to delete inactive nodes.

### DIFF
--- a/src/pathfinding_service/api.py
+++ b/src/pathfinding_service/api.py
@@ -472,7 +472,7 @@ class AddressMetadataResource(PathfinderResource):
                 return {"user_id": user_id, "capabilities": capabilities}, 200
 
         raise exceptions.AddressNotOnline(
-            address=checksummed_address, last_seem=user_manager.last_seem_online(address)
+            address=checksummed_address, last_seen=user_manager.last_seen_online(address)
         )
 
     @staticmethod
--- a/src/pathfinding_service/api.py
+++ b/src/pathfinding_service/api.py
@@ -473,7 +473,7 @@ class AddressMetadataResource(PathfinderResource):
 
         offline_since = datetime.now() - user_manager.seen_offline_at(address)
         raise exceptions.AddressNotOnline(
-            address=checksummed_address, seen_offline_since=offline_since.seconds
+            address=checksummed_address, seen_offline_since=offline_since.total_seconds()
         )
 
     @staticmethod
--- a/src/pathfinding_service/api.py
+++ b/src/pathfinding_service/api.py
@@ -471,7 +471,9 @@ class AddressMetadataResource(PathfinderResource):
                 capabilities = user_manager.get_address_capabilities(address)
                 return {"user_id": user_id, "capabilities": capabilities}, 200
 
-        raise exceptions.AddressNotOnline(address=checksummed_address)
+        raise exceptions.AddressNotOnline(
+            address=checksummed_address, last_seem=user_manager.last_seem_online(address)
+        )
 
     @staticmethod
     def _validate_address_argument(address: str) -> Address:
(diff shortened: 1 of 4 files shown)


fredo (commented):
What am I missing?

src/raiden_libs/user_address.py
```diff
@@ -69,6 +69,7 @@ def __init__(
         self._log = None
         self._listener_id: Optional[UUID] = None
         self._capabilities_schema = CapabilitiesSchema()
+        self._first_seen_offline: Dict[Address, Any] = {}
```
konradkonrad, line 72: I think we could type this as
```
Dict[Address, datetime]
```
or
```
Dict[Address, Optional[datetime]]
```
no?

src/raiden_libs/user_address.py
```diff
@@ -94,6 +95,9 @@ def known_addresses(self) -> Set[Address]:
         # may be modified while these values are used. Issue: #5240
         return set(self._address_to_userids)
 
+    def seen_offline_at(self, address: Address) -> datetime:
+        return self._first_seen_offline.get(address, datetime.fromtimestamp(0))
```
konradkonrad, line 99: At a later point (or maybe right now?) we will want to make the default return value the timestamp of the time the PFS came online. That way we won't convey incorrect age information on unknown (to the PFS) users. In detail:
right now a user that has not been seen by the PFS will be considered offline since 30+ years, while all the PFS knows is an offline time since it was started.

konradkonrad: Let me reword this:
We want to return
```
(datetime.now() - PathFindingService.started_at).seconds
```
where `PathFindingService.started_at` will just be set on `PathFindingService.__init__()` to `datetime.now()`

src/pathfinding_service/api.py
```diff
@@ -471,7 +471,10 @@ def get(self, checksummed_address: str) -> Tuple[Dict[str, Union[str, PeerCapabi
                 capabilities = user_manager.get_address_capabilities(address)
                 return {"user_id": user_id, "capabilities": capabilities}, 200
 
-        raise exceptions.AddressNotOnline(address=checksummed_address)
+        offline_since = datetime.now() - user_manager.seen_offline_at(address)
+        raise exceptions.AddressNotOnline(
+            address=checksummed_address, seen_offline_since=offline_since.seconds
```
konradkonrad, line 476: I believe using `datetime.timedelta.seconds` is misleadingly wrong:
```python
>>> (datetime(year=2012, month=5, day=1) - datetime(year=2010, month=5, day=1)).seconds
0
```

src/raiden_libs/user_address.py
konradkonrad: ...or the analogous considering https://github.com/raiden-network/raiden-services/pull/998#discussion_r630931188

src/pathfinding_service/api.py
netcriptus: Seems like the correct method to use here would be `total_seconds()`

```python
>>> (datetime(year=2012, month=5, day=1) - datetime(year=2010, month=5, day=1)).total_seconds()
63158400.0
```

konradkonrad: :+1:

src/raiden_libs/user_address.py
netcriptus: ✅

konradkonrad (approved):
LGTM!

fredo (approved):

PR #1005, opened 2021-05-12
Fix a crash of the `service_registry` script on private chains

palango:
Needed for https://github.com/raiden-network/light-client/issues/2699

PR #1006, opened 2021-05-17
SR: Fix txhash display on private chains

palango:
Follow-up for https://github.com/raiden-network/raiden-services/pull/1005 with a missing adaption.

### DIFF
--- a/src/raiden_libs/service_registry.py
+++ b/src/raiden_libs/service_registry.py
@@ -90,6 +90,9 @@ def etherscan_url_for_address(chain_id: int, address: Address) -> str:
 
 
 def etherscan_url_for_txhash(chain_id: int, tx_hash: HexBytes) -> str:
+    if chain_id not in CHAINID_TO_ETHERSCAN_PREFIX:
+        return tx_hash.hex()
+
     return f"https://{CHAINID_TO_ETHERSCAN_PREFIX[chain_id]}etherscan.io/tx/{tx_hash.hex()}"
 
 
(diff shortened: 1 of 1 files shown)


src/raiden_libs/service_registry.py
```diff
@@ -90,6 +90,9 @@ def etherscan_url_for_address(chain_id: int, address: Address) -> str:
 
 
 def etherscan_url_for_txhash(chain_id: int, tx_hash: HexBytes) -> str:
+    if chain_id not in CHAINID_TO_ETHERSCAN_PREFIX:
+        return tx_hash.hex()
```
konradkonrad, line 94: From the usage of `etherscan_url_for_txhash` it seems this value is always user facing and it may be nice to be a little more verbose:
```suggestion
        return f"transaction with hash '{tx_hash.hex()}'"
```

```
ag --python etherscan_url
src/raiden_libs/service_registry.py
81:def etherscan_url_for_address(chain_id: int, address: Address) -> str:
92:def etherscan_url_for_txhash(chain_id: int, tx_hash: HexBytes) -> str:
112:        f"\n\tSee {etherscan_url_for_txhash(web3.eth.chainId, transaction_hash)}"
321:        f"\n\t\tSee {etherscan_url_for_address(web3.eth.chainId, event_args['deposit_contract'])}"
360:        f"\n\tSee {etherscan_url_for_address(chain_id, service_address)}"
376:        f"\n\tSee {etherscan_url_for_address(chain_id, service_registry_address)}"
387:        f"\n\tSee {etherscan_url_for_address(chain_id, deposit_token_address)}"
```

Side note: do you want to do a similar change for `etherscan_url_for_address` ?

palango: I already did in #1005 , but there I forgot to fix this function.

konradkonrad: > do you want to do a similar change for `etherscan_url_for_address` ? Ah I see, that was in PR #1005 this one is a follow up to.

palango: Improved the message here and for `etherscan_url_for_address` as well.

konradkonrad (approved):

PR #1009, opened 2021-05-17
assure displayname is present on metadata response

### LINKED ISSUE #1008: Metadata Endpoint does not return displayname
Endpoint not returning the displayname: https://github.com/raiden-network/raiden-services/blob/master/src/pathfinding_service/api.py#L472

netcriptus:
Fixes #1008

### DIFF
--- a/src/pathfinding_service/api.py
+++ b/src/pathfinding_service/api.py
@@ -468,8 +468,15 @@ class AddressMetadataResource(PathfinderResource):
                 UserPresence.ONLINE,
                 UserPresence.UNAVAILABLE,
             ]:
+                displayname = user_manager._displayname_cache.userid_to_displayname.get(
+                    user_id, None
+                )
                 capabilities = user_manager.get_address_capabilities(address)
-                return {"user_id": user_id, "capabilities": capabilities}, 200
+                return {
+                    "user_id": user_id,
+                    "capabilities": capabilities,
+                    "displayname": displayname,
+                }, 200
 
         offline_since = datetime.now() - user_manager.seen_offline_at(address)
         raise exceptions.AddressNotOnline(
--- a/tests/pathfinding/test_api.py
+++ b/tests/pathfinding/test_api.py
@@ -540,6 +540,7 @@ def test_get_address_metadata(api_url: str, api_sut: PFSApi):
     response = requests.get(url)
     response_body = response.json()
     assert response.status_code == 200
+    assert response_body["displayname"] is None
     assert response_body["user_id"] == get_user_id_from_address(address)
     assert response_body["capabilities"].startswith("mxc://")
 


src/pathfinding_service/api.py
```diff
@@ -468,8 +468,15 @@ def get(self, checksummed_address: str) -> Tuple[Dict[str, Union[str, PeerCapabi
                 UserPresence.ONLINE,
                 UserPresence.UNAVAILABLE,
             ]:
+                displayname = user_manager._displayname_cache.userid_to_displayname.get(
+                    user_id, None
+                )
```
karlb, line 473: Could be shortened to
```suggestion
                displayname = user_manager._displayname_cache.userid_to_displayname.get(
                    user_id
                )
```

fredo:
LGTM. 

We should think about if for whatever reason, displayname is not available upon the call, if it makes sense that the PFS requests the displayname on the spot. This might avoid ugly races.

fredo (approved):

src/pathfinding_service/api.py
```diff
@@ -468,8 +468,15 @@ def get(self, checksummed_address: str) -> Tuple[Dict[str, Union[str, PeerCapabi
                 UserPresence.ONLINE,
                 UserPresence.UNAVAILABLE,
             ]:
+                displayname = user_manager._displayname_cache.userid_to_displayname.get(
```
karlb, line 471: What does it mean when there is no entry in `userid_to_displayname`? Did the user just not set a display name? For Raiden nodes, this must not be the case, right?

src/pathfinding_service/api.py
netcriptus: I added the default on purpose to make it explicit that this will be the returned value in case this user ID is not present. Do you think this is not useful?

karlb: It's not helpful for me personally, but I don't mind either way. It clearly is nitpicking.

src/pathfinding_service/api.py
netcriptus: This seems to be a rare case. This dictionary is populated [here](https://github.com/raiden-network/raiden/blob/develop/raiden/network/transport/matrix/utils.py#L142), and will remain empty in case this displayname is unknown to the Matrix server or if there was a communication problem.

fredo: Furthermore, Presence is not stored if the displayname cannot be validated. It creates a new state called [`Presence.SERVER_ERROR`](https://github.com/fredo/raiden-services/blob/6e688c35503e8fc1998c1f2e4b405ea7916a81fa/src/raiden_libs/user_address.py#L316) which will show the node as offline from the outer perspective. 
As a consequence, any node which is seen as online, has displayname available.

PR #1015, opened 2021-05-28
Register the MS for rinkeby

karlb:
This allows the MS to run, so that we can have successful scenarios on
rinkeby, too.


palango:
@karlb Do we still need this? Has this been done manually on rinkeby?

karlb:
> @karlb Do we still need this? Has this been done manually on rinkeby?

I used this change to register it. Before, the SP failed on rinkeby.

palango (approved):

PR #1038, opened 2021-07-12
Use released versions instead of git commits

### LINKED ISSUE #938: Manage `raiden` requirement
## Problem Definition

Without a new release of https://github.com/raiden-network/raiden we can't link a proper version that has all necessary features for a release. Therefore #937 pinned raiden to a commit. We need to make sure to change that afterwards for ongoing development.

palango:
Get rid of pinned git commits in the dependencies.

Fixes #938

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-git+https://github.com/raiden-network/raiden.git@9e4ac56117314bc4630e8bf68da2424c01989025
+raiden==3.0.0rc6
 
 sentry-sdk[flask]==1.4.3
 prometheus_client==0.11.0


palango:
So we need to wait for a new raiden release to include the contracts 0.38 which removes the solidity requirements.

konradkonrad (commented):
is it worth it to merge this RC dependency, or should we wait for the release?

palango:
This is now blocking https://github.com/raiden-network/raiden-services/pull/1100 , so I'd like to merge it now.

netcriptus (approved):

Issue #1040, opened 2021-07-15
Handle `ChannelSettled` in PFS


palango:
The PFS currently relies on the ChannelClosed event, and doesn't even handle the ChannelSettled event:
https://github.com/raiden-network/raiden-services/blob/37c6f4816e2a6a084bd4e7e9c0eb58b501870b66/src/pathfinding_service/service.py#L228-L229

This is a problem with the new cooperative settlement function, that only emit a `ChannelSettled` event: https://github.com/raiden-network/raiden-contracts/pull/1474#discussion_r667904905

The PFS needs to be adapted to listen to settlement events as well.

PR #1051, opened 2021-08-03
PFS: handle settled events for coop settle

palango:
Fixes https://github.com/raiden-network/raiden-services/issues/1040

Handle `ChannelSettled` events in the PFS. This wasn't necessary before, as there was always a `ChannelClosed` event emitted. But with cooperative settlement a single `ChannelSettled` might be emitted and must be handled.

### DIFF
--- a/src/pathfinding_service/service.py
+++ b/src/pathfinding_service/service.py
@@ -2,7 +2,7 @@ import collections
 import sys
 import time
 from dataclasses import asdict
-from typing import Dict, Iterator, List, Optional
+from typing import Dict, Iterator, List, Optional, Union
 
 import gevent
 import sentry_sdk
@@ -34,6 +34,7 @@ from raiden_libs.events import (
     Event,
     ReceiveChannelClosedEvent,
     ReceiveChannelOpenedEvent,
+    ReceiveChannelSettledEvent,
     ReceiveTokenNetworkCreatedEvent,
     UpdatedHeadBlockEvent,
 )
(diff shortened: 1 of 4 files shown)


src/pathfinding_service/service.py
```diff
@@ -285,6 +288,19 @@ def handle_channel_closed(self, event: ReceiveChannelClosedEvent) -> None:
             )
             metrics.get_metrics_for_label(metrics.ERRORS_LOGGED, metrics.ErrorCategory.STATE).inc()
 
+    def handle_channel_settled(self, event: ReceiveChannelSettledEvent) -> None:
```
karlb, line 291: `handle_channel_settled` and `handle_channel_closed` could be the same function if we are willing to sacrifice (or special case) the error logging in `handle_channel_closed`.

karlb (approved):

src/pathfinding_service/service.py
palango: I reworked it a bit, let me know if this is better.

If we want we could also store if the channel was closed in the DB and then act on this. But it feels a bit over-engineered for this case.

src/pathfinding_service/service.py
```diff
@@ -265,21 +266,29 @@ def handle_channel_opened(self, event: ReceiveChannelOpenedEvent) -> None:
...
         if channel_deleted:
-            token_network.handle_channel_closed_event(event.channel_identifier)
+            token_network.handle_channel_removed_event(event.channel_identifier)
         else:
-            log.error(
-                "Received ChannelClosed event for unknown channel",
+            # If a ChannelSettled event is received at this point, it is unclear if the channel
+            # is not found because the channel was settled non-cooperativly or because of some
+            # error. As the former is more likely, we lower the log level.
+            log_func = log.info if is_settled_event else log.info
```
karlb, line 289: It says `log.info` in both cases.

palango: 👀

PR #1054, opened 2021-08-11
Build latest image in actions

palango:
Docker hub doesn't build images on demand any more without a paid subscription. So lets do this stuff by hand 😩 

This image is used by the `services-dev` server which runs he nightly services.

### DIFF
--- a/.github/workflows/update_latest_image.yml
+++ b/.github/workflows/update_latest_image.yml
@@ -1,0 +1,24 @@
+name: Update `latest` image
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  push_to_registry:
+    name: Update `latest` image
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+
+      - name: Log in to Docker Hub
+        run: |
+          docker login --username $DOCKER_USERNAME --password $DOCKER_TOKEN
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          tags: raidennetwork/raiden-services:latest


.github/workflows/build_latest_image.yml
```diff
@@ -0,0 +1,26 @@
+name: Publish Docker image
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  push_to_registry:
+    name: Push Docker image to Docker Hub
```
weilbith, line 10: Nitpick: Isn't just the last step to push it? I at least includes the build as well. 🤷🏾

.github/workflows/build_latest_image.yml
```diff
@@ -0,0 +1,26 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_TOKEN }}
```
weilbith, line 20: I'm not sure if it is worth to use the action. As it is open source and by Docker themselves, I hope it is trustworthy. But in fact this whole tool just becomes useful if you use a custom registry at certain cloud providers etc. Else it comes down to [this little function](https://github.com/docker/login-action/blob/b30d77254f6c8aeebb0f00776ac12447e199919d/src/docker.ts#L25), which in fact is nothing more than `docker login --username $DOCKER_USERNAME --password $DOCKER_TOKEN`. So very simple. Just as some input here...

weilbith (commented):
I don't have access to the secrets here. But I assume you added them? 😉

.github/workflows/build_latest_image.yml
palango: Good point.

.github/workflows/build_latest_image.yml
palango: Tried to improve it.

weilbith: Very good. 👍🏾

weilbith (approved):
Let's hope it works. 🤞🏾

PR #1062, opened 2021-09-09
Update raiden and fix chain_id problems

karlb:
The chain_id problems have been introduced by usage of the chain_id opcode in the contracts.


palango (approved):
🎉

PR #1110, opened 2021-11-19
Fix unintentional creation of defaultdict key by __getitem__ calls

ezdac:
Before, the whitelisting mechanism of the services had a bug:

Several `__getitem__` calls meant to be **readonly**  to the internal defaultdict that tracks the whitelisted user-addresses were unintentionally causing the address being added to the dict with a default-value.
This in the end let the address pass the whiltelisting check:

https://github.com/raiden-network/raiden-services/blob/091c6ca8d18c0cc7b58f407ed7847c89a89fb338/src/raiden_libs/user_address.py#L311-L312

Which can potentially lead to performance issues and heavy load on the matrix servers.

This PR fixes this issue.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -3,6 +3,6 @@ coverage:
     patch: off
     project:
       default:
-        target: 90%
+        target: 88%
         threshold: 1
         base: auto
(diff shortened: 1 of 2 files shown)


palango (approved):
Nice find! 🏅 

Can you, while you're at it also lower the codecov coverage target: https://github.com/raiden-network/raiden-services/blob/master/.codecov.yml#L6 to something we hit, e.g. 88?

PR #1125, opened 2021-12-07
register: Display approximate cost before registration

### LINKED ISSUE #1105: Provide cost estimate in service registration script
## Problem Definition

With the current high eth price and also high gas costs registering or extending a service registration is pretty expensive.
Since the approximate gas needed is known in advance the service registration script should give users an estimate on the to-be expected cost.

(This was encountered during external testing of the RSB setup process.)

palango:
Resolves #1105

### DIFF
--- a/src/raiden_libs/service_registry.py
+++ b/src/raiden_libs/service_registry.py
@@ -29,6 +29,7 @@ from raiden_contracts.constants import (
     CONTRACT_SERVICE_REGISTRY,
     EVENT_REGISTERED_SERVICE,
 )
+from raiden_contracts.contract_manager import gas_measurements
 from raiden_libs.blockchain import get_web3_provider_info
 from raiden_libs.cli import blockchain_options, common_options, validate_address
 from raiden_libs.constants import CONFIRMATION_OF_UNDERSTANDING
@@ -200,7 +201,7 @@ def register(
 
 
 def get_token_formatter(
-    token_contract: Contract, min_sig_figures: int = 3
+    token_contract: Optional[Contract], min_sig_figures: int = 3
 ) -> Callable[[float], str]:
     """
     Return a function to nicely format token amounts.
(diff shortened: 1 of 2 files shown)


src/raiden_libs/service_registry.py
```diff
@@ -117,14 +118,15 @@ def checked_transact(
...
+    timeout = 60 * 10  # 10mins
+    transaction_receipt = web3.eth.wait_for_transaction_receipt(transaction_hash, timeout, 1.0)
     if wait_confirmation_interval:
         while (
             "blockNumber" not in transaction_receipt
             or web3.eth.block_number
             < transaction_receipt["blockNumber"] + DEFAULT_NUMBER_OF_BLOCK_CONFIRMATIONS
         ):
-            transaction_receipt = web3.eth.wait_for_transaction_receipt(transaction_hash)
+            transaction_receipt = web3.eth.wait_for_transaction_receipt(transaction_hash, timeout)
```
ezdac, line 129: The default is `poll_latency=0.1` - I assume this is the poll-interval (not documented in web3). Why are we polling 10x more often when waiting for the confirmation than for the initial block inclusion?

src/raiden_libs/service_registry.py
```diff
@@ -223,6 +229,19 @@ def format_token_amount(amount: float) -> str:
...
 
+def get_approximate_registration_cost(web3: Web3) -> int:
+    gas_costs = gas_measurements()
+    gas_cost = (
+        gas_costs["CustomToken.approve"]
+        + gas_costs["ServiceRegistry.deposit"]
+        + gas_costs["ServiceRegistry.setURL"]
+    )
+
+    gas_price = web3.eth.generate_gas_price() or 1
```
ezdac, line 240: in what scenario does this web3 method return `None`? Shouldn't the user be informed about that scenario (either warning and no gas estimation or a failure) rather than assuming the lowest possible gas_price then?
For this function this would mean returning None. Conversion to user-facing output would be done in the upper layers

ezdac (commented):
Looks good, I had 2 minor questions about the code

src/raiden_libs/service_registry.py
ezdac: I just noticed that this was there before - but maybe you know the reason anyways :)

palango: No, actually no clue, but I can set it to 1.0 as well.

src/raiden_libs/service_registry.py
palango: Good question, seems like this only happens when no gas price strategy is set. So I'd just remove the `None` case.

ezdac (approved):
Perfect, thanks!

PR #1144, opened 2022-01-25
adapt monitoring to use timestamps

netcriptus:
Adapt timeouts to use timestamps instead of block numbers


src/monitoring_service/handlers.py
```diff
@@ -57,9 +57,16 @@ def latest_committed_block(self) -> BlockNumber:
     def latest_confirmed_block(self) -> BlockNumber:
         return BlockNumber(self.get_latest_unconfirmed_block() - self.required_confirmations)
 
+    @property
+    def latest_timestamp(self) -> int:
+        return self.web3.eth.get_block("latest").timestamp  # type: ignore
```
weilbith, line 62: This does not work as intended. This makes "time updates" dependent on seeing new blocks. But this is exactly not possible anymore on roll-ups. Blocks gets created on-demand. This means it could be that a channel gets closed in one block and until the settle timeout expires no new block will be created. This means you clock is stuck in the past.

I'm not sure if the `context` is the correct place for this, but in fact it should be just the current system time. I just shimmed quickly through the code references to this property. It might be still necessary to access the last blocks timestamp somewhere.

src/monitoring_service/handlers.py
```diff
@@ -139,7 +146,7 @@ def channel_closed_event_handler(event: Event, context: Context) -> None:
         # Unfortunately, parity does the gas estimation on the current block
         # instead of the next one, so we have to wait for the first allowed
         # block to be finished to send the transaction successfully on parity.
-        trigger_block = _first_allowed_block_to_monitor(
+        trigger_block_timestamp = _first_allowed_timestamp_to_monitor(
```
weilbith, line 149:
```suggestion
        trigger_timestamp = _first_allowed_timestamp_to_monitor(
```

weilbith (commented):
Alright, I made a first review where I was just looking at the changes within `src/`. I know this is still under development and things will change.
Furthmore I must put a disclaimer here, because I'm completely new to the `raiden-services` code-base. Unfortunately this is a very fundamental change and it requires deeper knowledge of the internals for all the places where the PR made no changes. I had a discussion with Karl to get some details, but thats it. So the overall value of the review isn't too high.

Besides the comments, the most important part is that we need to adapt the main loop of the monitoring service. We do not wait anymore for the next block and only the continue to handle scheduled events etc. We check for new blocks, but if there are no, we just continue. Events are now scheduled based on timestamps and time runs without new blocks being mined. To not overheat the CPU (:wink:) we should probably ensure that this `while True` loop does not run frictionless and add a conditional `sleep` somewhere in case nothing is to do.
I think it makes no sense to completely rework how events are scheduled in time and how they are picked up. The existing polling mechanism should do the job fine enough and requires minimal changes.

src/monitoring_service/handlers.py
netcriptus: If you are sure about this, then we might have a bigger problem, because this is how the new version of the contracts are also implemented.

I did not find how to get the system time using Solidity, and the manuals have examples using `block.timestamp` to control current time.

src/monitoring_service/handlers.py
```diff
@@ -57,9 +57,16 @@ def latest_committed_block(self) -> BlockNumber:
...
 
+    @property
+    def latest_timestamp(self) -> int:
+        return self.web3.eth.get_block("latest").timestamp  # type: ignore
+
     def get_latest_unconfirmed_block(self) -> BlockNumber:
         return self.web3.eth.block_number
 
+    def get_timestamp(self, block_number: BlockNumber) -> int:
+        return self.web3.eth.get_block(block_number).timestamp  # type: ignore
```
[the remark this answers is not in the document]
weilbith: I "fixed" this in my latest commit.

✔️

src/monitoring_service/service.py
```diff
@@ -180,7 +180,7 @@ def _trigger_scheduled_events(self) -> None:
         a chain reorg.
         """
         triggered_events = self.context.database.get_scheduled_events(
-            max_trigger_block=self.context.get_latest_unconfirmed_block()
+            max_trigger_timestamp=self.context.latest_timestamp
```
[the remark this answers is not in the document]
weilbith: ✔️

src/monitoring_service/handlers.py
weilbith: I resolved the part where this gets used except for the IOU stuff. I'll resolve the comment here as the main reason got addressed and there are more comments for IOUs. Hopefully this will become removed entiery soon.

src/monitoring_service/handlers.py
```diff
@@ -339,6 +346,7 @@ def monitor_new_balance_proof_event_handler(event: Event, context: Context) -> N
 
         context.database.upsert_channel(channel)
 
+    # breakpoint()
```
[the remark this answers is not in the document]
netcriptus: you can safely ignore these kind of traces in the code for now, I'll clean this up once we're ready to turn this into a pull request :)

requirements.txt
```diff
@@ -1,4 +1,4 @@
-raiden==3.0.0rc9
+git+https://github.com/raiden-network/raiden.git@a5bdb80e0750901646cc3b7691bb1c9647ff5cdf#egg=raiden
```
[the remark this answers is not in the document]
netcriptus: I'm not sure if we can create an Arbitrum release for Raiden, given that the app is broken. We would have to implement all the modifications for Arbitrum to create a proper release.

src/monitoring_service/handlers.py
```diff
@@ -349,7 +357,7 @@ def monitor_new_balance_proof_event_handler(event: Event, context: Context) -> N
         # Unfortunately, parity does the gas estimation on the current block
         # instead of the next one, so we have to wait for the first allowed
         # block to be finished to send the transaction successfully on parity.
-        trigger_block = BlockNumber(channel.closing_block + channel.settle_timeout + 1)
+        trigger_block_timestamp = (channel.closing_block + 1) * 15
```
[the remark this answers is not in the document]
netcriptus: The trigger block for the event should be the block right after the closing block plus the timeout.

When it comes to block numbers, that's easy to estimate. But now we need the timestamp of a block in the future. The best estimation we have for that block is to consider the 15 seconds/block time average. (And you probably noticed the missing settle timeout, that's because by the time of this change we still didn't have access to the token network timeout, so it will be included in the next commit).

src/pathfinding_service/constants.py
```diff
@@ -19,8 +19,8 @@
 
 DEFAULT_INFO_MESSAGE = "This is your favorite PFS."
 
-# When a new IOU session is started, this is the minimum number of blocks
-# between the current block and `expiration_block`.
+# When a new IOU session is started, this is the minimum number of seconds
+# between the current block and `expiration_timestamp`.
```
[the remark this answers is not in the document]
weilbith: Further note: I think the number value below needs to be adapted or at least verified again.

requirements.txt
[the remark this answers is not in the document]
weilbith: This is !@#$

src/pathfinding_service/model/token_network.py
```diff
@@ -211,19 +207,6 @@ def _check_validity_and_calculate_fees(self) -> Optional[List[FeeAmount]]:
...
-            if ratio < DEFAULT_SETTLE_TO_REVEAL_TIMEOUT_RATIO:
-                log.debug(
-                    "Path invalid because of too low reveal timeout ratio",
-                    edge=edge,
-                    settle_timeout=edge["view"].settle_timeout,
-                    reveal_timeout=edge["view"].reveal_timeout,
-                    ratio=ratio,
-                    required_ratio=DEFAULT_SETTLE_TO_REVEAL_TIMEOUT_RATIO,
-                )
-                return None
```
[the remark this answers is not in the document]
netcriptus: Because settle timeout is no longer configurable on a channel. Therefore we don't need to check if the configuration is correct.

requirements.txt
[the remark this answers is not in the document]
netcriptus: I can think of 2 alternatives, and none of them sound great:

1. Implement the changes for Arbitrum in the python client, even though it will not be running on Arbitrum, and then we can create a release.

2. We implement on the services the part of the client we are importing, ending the dependency, and cutting the strings to the python client.

Both options demand a considerable amount of extra effort.

src/monitoring_service/handlers.py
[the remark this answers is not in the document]
weilbith: I don't think this is correct. It should be just the timestamp of the closing block plus the settle timeout. Because that is  what the contract is validating when you call the settle function. Correct? There is nothing about new blocks and block time involved here anymore.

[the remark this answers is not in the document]
netcriptus: The comment above this explains that this is something very particular to parity

src/pathfinding_service/api.py
```diff
@@ -263,21 +263,22 @@ def process_payment(  # pylint: disable=too-many-branches
 
     # Compare with known IOU
     latest_block = pathfinding_service.blockchain_state.latest_committed_block
+    latest_timestamp = pathfinding_service.web3.eth.get_block(latest_block).timestamp
```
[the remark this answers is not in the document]
weilbith: This seems a little tricky as below it gets tested for exact matches with an old IOU `claimable_until`. This solution might work, but I seriously have no clue if it is correct.

src/pathfinding_service/model/token_network.py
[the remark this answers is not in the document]
weilbith: 👍🏾

src/monitoring_service/handlers.py
```diff
@@ -57,6 +58,10 @@ def latest_committed_block(self) -> BlockNumber:
     def latest_confirmed_block(self) -> BlockNumber:
         return BlockNumber(self.get_latest_unconfirmed_block() - self.required_confirmations)
 
+    @property
+    def latest_timestamp(self) -> int:
+        return self.web3.eth.get_block("latest").timestamp  # type: ignore
```
karlb, line 63: Is this used anywhere?

src/monitoring_service/schema.sql
```diff
@@ -9,7 +9,8 @@ INSERT INTO blockchain DEFAULT VALUES;
 
 
 CREATE TABLE token_network (
-    address                 CHAR(42) PRIMARY KEY
+    address                 CHAR(42) PRIMARY KEY,
+    settle_timeout          INT
```
karlb, line 13: Since this comes from a uint256, it could theoretically overflow SQLite's int type. Practically, it probably can't.

src/pathfinding_service/claim_fees.py
```diff
@@ -30,10 +31,10 @@
     help="Current RDN/ETH price, used to check claimed amount > transaction cost",
 )
 @click.option(
-    "--expires-within",
-    default=4 * 60 * 24 * 7,  # one week
+    "--claimable-until-within",
+    default=60 * 60 * 24 * 7,  # one week
     type=click.IntRange(min=1),
-    help="Only IOUs which expire withing this number of blocks will be claimed",
+    help="Only IOUs which are claimable until the time range of now plus this number of seconds will be claimed",  # noqa: E501
```
karlb, line 37: Within the Raiden codebase, the intended naming system so far was that points in time are named "expiration" and time ranges "timeout" (which was not applied correctly in every case). The naming in this change works well in many cases, but in this specific case, I find it hard to read and understand.

tests/monitoring/monitoring_service/test_handlers.py
```diff
@@ -85,12 +86,14 @@ def setup_state_with_open_channel(context: Context) -> Context:
...
         participant2=DEFAULT_PARTICIPANT2,
-        settle_timeout=DEFAULT_SETTLE_TIMEOUT,
         block_number=BlockNumber(42),
     )
     assert context.database.channel_count() == 0
 
     channel_opened_event_handler(event, context)
+    context.web3.eth.get_block = lambda x: Mock(
+        timestamp=context.web3.eth.block_number * 15 if x == "latest" else x * 15
+    )
```
karlb, line 96: We might miss some edge cases during testing when time and block number always progress proportionally. But I don't have a good suggestion, since most changes would make the overall setup more complicated.

karlb (approved):
Looks good!

As noted in the TODO, we really have to think about how to deal with time differences between the service and the blockchain. Some operations will fail if executed too early or too late and we just can't know the blockchain time before call/estimate. That was easier with block times.

I assume we will have to start trying early and periodically retry for a certain time range. I assume the estimate will fail if the time is not right, so that it won't incur any gas costs.

Maybe the contract could give a hint in revert messages, which call might succeed at a later time?

requirements.txt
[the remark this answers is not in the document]
weilbith: ---

We agreed to use a git tag on the `raiden` repo to ensure the commit will not be lost due to the GitHub garbage collector.

src/monitoring_service/schema.sql
weilbith: So better use `HEX_INT`? 🤔

src/pathfinding_service/claim_fees.py
weilbith: That is very true. I made this change very "blindly" without much feelings. Do you have an idea what would sound better? Feel free to many suggestions.

src/monitoring_service/schema.sql
karlb: It's a bit ugly when looking at the db tables, but that is what I chose to be on the safe side whenever uint256 were stored.

andrevmatos: Not sure if it's worth to change, but `settle_timeout` isn't variable per `TokenNetwork`, but constant for all TNs in a `TokenNetworkRegistry`. Maybe we could avoid storing it, and instead just recover it from TNR every startup, and avoid having to deal with the field at all?

src/monitoring_service/handlers.py
```diff
@@ -176,7 +178,7 @@ def channel_closed_event_handler(event: Event, context: Context) -> None:
             "Settle period timeout is in the past, skipping",
             token_network_address=event.token_network_address,
             identifier=channel.identifier,
-            settle_until=settle_until,
+            settle_until=settleable_after,
```
andrevmatos, line 177: Is this correct? Assigning a variable `_after` to a `_until` raises some flags

andrevmatos (approved):
It looks fine for me. This is a complex refactoring, and final word should be given by testing by hand and on SP, but I'm confident this should be working or close to. Thank you guys!

src/monitoring_service/handlers.py
weilbith: This was a good catch. Especially because it was hint to pure naming in the above code. While this was still from the old code, it was simply not good and we took the opportunity to improve here. For the better naming I discussed that with Fernando and Paul. The commit addressing this got pushed.

weilbith: ✔️

Issue #1153, opened 2022-02-21
Monitoring Service - do not spam Arbitrum node with retries


weilbith:
With Arbitrum we face the issue of unsynchronized clocks (#1149). With #1152 this gets solved by a retry mechanism. While effective, this mechanism can start to spam the Arbitrum node depending on how much the time difference is and how fast the MS main loop works. Already a few seconds can lead to a lot of traffic as it always tries to redo the transaction as soon as possible.
The concern is that the centralized Arbitrum system could have mechanisms to block spamming nodes. In result the (all) monitoring services could become blocked. While this might not be the case it is a security issue as offline users can loose funds due to this.

A first solution approach could be to calculate the rescheduling event time according to the settle timeout. We have all the information to know from which point in time (according to our clock) the channel can still receive balance proof updates (`settleable_after`). Monitoring services can start to hand in their proof between 30-80% of the `settle_timeout` after a channel gets closed. This means from this point in time that is individual for each service, till `settleable_after` there is a timespan over which we can distribute our retries. Instead of spamming the Arbitrum node as fast as possible other the whole timespan (or actual just until the clock on the node matches) we can calculate a save time offset to which we reschedule the transaction retry by the monitoring service.

weilbith:
#1152 has changed after creation of this issue: failed transactions get retried with a delay of 10s.

taleldayekh:
@karlb Can you please support on this one?

PR #1158, opened 2022-02-24
Reduce the amount of retries for too early events

karlb:
Closes https://github.com/raiden-network/raiden-services/issues/1153.

I also cherry picked d9f200dac55ed00ff19348783ebedf2a20ebe50d from master, along with 9150691.

### DIFF
--- a/src/monitoring_service/service.py
+++ b/src/monitoring_service/service.py
@@ -19,7 +19,6 @@ from monitoring_service.constants import (
     KEEP_MRS_WITHOUT_CHANNEL,
 )
 from monitoring_service.database import Database
-from monitoring_service.events import ScheduledEvent
 from monitoring_service.exceptions import TransactionTooEarlyException
 from monitoring_service.handlers import HANDLERS, Context
 from raiden.utils.typing import (
@@ -88,12 +87,7 @@ def handle_event(event: Event, context: Context) -> None:
                     num_scheduled_events=context.database.scheduled_event_count(),
                 )
             except TransactionTooEarlyException:
-                log.debug("Event executed too early. Rescheduling for retry", handled_event=event)
-                context.database.upsert_scheduled_event(
-                    ScheduledEvent(
-                        trigger_timestamp=int(datetime.utcnow().timestamp()) + 10, event=event
-                    )
-                )
+                raise  # handled in _trigger_scheduled_events
             except Exception as ex:  # pylint: disable=broad-except
                 log.error("Error during event handler", handled_event=event, exc_info=ex)
                 sentry_sdk.capture_exception(ex)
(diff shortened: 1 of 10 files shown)


src/monitoring_service/service.py
```diff
@@ -197,8 +191,22 @@ def _trigger_scheduled_events(self) -> None:
...
-            self.context.database.remove_scheduled_event(scheduled_event)
+            try:
+                handle_event(event, self.context)
+            except TransactionTooEarlyException:
+                log.debug(
+                    "Event executed too early. "
+                    "Retry later and don't try any other scheduled events right now.",
+                    handled_event=event,
+                )
+                # When the scheduled event with the lowest timestamp fails with
```
weilbith, line 206: From the code in the database module I'm not sure: is the event list actually sorted by timestamps?
In any case I would probably help to also express this somewhere. At least in the docs of the database function or even in the function name itself.

weilbith (changes requested):
I like the basic solution. But as we both discussed, I think we need a new queue only for transactions which follows this approach. The problem is that the current priority queue of scheduled events we are working on is also used for many other events. Their handlers do not execute a transaction on the blockchain. I'm really not sure if it is good or possible to block all of these other events as they in-fact define the whole reactive functionality of the service. All the following events get blocked without a transaction in their handler:
- `UpdatedHeadBlockEvent`
- `ReceiveTokenNetworkCreatedEvent`
- `ReceiveChannelOpenedEvent`
- `ReceiveMonitoringNewBalanceProofEvent`
- `ReceiveChannelClosedEvent`
- `ReceiveChannelSettledEvent`
- `ReceiveMonitoringRewardClaimedEvent`

Especially the `UpdatedHeadBlockEvent` is quite important as it sets the internal blockchain state which is used to fetch new blocks if I'm not mistaken.
So are you really sure we can get away with blocking the main event queue here?

src/monitoring_service/service.py
karlb: Good catch! We really need to sort explicitly now. I'll push a commit for that.

karlb: Good catch! We really need to sort explicitly now. I'll push a commit for that.

karlb: Good catch! We really need to sort explicitly now. I'll push a commit for that.

weilbith: Got resolved with [e66c720](https://github.com/raiden-network/raiden-services/pull/1158/commits/e66c7204fb65678bcff63a51a35a789f1ca10bea) 👍🏾

weilbith (approved):
I'm positive that this does the job now. 👍🏾 
Let's see how it looks in practice.

PR #1159, opened 2022-02-24
fix usage of eth-tester import in production

weilbith:
See commit message for details.

### DIFF
--- a/src/monitoring_service/handlers.py
+++ b/src/monitoring_service/handlers.py
@@ -2,7 +2,6 @@ from dataclasses import dataclass
 from datetime import datetime
 
 import structlog
-from eth_tester.exceptions import TransactionFailed
 from eth_utils import encode_hex
 from web3 import Web3
 from web3.contract import Contract
@@ -530,8 +529,8 @@ def action_monitoring_triggered_event_handler(event: Event, context: Context) ->
             )
         )
     except Exception as exc:  # pylint: disable=broad-except
-        error_message = exc.args[0]
-        if isinstance(exc, TransactionFailed) and "not allowed to monitor" in error_message:
+        error_message = exc.args[0] if len(exc.args) > 0 else ""
+        if "not allowed to monitor" in error_message:
             raise TransactionTooEarlyException
         first_allowed = _first_allowed_timestamp_to_monitor(
             event.token_network_address, channel, context


weilbith:
Ah, damn. We need to enable the CI pipline for the `arbitrum` branch now.

weilbith:
I just updated the RSB for Arbitrum Rinkeby with this branch and it solves the issue as expected. Though I haven't yet tested a balance proof update yet. Is on the list. But it is not guaranteed that this code will run if the clock is behind the chain or in sync.

karlb:
> Ah, damn. We need to enable the CI pipline for the `arbitrum` branch now.

I included the relevant commit in https://github.com/raiden-network/raiden-services/pull/1158. I can split it out into a separate PR, if you like.

src/monitoring_service/handlers.py
```diff
@@ -533,8 +532,8 @@ def action_monitoring_triggered_event_handler(event: Event, context: Context) ->
             )
         )
     except Exception as exc:  # pylint: disable=broad-except
-        error_message = exc.args[0]
-        if isinstance(exc, TransactionFailed) and "not allowed to monitor" in error_message:
+        error_message = exc.args[0] if len(exc.args) > 0 else ""
```
karlb, line 535: I always wonder why there is no `list.get` method like there is for dicts. Would be nice, here.

karlb (approved):
Not the cleanest solution in the world, but simple and should be good enough.

weilbith:
> Not the cleanest solution in the world, but simple and should be good enough.

I'll rebase on it when merged. 👍🏾

src/monitoring_service/handlers.py
weilbith: I am not aware of such a function. But yes, I would also love to have it. Do you know of any other solution. Haven't written much Python code in a long time.

karlb: It's either your way or the "easier to ask forgiveness than permission" approach of
```python
try:
    error_message = exc.args[0]
except IndexError:
    error_message = ""
```
So no really elegant solution that I know of.

weilbith: Okay, than I would rather go with the current state if that is okay. It is only a single line and good enough to read. Nothing advanced.

PR #1165, opened 2022-03-01
add support for the new 'rpc' gas price

### LINKED ISSUE #1162: Add special gas price calculation for Arbitrum networks
Gas prizes on Arbitrum based chains work slightly different. Unfortunately does the connected Ethereum node not respond with a proper value when the default `web3.py` gas price strategy is asking it for a price. In result the transactions of the monitoring service do fail at the moment for Arbitrum setups.
This was already an issue for the `raiden-contracts` deployment tools. But there we only added a quick-and-dirty solution for test network deployments.
There is already some functionality to detect Arbitrum chains. So far this is used to set the number of confirmation blocks. This can be re-used for this case too.

weilbith:
Closes #1162 

The parser for the gas price parameter has become extended with a new price strategy. This 'rpc' strategy is especially important to run the RSB for Arbitrum networks.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-git+https://github.com/raiden-network/raiden.git@arbitrum-v0.1.0#egg=raiden
+git+https://github.com/raiden-network/raiden.git@arbitrum-v0.2.0#egg=raiden
 
 sentry-sdk[flask]==1.5.6
 prometheus_client==0.12.0
--- a/src/raiden_libs/cli.py
+++ b/src/raiden_libs/cli.py
@@ -157,11 +157,12 @@ def blockchain_options(contracts: List[str]) -> Callable:
                 "Set the gas price for ethereum transactions. If not provided "
                 "the 'normal' gas price startegy is used.\n"
                 "Available options:\n"
-                '"fast" - transactions are usually mined within 60 seconds\n'
                 '"normal" - transactions are usually mined within 5 minutes\n'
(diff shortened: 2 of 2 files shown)


weilbith:
@palango  where do you think does the documentation (recommending this option for Arbitrum RSB providers) fit better: here or at raiden-network/raiden-service-bundle ?

weilbith:
I don't think CodeCov does matter here? 🤔

palango:
> @palango where do you think does the documentation (recommending this option for Arbitrum RSB providers) fit better: here or at raiden-network/raiden-service-bundle ?

It'd put in in the RSB docs, and probably preset it as well if you have an arbitrum branch.

> I don't think CodeCov does matter here? 🤔

Nvm for those small changes.

palango (approved):

weilbith:
> It'd put in in the RSB docs, and probably preset it as well if you have an arbitrum branch.

Kinda challenging. The click option has a default set. So I have a hard time to determine if this was manually set by the user or not. I would have to remove the default an then set defaults programmatically. There I could then take Arbitrum into account. Do you think this is a good solution?

weilbith:
> MS_GAS_PRICE_STRATEGY=rpc

That is what I did for the Rinkeby RSB. But I thought you asked to automate this. Anyway I have the changes locally. Doing the commit now and then you can check that our.

palango:
All I meant I'd pu this in the docker compose file

netcriptus (approved):

palango (approved):

PR #1167, opened 2022-03-02
retry too early claim reward transactions

### LINKED ISSUE #1166: Monitoring Service - catch too early claim reward transactions
As for the `monitor` transaction, the `claimReward` must also run after a specific timestamp. Due to possibly unsynchronized clocks between the monitoring service and the blockchain (Arbitrum) node, this can fail. Therefore this case must be catched and the event should end up in the same retry mechanism as we already have established for the `monitor`ing.

weilbith:
Closes #1166

I'll test this now already in practice for our Arbitrum Rinkeby RSB setup.


weilbith:
Practical test has worked out successfully. 🎉

netcriptus (approved):

PR #1172, opened 2022-03-10
fix wrong UTC timestamp usage

### LINKED ISSUE #1171: Fix wrong conversation of UTC timestamp to POSIX format
For the changes to become compatible with Arbitrum (contracts) we introduced the usage of the following code at multiple locations to receive a POSIX formatted timestamp for the current UTC time: `datetime.utcnow().timestamp()`. What we didn't know was that the `timestamp()` function will convert this back into the local timezone again.
In result we get issues in the scenario player as the Raiden node sending an IOU using the correct time, has too big time difference to the local time of the Pathfinding service.
Please note that `datetime.utcnow()` indeed gives a correct date in the UTC timezone. But it is a timezone unaware `datetime` instance. Anyway, this means that all other usages where we exclusively work with `datetime` objects without "conversions" to POSIX format are just fine and need no adaption.
The proper way (according to the Python documentation), is to use `datetime.now(tz=timezone.utc).timestamp()` (in fact creating a timezone aware instance).

weilbith:
Fixes #1171

### DIFF
--- a/src/monitoring_service/handlers.py
+++ b/src/monitoring_service/handlers.py
@@ -1,5 +1,4 @@
 from dataclasses import dataclass
-from datetime import datetime
 
 import structlog
 from eth_utils import encode_hex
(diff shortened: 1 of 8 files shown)


weilbith:
CI is expected to fail for dependency issues on `master` right now. I'll rebase as soon as they got resolved.

netcriptus (approved):

